PR #2, opened 2021-04-26
Add Error Prone Compiler Check

vthacker:
Add Error Prone check at compile time to catch any bugs that can be caught by static analysis

This also fixes a bug that was found

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project kaldb: Compilation failure: Compilation failure: 
[ERROR] /Users/vthacker/slack-repos/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java:[334,36] error: [CollectionIncompatibleType] Argument 'entry.getKey()' should not be passed to this method; its type Long is not compatible with its collection's type argument String
[ERROR]     (see https://errorprone.info/bugpattern/CollectionIncompatibleType)
[ERROR] /Users/vthacker/slack-repos/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java:[341,31] error: [CollectionIncompatibleType] Argument 'entry.getKey()' should not be passed to this method; its type Long is not compatible with its collection's type argument String
[ERROR]     (see https://errorprone.info/bugpattern/CollectionIncompatibleType)
[ERROR] -> [Help 1]
```

### DIFF
--- a/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
+++ b/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
@@ -319,7 +319,7 @@ public class ChunkManager<T> {
     LOG.info("Closed chunk manager.");
   }
 
-  public void removeStaleChunks(List<Map.Entry<Long, Chunk<T>>> staleChunks) {
+  public void removeStaleChunks(List<Map.Entry<String, Chunk<T>>> staleChunks) {
     if (staleChunks.isEmpty()) return;
 
     LOG.info("Stale chunks to be removed are: {}", staleChunks);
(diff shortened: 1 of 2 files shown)


src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -319,7 +319,7 @@ public void close() {
     LOG.info("Closed chunk manager.");
   }
 
-  public void removeStaleChunks(List<Map.Entry<Long, Chunk<T>>> staleChunks) {
+  public void removeStaleChunks(List<Map.Entry<String, Chunk<T>>> staleChunks) {
```
vthacker, line 322: found by error-prone

mansu (approved):

PR #6, opened 2021-04-27
Create a maven build GH action.

mansu:
Crate a maven build github action that runs on every diff.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,25 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Java CI with Maven
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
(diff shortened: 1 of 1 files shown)


.github/workflows/maven.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+    - name: Build with Maven
```
vthacker, line 24:
```
  - name: Cache Maven packages
    uses: actions/cache@v2
    with:
      path: ~/.m2
      key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
      restore-keys: ${{ runner.os }}-m2
```

Perhaps we can add caching here?

.github/workflows/maven.yml
```diff
@@ -0,0 +1,25 @@
...
+      run: mvn -B package --file pom.xml
```
vthacker, line 25:
```
 mvn -B package --file pom.xml
```
this takes 2 mins for me to run locally. Let's see how fast does the server run it at :)

.github/workflows/maven.yml
vthacker: Not a blocker. We can add it later when needed.

vthacker (approved):

.github/workflows/maven.yml
mansu: I considered adding caching but decided against it for now since this process is new and also it only added 10-20 more seconds to the build process. Locally the tests take about 1.5 mins to run.

PR #9, opened 2021-04-29
Emit metrics for Kafka records processed

mansu:
Emit messages received and messages failed counters from Kafka so we can track the Kafka message consumption more closely.


vthacker (approved):
looks good!

PR #10, opened 2021-04-30
Expose metrics via /metrics endpoint in prometheus exposition format. Configurable server port.

mansu:
Add a prometheus metrics endpoint.

Tested on dev server.

Make server port configurable. Print server port so we can ping the server.

### DIFF
--- a/src/main/java/com/slack/kaldb/server/Kaldb.java
+++ b/src/main/java/com/slack/kaldb/server/Kaldb.java
@@ -76,7 +76,7 @@ public class Kaldb {
     new JvmGcMetrics().bindTo(prometheusMeterRegistry);
     new ProcessorMetrics().bindTo(prometheusMeterRegistry);
     new JvmThreadMetrics().bindTo(prometheusMeterRegistry);
-    LOG.info("Done Registering standard JVM metrics.");
+    LOG.info("Done registering standard JVM metrics.");
   }
 
   public void close() {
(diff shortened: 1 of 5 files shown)


src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -26,15 +29,12 @@
...
+  private final PrometheusMeterRegistry prometheusMeterRegistry;
 
   public Kaldb(Path configFilePath) throws IOException {
     LOG.info("Starting Metrics setup.");
-    // TODO: Initialize metrics and set up a metrics end point.
-    // Metrics.addRegistry(Metrics.globalRegistry);
-    meterRegistry = new SimpleMeterRegistry();
-    Metrics.addRegistry(meterRegistry);
-    // TODO: Expose a metrics scraper endpoint.
+    prometheusMeterRegistry = new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
```
vthacker, line 36: could this be a static final?

vthacker: `Kaldb` is initialed once so in some ways it's not needed as of now, but good so that we don't make any mistakes down the road?

mansu: Good idea! Changed!

vthacker (approved):

PR #12, opened 2021-05-06
Cleanup SearchResult API - Remove count parameter

vthacker:
The `count` and `total_count` parameters can cause confusion - 

Essentially `count` is the number of results returned. While `total_count` is the number of hits

Since we always return the `count` number of results finding count is as simple as getting the size of the search results.

Hopefully this change makes the API clear and also when writing tests we don't mix `count and total_count` for times when test cases add/match less than 10 docs and both values can be the same

### DIFF
--- a/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
+++ b/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
@@ -77,9 +77,19 @@ public class ReadWriteChunkImplTest {
       chunk.commit();
 
       SearchResult<LogMessage> results =
+          chunk.query(new SearchQuery(MessageUtil.TEST_INDEX_NAME, "*:*", 0, MAX_TIME, 10, 1000));
+      assertThat(results.totalCount).isEqualTo(100);
+
+      results =
           chunk.query(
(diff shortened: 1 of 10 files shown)


src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
```diff
@@ -315,16 +327,22 @@ public void testCommitBeforeSnapshot() {
...
       registry = new SimpleMeterRegistry();
-      LuceneIndexStoreImpl logStore =
+      logStore =
           LuceneIndexStoreImpl.makeLogStore(
               temporaryFolder.newFolder(), commitInterval, refreshInterval, registry);
       chunk = new ReadWriteChunkImpl<>(logStore, "testDataSet");
     }
 
+    @BeforeEach
+    public void cleanup() throws IOException {
```
mansu, line 342: Since we create a fresh log store for every test in a temporary folder do we need this additional cleanup before running every test?

mansu (commented):
The count change looks good.

src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
vthacker: Maybe I missed it, but I saw the temporary store being created for every test Class. And then there was no cleanup between two tests within the test class.

PR #13, opened 2021-05-09
Add merge metrics

vthacker:
Add merge metrics to gain insights into how merging performs during indexing loads

The default merge scheduler is CMS so we extend that and add metrics in the wrapper merge scheduler

### DIFF
--- a/src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
+++ b/src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
@@ -61,7 +61,6 @@ public class LuceneIndexStoreImplTest {
           findAllMessages(
               forgivingLogStore.logSearcher, MessageUtil.TEST_INDEX_NAME, "Message1", 10, 1);
       assertThat(results.size()).isEqualTo(1);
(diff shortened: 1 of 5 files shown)


src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
```diff
@@ -61,7 +61,6 @@ public void testSimpleIndexAndQuery() {
           findAllMessages(
               forgivingLogStore.logSearcher, MessageUtil.TEST_INDEX_NAME, "Message1", 10, 1);
       assertThat(results.size()).isEqualTo(1);
-      assertThat(forgivingLogStore.metricsRegistry.getMeters().size()).isEqualTo(4);
```
vthacker, line 64: every time we add a new metric we'll have to change all of this. I don't think this is adding any value.

If the test cares about the metric it should just assert that like it's doing here. So we can remove the total registry count

mansu: Yes, they have outlived their usefulness. I have been removing these also. Please feel free to remove them when you find them.

mansu (approved):

PR #14, opened 2021-05-11
A ZK based MetadataStore with basic CRUD operations

mansu:
A ZK based metadata store with basic CRUD operations and unit tests. 

The MetadataStore interface also makes the higher level classes oblivious to underlying implementation thus leaving room for other pluggable physical store implementations in the future.

### DIFF
--- a/src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
+++ b/src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
@@ -29,9 +29,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Metadata Store implements a store in zookeeper. This class interacts with ZK using the apache
- * curator library. It also abstracts it's users from the details of ZK and provides a simpler
- * higher level API.
+ * ZookeeperMetadataStore implements a store in zookeeper. This class interacts with ZK using the
+ * apache curator library. It also abstracts it's users from the details of ZK and provides a
+ * simpler higher level API.
  *
  * <p>All the public calls in this class are executed in it's own thread pool which makes the API
  * non-blocking while allowing for parallelism. The actual methods are implemented as private
--- a/src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
+++ b/src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
@@ -1,0 +1,337 @@
+package com.slack.kaldb.metadata;
+
+import static com.slack.kaldb.util.ArgValidationUtils.ensureNonEmptyString;
+import static com.slack.kaldb.util.ArgValidationUtils.ensureTrue;
+
+import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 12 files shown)


src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,337 @@
...
+
+  // A thread pool to run all the metadata store operations in.
+  private final ListeningExecutorService metadataExecutorService;
+
+  public ZookeeperMetadataStore(
+      String zkHostPath,
+      String zkPathPrefix,
+      int sessionTimeoutMs,
+      int connectionTimeoutMs,
+      RetryPolicy retryPolicy,
```
vthacker, line 66: we _might_ need a retry policy to be configurable based on the operation - I know for solr there were some operations that were best effort , some that could not be retried etc. I'll have to refresh my memory but thought I'd point out.

for now I think it's okay to keep it simple till we feel the need to

src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
```diff
@@ -0,0 +1,337 @@
...
+              LOG.info("Curator connection state changed to {}", connectionState);
+            });
+
+    /*
+     * If a ZK session expires, we need to create all the watches and ephemeral nodes again.
+     * In such a case, any ephermeral nodes and watches would expire and need to be re-created.
+     * To keep it simple for now, we terminate the process and let the process initialization
+     * register those nodes.
+     */
+    curator
```
vthacker, line 123: can you explain this a little to me ? Once we hit session expiry ( say a GC pause bigger than session timeout ) will we stop indexing ?

vthacker: I think I get it - `RuntimeHalterImpl` is what would happen - I am unsure if we want to do this rather than re-connect.

We can do the re-connect safely in a followup PR

vthacker (approved):
lgtm!

src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
mansu: Agree. Yeah, this is basic API for ZK CRUD. We will have another layer on top (one for snapshot metadata and one for node metadata etc..) on top of this API which will configure the setting for each use case. All those customizations would go into that API. The next diff would clarify this configuration more.

src/main/java/com/slack/kaldb/metadata/ZookeeperMetadataStore.java
mansu: The general philosophy for now is to make the process crash the process for now whenever we see a mis-configuration. This is preferable at this stage than having the process continue and result in an unrelated error which would be much more hard to debug.

In this specific case, we can re-connect the session, but we need a way to re-register all the watches etc again. Since those use cases don't exist yet, I decided to keep it simple and implement re-connect and other features in a follow up PR.

PR #8, opened 2021-04-29
Reserve commit point during S3 copy

vthacker:
Currently when we copy the files over to S3 we risk files being changed under-the-hood by lucene. Lucene can add/remove files in the index as and when documents get flushed/committed to disk 

We should use the `SnapshotDeletionPolicy#snapshot` and `SnapshotDeletionPolicy#release` API to ensure we're copying a point in time of the index that who's reference we can hold onto while the copying task is running.

```
// starting to roll over
[INFO ] 2021-04-23 11:01:59.048 [pool-3-thread-1] RollOverChunkTask - Start chunk roll over ChunkInfo{chunkId='log_data__1619200835341', chunkCreationTimeSecsSinceEpoch=1619200835, chunkLastUpdatedTimeSecsSinceEpoch=1619200919, dataStartTimeSecsSinceEpoch=1619062542, dataEndTimeSecsSinceEpoch=1619113903, chunkSnapshotTime=0}

[INFO ] 2021-04-23 11:01:59.048 [pool-3-thread-1] ReadWriteChunkImpl - Started RW chunk pre-snapshot ChunkInfo{chunkId='log_data__1619200835341', chunkCreationTimeSecsSinceEpoch=1619200835, chunkLastUpdatedTimeSecsSinceEpoch=1619200919, dataStartTimeSecsSinceEpoch=1619062542, dataEndTimeSecsSinceEpoch=1619113903, chunkSnapshotTime=0}

// Calling commit and then SearcherManager#maybeRefresh()
[INFO ] 2021-04-23 11:01:59.048 [pool-3-thread-1] LuceneIndexStoreImpl - Indexer starting commit for: /mnt/kaldb/data/indices/a48e3955-947b-4a87-a8ed-366a295a8a78

//Copying to S3 then throws this error ( sometimes )

[ERROR] 2021-04-23 11:02:06.392 [pool-3-thread-1] ReadWriteChunkImpl - Exception when copying RW chunk ChunkInfo{chunkId='log_data__1619200835341', chunkCreationTimeSecsSinceEpoch=1619200835, chunkLastUpdatedTimeSecsSinceEpoch=1619200919, dataStartTimeSecsSinceEpoch=1619062542, dataEndTimeSecsSinceEpoch=1619113903, chunkSnapshotTime=0} to S3.
java.io.UncheckedIOException: java.nio.file.NoSuchFileException: /mnt/kaldb/data/indices/a48e3955-947b-4a87-a8ed-366a295a8a78/_1f.dii
[...]

PR #23, opened 2021-06-10
reserve commit point during S3 copy

vthacker:
More details on https://github.com/slackhq/kaldb/pull/8

This is another approach instead of using IndexCommitRefHolder

### DIFF
--- a/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
+++ b/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
@@ -153,7 +153,7 @@ public class ReadWriteChunkImpl<T> implements Chunk<T> {
       return false;
     } finally {
       if (indexCommit != null) {
-        logStore.decIndexCommitRef(indexCommit);
+        logStore.releaseIndexCommitRef(indexCommit);
       }
     }
   }
--- a/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
+++ b/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
@@ -152,7 +152,7 @@ public class ReadWriteChunkImpl<T> implements Chunk<T> {
       LOG.error("Exception when copying RW chunk " + chunkInfo + " to S3.", e);
       return false;
     } finally {
-      logStore.releaseIndexCommitRef(indexCommit);
+      logStore.releaseIndexCommit(indexCommit);
     }
   }
 
--- a/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
+++ b/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
 import java.util.Collection;
+import org.apache.lucene.index.IndexCommit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -131,9 +132,11 @@ public class ReadWriteChunkImpl<T> implements Chunk<T> {
   public boolean snapshotToS3(String bucket, String prefix, S3BlobFs s3BlobFs) {
     LOG.info("Started RW chunk snapshot to S3 {}", chunkInfo);
 
+    IndexCommit indexCommit = null;
     try {
       Path dirPath = logStore.getDirectory().toAbsolutePath();
-      Collection<String> activeFiles = logStore.activeFiles();
+      indexCommit = logStore.getIndexCommit();
+      Collection<String> activeFiles = indexCommit.getFileNames();
       LOG.info("{} active files in {} in index", activeFiles.size(), dirPath);
       for (String fileName : activeFiles) {
         LOG.debug("File name is {}}", fileName);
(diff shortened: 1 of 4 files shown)


src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
```diff
@@ -148,6 +151,10 @@ public boolean snapshotToS3(String bucket, String prefix, S3BlobFs s3BlobFs) {
     } catch (Exception e) {
       LOG.error("Exception when copying RW chunk " + chunkInfo + " to S3.", e);
       return false;
+    } finally {
+      if (indexCommit != null) {
```
mansu, line 155: It's better to move the null check into the function instead of the call site.

src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
```diff
@@ -131,9 +132,11 @@ public void preSnapshot() {
   public boolean snapshotToS3(String bucket, String prefix, S3BlobFs s3BlobFs) {
     LOG.info("Started RW chunk snapshot to S3 {}", chunkInfo);
 
+    IndexCommit indexCommit = null;
```
mansu, line 135: Null initialization is redundant and can be omitted.

src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -239,21 +242,21 @@ public String toString() {
...
+    try {
+      return snapshotDeletionPolicy.snapshot();
+    } catch (IOException e) {
+      LOG.error("Tried to snapshot index commit but failed", e);
     }
+    return null;
+  }
 
-    DirectoryReader directoryReader;
+  public void decIndexCommitRef(IndexCommit indexCommit) {
```
mansu, line 254: Let's rename the function releaseIndexCommitRef to better reflect it's intended function.

src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
```diff
@@ -359,8 +360,8 @@ private void testS3Snapshot(String bucket, String prefix) throws Exception {
       assertThat(getCount(COMMITS_COUNTER, strictLogStore.metricsRegistry)).isEqualTo(1);
 
       Path dirPath = logStore.getDirectory().toAbsolutePath();
-      Collection<String> activeFiles = logStore.activeFiles();
-
+      IndexCommit indexCommit = logStore.getIndexCommit();
```
mansu, line 363: Let's inline this variable in both locations.

mansu (commented):
The diff looks good overall!

src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
vthacker: makes sense!

src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
vthacker: image

I can't remove the null check otherwise the compiler is complaining at me

src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
vthacker: sorry can you elaborate on what you mean here?

src/main/java/com/slack/kaldb/logstore/LogStore.java
```diff
@@ -24,6 +24,8 @@
 
   Path getDirectory();
 
-  Collection<String> activeFiles();
+  public IndexCommit getIndexCommit();
   // TODO: Add an isReadOnly and setReadOnly API here.
+
+  public void releaseIndexCommitRef(IndexCommit indexCommit);
```
mansu, line 30: Please move this method before the TODO. 

Also, add a comment on how these 2 methods should be used. We also need to document that the IndexCommit data can vary even if no data is added to index between calls since merge can change the index commit.

vthacker: done

src/main/java/com/slack/kaldb/logstore/LogStore.java
```diff
@@ -24,6 +24,17 @@
...
+  /**
+   * After a commit, lucene may merge multiple segments into one in the background. So, getting a listing of files in a lucene index will not return a stable set of files. 
+   * To address this, we can get an IndexCommit, which provides a consistent set of files at a given point in time and also protects the files from deletion.
+   * If a background merge happens after an indexCommit is created, lucene holds on to those files on disk even if the background merge has marked that file for deletion. 
+   * So, we need to release the index commit object once it's no longer needed using the releaseIndexCommit api call. 
+   * Since lucene can always decide to run background merges on an index, multiple calls to getIndexCommit, may return different lists of active files, even if no new commits are performed on the index.
+   */
+  public IndexCommit getIndexCommit();
+
+  public void releaseIndexCommitRef(IndexCommit indexCommit);
```
mansu, line 36: Can you please rename this method releaseIndexCommit to be consistent with the get call.

vthacker: > Let's rename the function releaseIndexCommitRef to better reflect it's intended function.

I renamed it to what you suggested last night :)

mansu: Sorry. I thought the Ref suffix was in the get call also. Can we drop the suffix here?

PR #27, opened 2021-06-14
No Need for  KalDBMergeScheduler DEBUG logging

vthacker:
We don't need to debug the merge scheduler right now


mansu (approved):

mansu (commented):
On a second thought, can we remove this since the issue is likely solved?

vthacker:
> On a second thought, can we remove this since the issue is likely solved?

I'm leaving it there to enable any debug logging in the future quickly

PR #32, opened 2021-06-17
Refactor Query Layer

vthacker:
This PR does two small things

1. Rename `KaldbService` to `KaldbSearcher` - This will serve as the query layer for local data search
2. Extract out adding the search service from the indexer code  - We want to take advantage of this in a followup PR which will add a multi node search query layer

### DIFF
--- a/src/main/java/com/slack/kaldb/server/Kaldb.java
+++ b/src/main/java/com/slack/kaldb/server/Kaldb.java
@@ -49,8 +49,7 @@ public class Kaldb {
     setupMetrics();
 
     // Create an indexer and a grpc search service.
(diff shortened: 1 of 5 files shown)


bryanlb (dismissed):
I'd suggest maybe a slightly more descriptive name for `KaldbSearcher`. Based on your PR description:
> Rename KaldbService to KaldbSearcher - This will serve as the query layer for local data search

Is there any way we can convey this information in the class name?

This isn't a blocking change, so feel free to ignore and merge if you prefer - or address in a later PR. 👍

bryanlb (approved):

PR #35, opened 2021-06-23
Add docker compose dependency setup

bryanlb:
Adds ability to boot all required dependencies via docker compose. Also provides a default config file and runconfig.

### DIFF
--- a/config/grafana/provisioning/datasources/automatic.yml
+++ b/config/grafana/provisioning/datasources/automatic.yml
@@ -1,0 +1,26 @@
+apiVersion: 1
+
+deleteDatasources:
+  - name: KalDB
+    orgId: 1
+
+datasources:
+  - name: KalDB
+    type: elasticsearch
+    access: proxy
+    url: "host.docker.internal:8080"
+    password:
+    user:
+    database: api_log
+    basicAuth: false
+    basicAuthUser:
+    basicAuthPassword:
+    withCredentials:
+    isDefault: true
+    jsonData:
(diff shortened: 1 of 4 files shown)


config/grafana/provisioning/datasources/automatic.yml
```diff
@@ -0,0 +1,26 @@
...
+    basicAuthPassword:
+    withCredentials:
+    isDefault: true
+    jsonData:
+      esVersion: 7.0.0
+      logLevelField: "loglevel"
+      logMessageField: "message"
+      maxConcurrentShardRequests: 5
+      timeField: "@timestamp"
+    version: 1
```
mansu, line 26: Should we add a new line at the end of these files?

docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3"
+services:
+  zookeeper:
+    image: 'wurstmeister/zookeeper:latest'
```
mansu, line 4: Let's pin these to specific versions so new releases don't break this setup.

Zk - 3.6
Kafka: 2.8
Grafana: 7/8

mansu:
The ability to test these components locally is awesome!

docker-compose.yml
bryanlb: Switched from `wurstmeister` to `bitnami` images as they had slightly newer versions, and much better tagging. Otherwise they should be effectively identical for our use case.

vthacker (approved):

PR #36, opened 2021-06-23
Add yaml config support

bryanlb:
Adds optional yaml config support, with environment variable substitution.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,0 +1,27 @@
+serverPort: 8080
+
+indexerConfig:
+  maxMessagesPerChunk: 100
+  maxBytesPerChunk: 100000
+  commitDurationSecs: 10
+  refreshDurationSecs: 11
+  staleDurationSecs: 7200
+  dataTransformer: ${INDEXER_DATA_TRANSFORMER:-api_log}
+  dataDirectory: "/tmp"
+
+kafkaConfig:
+  kafkaTopic: ${KAFKA_TOPIC:-test-topic}
+  kafkaTopicPartition: ${KAFKA_TOPIC_PARTITION:-0}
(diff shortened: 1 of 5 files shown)


config/config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  maxBytesPerChunk: 100000
+  commitDurationSecs: 10
+  refreshDurationSecs: 11
+  staleDurationSecs: 7200
+  dataTransformer: "api_log"
+  dataDirectory: "/tmp"
+
+kafkaConfig:
+  kafkaTopic: ${KAFKA_TOPIC:-test-topic}
+  kafkaTopicPartition: "0"
```
vthacker, line 14: Can we make kafkaTopic/kafkaTopicPartition/kafkaBootStrapServers/kafkaClientGroup/s3Bucket/dataTransformer

read an ENV var as well with the default that works best for local debugging ?

src/test/java/com/slack/kaldb/config/KaldbConfigTest.java
```diff
@@ -135,6 +136,53 @@ public void testParseKaldbConfigFile() throws IOException {
     assertThat(indexerConfig.getDataDirectory()).isEqualTo("/tmp");
   }
 
+  @Test
```
vthacker, line 139: Can we add one more test that sets an env var and then asserts if the value is set correctly?

This example ( https://stackoverflow.com/questions/8168884/how-to-test-code-dependent-on-environment-variables-using-junit ) ensures that we unset it after the test

I guess we can do it ourselves in try/finally block

bryanlb: Per our Slack conversation we'll leave this test out for now. We can revisit this if we decide on a test framework like Mockito or similar. I've added a comment to the test line in question explaining this decision as well.

vthacker (approved):

PR #37, opened 2021-06-24
query service

vthacker:
Add a strawman hack query service

### DIFF
--- a/src/main/java/com/slack/kaldb/server/Kaldb.java
+++ b/src/main/java/com/slack/kaldb/server/Kaldb.java
@@ -25,6 +25,7 @@ import io.micrometer.prometheus.PrometheusMeterRegistry;
 import io.netty.handler.codec.http.DefaultHttpHeaders;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.HashSet;
 import java.util.concurrent.CompletableFuture;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


src/test/java/com/slack/kaldb/server/KaldbIndexerTest.java
```diff
@@ -144,12 +110,20 @@ public void testIndexFromKafkaSearchViaGrpcSearchApi() throws Exception {
...
+            TEST_KAFKA_PARTITION,
+            KALDB_TEST_CLIENT,
+            TEST_S3_BUCKET,
+            8081);
     KaldbConfig.initFromConfigObject(kaldbCfg);
 
     // Create an indexer, an armeria server and register the grpc service.
     ServerBuilder sb = Server.builder();
-    sb.http(0);
+    sb.http(8080);
```
mansu, line 126: I think hard coding the port number causes the unit tests to be flaky when tests are run in parallel.  So, I think leaving the port to be 0 and letting the library auto-assign a port may be more reliable.

src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -36,53 +37,90 @@
...
+      // subclass of this class?
+      // TODO: Start in a background thread.
+      indexer.start();
+
+      // TODO: On CTRL-C shut down the process cleanly. Ensure no write locks in indexer. Guava
+      // ServiceManager?
+    }
+
+    if (roles.contains(READ_NODE_ROLE)) {
+      setupMetrics(readPromMeterRegistry);
```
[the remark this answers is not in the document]
vthacker: I think so? The node can just be a query or cache service node right?

src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -36,53 +37,90 @@
...
+      Server server = sb.build();
+
+      CompletableFuture<Void> serverFuture = server.start();
+      serverFuture.join();
+      LOG.info("Started indexer server on port: {}", serverPort);
+
+      // TODO: Instead of passing in the indexer, consider creating an interface or make indexer of
+      // subclass of this class?
+      // TODO: Start in a background thread.
+      indexer.start();
```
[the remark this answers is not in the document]
vthacker: indexer runs in a thread pool

src/test/java/com/slack/kaldb/server/KaldbQueryServiceTest.java
```diff
@@ -0,0 +1,153 @@
...
+    indexingServer = newIndexingServer(kaldbConfig);
+    indexingServer.start().join();
+
+    // Produce messages to kafka, so the indexer can consume them.
+    final LocalDateTime startTime = LocalDateTime.of(2020, 10, 1, 10, 10, 0);
+    produceMessagesToKafka(broker, startTime);
+    Thread.sleep(1000); // Wait for consumer to finish consumption and roll over chunk.
+    assertThat(getCount(MESSAGES_RECEIVED_COUNTER, indexerMetricsRegistry)).isEqualTo(100);
+
+    readServer = newReadServer(kaldbConfig);
```
mansu, line 76: newQueryServer?

src/test/java/com/slack/kaldb/server/KaldbQueryServiceTest.java
```diff
@@ -0,0 +1,153 @@
...
+    readServer.start().join();
+
+    // We want to query the indexing server
+    List<String> servers = new ArrayList<>();
+    servers.add("gproto+http://127.0.0.1:8080/");
+    KaldbQueryService.servers = servers;
+    readServiceStub =
```
mansu, line 83: I think it would also be awesome if we can assert some params on the search result.

mansu (approved):

PR #38, opened 2021-06-24
Add a snapshot store.

mansu:
Added a new package called core under metadata that contains the KaldbMetadataStore. This abstract base class is a wrapper class around ZkMetadataStore with additional logic to serialize and de-serialize the KaldbMetadata objects using KaldbMetadataSerializer class on a specific path. This modularity will help us create other metadata stores very easily.

There are derived classes from KalDbMetadataStore that adds various capabilities like creatable and updatable and also abstract the details between Ephemeral and Persistent nodes. We also add a generic listener API on the CacheableMetadataStore that hides the ZK complexities from it's consumers.

SnapshotStore is an instance of the KaldbMetadata store that supports all operations except updates.

Added unit tests for new code.


src/main/java/com/slack/kaldb/metadata/core/CorruptMetadataNodeException.java
```diff
@@ -0,0 +1,11 @@
+package com.slack.kaldb.metadata.core;
+
+public class CorruptMetadataNodeException extends RuntimeException {
```
bryanlb, line 3: We probably don't need a custom exception here - per Effective Java we should favor standard exceptions (#72)  as it makes the API easier to learn, has a smaller memory footprint, and reduces the time loading. An `IllegalStateException` likely makes the most sense here.

src/main/java/com/slack/kaldb/metadata/core/PersistentMutableMetadataStore.java
```diff
@@ -0,0 +1,46 @@
...
+
+  private final boolean updatable;
+
+  public PersistentMutableMetadataStore(
+      boolean shouldCache,
+      boolean updatable,
+      String storeFolder,
+      MetadataStore metadataStore,
+      MetadataSerializer<T> metadataSerializer,
+      Logger logger)
```
bryanlb, line 20: Why are we passing in the logger instead of initializing it in this class? This makes invoking logging inconsistent (ie, some cases you need to do `LOG.error()` vs `logger.error()`.

bryanlb (changes requested):

src/main/java/com/slack/kaldb/metadata/core/CacheableMetadataStore.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * This store adds a cache on top of the KaldbMetadataStore. Optionally, this class can disable
+ * caching by passing in a feature flag.
+ *
+ * <p>Instantiating an instance of this class creates the cache, adds a listener and starts the
+ * cache. The listener doesn't notify the watchers on cache initialization but fires on subsequent
+ * cache events. To close the cache, call the close method.
+ */
+public abstract class CacheableMetadataStore<T extends KaldbMetadata>
```
vthacker, line 21: this name is super confusing as we already have CachedMetadataStore.

I think we should rename CachedMetadataStore to something else ? CachedMetadataStoreImpl  can just be CachedZKNode or something?

CacheableMetadataStore and CachedMetadataStore is tough to distinguish otherwise IMO

mansu: Yes, I will prefix all the classes in zookeeper package with ZK. I haven't done that yet to keep the changes to a minimum. I can change it in a separate diff once this is merged.

src/main/java/com/slack/kaldb/metadata/core/CorruptMetadataNodeException.java
mansu: I created a new exception in this case to indicate a distinct application failure state. In this case, this exception is used to perform cache invalidation in CacheableMetdataStore if the data is malformed.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -0,0 +1,136 @@
...
+        "SnapshotStoreFolder can't be null or empty.");
+    checkNotNull(logger, "Logger can't be null or empty");
+
+    this.metadataStore = metadataStore;
+    this.storeFolder = storeFolder;
+    this.metadataSerializer = metadataSerializer;
+    this.logger = logger;
+
+    // Create the path to the store in ZK. However, since 2 different processes may create
+    // a path at the same time, ignore the exception if node already exists.
```
vthacker, line 52: I guess it's not clear by looking at this PR to me when can two threads call create in parallel for the same path

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  protected String getPath(String snapshotName) {
+    return ZKPaths.makePath(storeFolder, snapshotName);
+  }
+
+  // TODO: byte arrays every where.
+  @SuppressWarnings("UnstableApiUsage")
+  public ListenableFuture<T> get(String path) {
```
vthacker, line 66: can we not use the name `get` here? 

Since `get` means a blocking call when used in a CompletableFuture / ListenableFuture context this method name could be something like `getData` ?

src/main/java/com/slack/kaldb/metadata/core/PersistentMutableMetadataStore.java
mansu: All of the log stores would use the same base class. So, it would be very confusing to understand which class is taking an action if we used the same base class logger. We pass in the child class logger here it's obvious which store is logging making the change. So, all SnapshotStore api calls will be logged under it's name as opposed to KaldbMetadataStore, which is also called by SearchStore and other stores in the future.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -0,0 +1,136 @@
...
+    String nodePath = getPath(path);
+    Function<String, T> deserialize =
```
vthacker, line 68: `Guava's functional primitives can be replaced with Java API` - this is a warning generated from the google error code or the IDE

src/main/java/com/slack/kaldb/metadata/core/CorruptMetadataNodeException.java
bryanlb: Wouldn't an `InvalidStateException` do the same thing? We're never actually catch a `CorruptMetadataNodeException` - we just throw it and then catch the `RuntimeException`. 

 As a consumer of this class any `RuntimeException` would necessitate the invalidation of the metadata store.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -0,0 +1,136 @@
...
+        new Function<>() {
+          @Override
+          public @Nullable List<T> apply(@Nullable List<String> paths) {
+            if (paths == null) return Collections.emptyList();
+
+            List<ListenableFuture<T>> getFutures = new ArrayList<>(paths.size());
+            for (String path : paths) {
+              getFutures.add(get(path));
+            }
+            ListenableFuture<List<T>> response = Futures.successfulAsList(getFutures);
```
vthacker, line 119: what happens to the child nodes that failed? 

`successfulAsList` will skip it silently - is that what we want? Do we have a test case where this fails ever?

src/main/java/com/slack/kaldb/metadata/core/CacheableMetadataStore.java
vthacker: sounds good!

vthacker (approved):

src/main/java/com/slack/kaldb/metadata/core/CorruptMetadataNodeException.java
mansu: Good point. Replaced it with an IllegalStateException. I added it initially to do cache invalidation, but ended up not using it.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
mansu: 2 indexers can instantiate a snapshot store at the same time.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
mansu: Yeah, I tried replacing it with a Java function but the code became more complex to read due to type casting. So, leaving it in.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
mansu: Yes, for now a partial result would be more valuable than a failed data.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
mansu: changed to getNode.

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -0,0 +1,136 @@
...
+    // Create the path to the store in ZK. However, since 2 different processes may create
+    // a path at the same time, ignore the exception if node already exists.
+    try {
+      metadataStore.create(storeFolder, "", true).get();
+    } catch (NodeExistsException e) {
+      // ignore exception, since node creation is idemponent.
+    }
+  }
+
+  protected String getPath(String snapshotName) {
```
vthacker, line 60: `getPath` calls `makePath` ?

vthacker (approved):

src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
mansu: Yeah, this ZK function is not named correctly. It just constructs the path from the given params.

bryanlb (approved):

PR #39, opened 2021-06-24
Add a Cached store implementation

mansu:
Add a cached store implementation that implements ZK node caching.

The cache is not currently immune to ZK restarts and connection loss but we will fix it in a future diff.


bryanlb (dismissed):

src/main/java/com/slack/kaldb/metadata/zookeeper/CachedMetadataStoreImpl.java
```diff
@@ -0,0 +1,204 @@
...
+    // Create a curator cache but don't store any data in it since CacheStorage only allows
+    // storing data as a byte array. Instead use the curator cache implementation for
+    // managing persistent watchers and other admin tasks. Instead add a listener which would
+    // cache the data locally as a POJO using a serializer. In future, this also allows us to store
+    // the data in a custom data structure other than a hash table. Currently, if we lose a ZK
+    // connection the cache will grow stale but this class is oblivious of it.
+    // TODO: Add a mechanism to detect a stale cache indicate that a cache is stale.
+    cache =
+        CuratorCache.bridgeBuilder(curator, path)
+            .withExecutorService(executorService)
```
vthacker, line 94:
```
        if ( !forceTreeCache && Compatibility.hasPersistentWatchers() )
        {
            if ( executorService != null )
            {
                LoggerFactory.getLogger(getClass()).warn("CuratorCache does not support custom ExecutorService");
            }
            CuratorCacheStorage storage = cacheData ? CuratorCacheStorage.standard() : CuratorCacheStorage.dataNotCached();
            return new CuratorCacheImpl(client, storage, path, options, null);
        }
```

If you see the build code we don't need an executor service

vthacker: Why not instantiate CuratorCacheImpl directly since we know we will be using ZK 3.6+ ?

mansu: I preferred the bridge builder since the it is also backwards compatible with older versions of ZK also. Several recipes also use this pattern so it is widely supported and safe to do so.

vthacker (approved):

PR #40, opened 2021-06-24
Add snapshot metadata class and a serializer.

mansu:
Add a snapshot metadata class and a serializer/deserializer which can store metadata in ZK.

This is the first diff in a series of diffs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -153,6 +153,7 @@
             <version>${jackson.version}</version>
         </dependency>
 
+        <!-- YAML config parsing dependencies -->
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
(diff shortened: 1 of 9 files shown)


vthacker (approved):

pom.xml
```diff
@@ -153,24 +153,25 @@
             <version>${jackson.version}</version>
         </dependency>
 
+        <!-- YAML config parsing dependencies -->
```
bryanlb, line 156: Fwiw these comments tend to go stale pretty quickly as the project grows. For example - Jackson `core` and `databind` are now general serialization dependencies used for both json and yaml parsing. What will happen over time is folks will use a dependency we've already pulled in (ie, `commons-text`) and not realize there's an associated POM comment that is no longer valid.

bryanlb:
Can we get some of these commits squashed? There seems to be a lot of variations of "add test" without any additional detail.

vthacker:
> Can we get some of these commits squashed? There seems to be a lot of variations of "add test" without any additional detail.

image

I think we can do that while committing - It's the default option for me atleast

pom.xml
mansu: I agree this is true with code comments but for the dependencies it seemed to work well atleast so far. I am flexible either ways but these 2 dependencies are related to yaml, so I thought I call them out explicitly.

mansu:
Sure, I can squash and merge the commits.

bryanlb (approved):

PR #43, opened 2021-06-28
Move additional config options to ENV overrides

bryanlb:
We needed to override some additional `indexerConfig` options on a per-env basis, so this moves all the remaining config options to be set via env var.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,22 +1,22 @@
-serverPort: 8080
+serverPort: ${KALDB_SERVER_PORT:-8080}
 
 indexerConfig:
-  maxMessagesPerChunk: 100
-  maxBytesPerChunk: 100000
-  commitDurationSecs: 10
-  refreshDurationSecs: 11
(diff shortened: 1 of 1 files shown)


config/config.yaml
```diff
@@ -1,22 +1,22 @@
-serverPort: 8080
+serverPort: ${KALDB_SERVER_PORT:-8080}
 
 indexerConfig:
-  maxMessagesPerChunk: 100
-  maxBytesPerChunk: 100000
-  commitDurationSecs: 10
-  refreshDurationSecs: 11
-  staleDurationSecs: 7200
+  maxMessagesPerChunk: ${INDEXER_MAX_MESSAGES_PER_CHUNK:-100}
```
mansu, line 4: I think it's best to bump up this default to 500k. Otherwise, we may have chunk roll over exceptions?

mansu (approved):

vthacker (approved):

PR #46, opened 2021-06-30
add tests for nodeRoles

vthacker:
add tests for nodeRoles

### DIFF
--- a/src/main/java/com/slack/kaldb/config/KaldbConfig.java
+++ b/src/main/java/com/slack/kaldb/config/KaldbConfig.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.util.JsonFormat;
 import com.slack.kaldb.proto.config.KaldbConfigs;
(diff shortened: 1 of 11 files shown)


src/main/java/com/slack/kaldb/config/KaldbConfig.java
```diff
@@ -26,7 +27,19 @@
...
     KaldbConfigs.KaldbConfig.Builder kaldbConfigBuilder = KaldbConfigs.KaldbConfig.newBuilder();
     JsonFormat.parser().ignoringUnknownFields().merge(jsonStr, kaldbConfigBuilder);
-    return kaldbConfigBuilder.build();
+    KaldbConfigs.KaldbConfig kaldbConfig = kaldbConfigBuilder.build();
+    validateConfig(kaldbConfig);
+    return kaldbConfig;
+  }
+
+  public static void validateConfig(KaldbConfigs.KaldbConfig kaldbConfig) {
+    // We don't need further checks for node roles since JSON parsing will throw away roles not part
```
mansu, line 36: Let's use a pre-condition instead for this check. 
 https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html

Does the validation need to be here? I think it may be better to move it closer to main in KalDb.java.

src/test/java/com/slack/kaldb/testlib/KaldbConfigUtil.java
```diff
@@ -6,8 +6,8 @@
 
 public class KaldbConfigUtil {
   // Initialize kaldb config with empty for tests.
-  public static void initEmptyConfig() throws InvalidProtocolBufferException {
-    KaldbConfig.initFromJsonStr("{}");
+  public static void initEmptyIndexerConfig() throws InvalidProtocolBufferException {
```
mansu, line 9: Since the config file also has a query service config, can we add the query service to the roles also? That way, the function name also doesn't need to change.

vthacker: which config file? Any reason you don't want to change the name of the function?

src/main/java/com/slack/kaldb/config/KaldbConfig.java
vthacker: if we move it to kalDB how do you have any recommendations on how to unit test it?

Currently KaldbcConfigTest has a lot of tests - How will we migrate those to use Kaldb.java?

bryanlb (approved):

src/test/resources/test_config.json
```diff
@@ -1,4 +1,6 @@
 {
+  "nodeRoles": ["INDEX"],
```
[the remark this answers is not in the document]
vthacker: done

mansu (approved):

PR #47, opened 2021-06-30
Switch from LocalDateTime to Instant

bryanlb:
This fixes an issue where we were losing timezone information as compared to Elasticsearch. This also should make it easier to work with time in general, as we are guaranteed to have timezone information at all times.


src/test/java/com/slack/kaldb/logstore/search/StatsCollectorTest.java
```diff
@@ -1,21 +1,16 @@
 package com.slack.kaldb.logstore.search;
 
-import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.COMMITS_COUNTER;
-import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.MESSAGES_FAILED_COUNTER;
-import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.MESSAGES_RECEIVED_COUNTER;
-import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.REFRESHES_COUNTER;
+import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.*;
```
mansu, line 3: Can we please undo the wildcard imports? We avoid them since they make reading code harder.

mansu (commented):
Thanks for making this change! This change looks good.

src/test/java/com/slack/kaldb/logstore/search/StatsCollectorTest.java
bryanlb: Yeah, I've fixed this. The default behavior for IntelliJ is to optimize these apparently. 🤔

vthacker (approved):
+1 LGTM

The import formatting is not something in our control anyways

src/test/java/com/slack/kaldb/logstore/search/StatsCollectorTest.java
mansu: Thought so too. Thanks!

mansu (approved):

mansu:
test

PR #48, opened 2021-07-02
Add SearchResult from proto convesion 

vthacker:
Add SearchResult from proto convesion 

We need this for aggregation in the query service

### DIFF
--- a/src/main/java/com/slack/kaldb/logstore/LogMessage.java
+++ b/src/main/java/com/slack/kaldb/logstore/LogMessage.java
@@ -142,4 +142,17 @@ public class LogMessage extends LogWireMessage {
   public LogWireMessage toWireMessage() {
     return new LogWireMessage(getIndex(), getType(), id, source);
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    LogMessage that = (LogMessage) o;
+    return timeSinceEpochMilli == that.timeSinceEpochMilli;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(timeSinceEpochMilli);
+  }
 }
--- a/src/main/java/com/slack/kaldb/logstore/LogMessage.java
+++ b/src/main/java/com/slack/kaldb/logstore/LogMessage.java
@@ -1,8 +1,14 @@
 package com.slack.kaldb.logstore;
 
+import com.google.common.base.Objects;
 import java.time.*;
-import java.util.*;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
```diff
@@ -42,6 +65,7 @@ public static SearchQuery fromSearchRequest(KaldbSearch.SearchRequest searchRequ
     searchResultBuilder.setFailedNodes(searchResult.failedNodes);
     searchResultBuilder.setTotalNodes(searchResult.totalNodes);
     searchResultBuilder.setTotalSnapshots(searchResult.totalSnapshots);
+    searchResultBuilder.setSnapshotsWithReplicas(searchResult.snapshotsWithReplicas);
```
vthacker, line 68: was missing

src/main/java/com/slack/kaldb/logstore/LogMessage.java
```diff
@@ -142,4 +142,17 @@ public void addProperty(String key, Object value) {
   public LogWireMessage toWireMessage() {
     return new LogWireMessage(getIndex(), getType(), id, source);
   }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    LogMessage that = (LogMessage) o;
+    return timeSinceEpochMilli == that.timeSinceEpochMilli;
```
bryanlb, line 151: This doesn't seem right - we would expect to see multiple messages with identical timestamps.

bryanlb (changes requested):

src/main/java/com/slack/kaldb/logstore/LogMessage.java
vthacker: it does not! Unsure what else fields can I use to compare?

src/test/java/com/slack/kaldb/server/SearchResultTest.java
```diff
@@ -0,0 +1,54 @@
...
+    assertThat(protoSearchResult.getFailedNodes()).isEqualTo(1);
+    assertThat(protoSearchResult.getTotalNodes()).isEqualTo(5);
+    assertThat(protoSearchResult.getTotalSnapshots()).isEqualTo(7);
+    assertThat(protoSearchResult.getSnapshotsWithReplicas()).isEqualTo(7);
+    assertThat(protoSearchResult.getBucketsCount()).isEqualTo(1);
+
+    SearchResult<LogMessage> convertedSearchResult =
+        KaldbLocalSearcher.fromSearchResultProto(protoSearchResult);
+
+    assertThat(convertedSearchResult).isEqualTo(searchResult);
```
mansu, line 52: Can you please an additional test using a hashmap or a set to test the hashing logic?

src/main/java/com/slack/kaldb/logstore/LogMessage.java
bryanlb: Per our discussion, `_source` will be expected to contain a significant amount of unique attributes. We probably should ensure this map is the same across objects.

bryanlb (dismissed):

bryanlb (approved):

PR #49, opened 2021-07-03
multi node search

vthacker:
Extend the query service to search across multiple nodes


src/test/java/com/slack/kaldb/testlib/TestKafkaServer.java
```diff
@@ -30,9 +30,9 @@
   public static final int TEST_KAFKA_PARTITION = 0;
 
   // Create messages, format them into murron protobufs, write them to kafka
-  public static void produceMessagesToKafka(EphemeralKafkaBroker broker, Instant startTime)
+  public static void produceMessagesToKafka(
```
vthacker, line 33: changes needed to index into two servers

pom.xml
```diff
@@ -260,6 +260,18 @@
             <version>${http.core.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>futures-extra</artifactId>
+            <version>4.2.0</version>
+        </dependency>
+
+        <dependency>
```
vthacker, line 269: will be fixed when we commit PR #52

src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -128,12 +129,20 @@ public ChunkManager(
...
         "Created a chunk manager with prefix {} and dataDirectory {}",
         chunkDataPrefix,
         dataDirectory);
   }
 
+  /*
+     One day we will have to think about rate limiting/backpressure and we will revisit this so it could potentially reject threads if the pool is full
+  */
+  private static ExecutorService queryThreadPool() {
+    return Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
```
bryanlb, line 143: This probably would be better to be a configuration value instead of just using the runtime. The `availableProcessors` historically hasn't been super great at correctly detecting vCPUs. Additionally when deploying to different hardware configurations it can make detecting performance issues more difficult.

src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -263,18 +272,50 @@ public void rollOverActiveChunk() {
...
-    for (Chunk<T> chunk : chunkMap.values()) {
-      if (chunk.containsDataInTimeRange(
-          query.startTimeEpochMs / 1000, query.endTimeEpochMs / 1000)) {
-        LOG.debug("Searching chunk {}", chunk.id());
-        searchResults.add(chunk.query(query));
-      }
-    }
+  public CompletableFuture<SearchResult<T>> query(SearchQuery query) {
+
+    SearchResult<T> empty =
```
bryanlb, line 277: `SearchResult` should probably provide a no-argument constructor instead.

src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
```diff
@@ -38,33 +39,41 @@ public static SearchQuery fromSearchRequest(KaldbSearch.SearchRequest searchRequ
...
+          protoSearchResult.getTookMicros(),
+          protoSearchResult.getTotalCount(),
+          histogramBuckets,
+          protoSearchResult.getFailedNodes(),
+          protoSearchResult.getTotalNodes(),
+          protoSearchResult.getTotalSnapshots(),
+          protoSearchResult.getSnapshotsWithReplicas());
+    } catch (IOException e) {
+      // TODO
+      return new SearchResult<>(null, 1, 1, null, 0, 0, 0, 0);
```
bryanlb, line 67: Why don't we return null in this case? If I cannot successfully parse a protomessage I would expect either an exception, or a null to be returned. Or we could make this return an `Optional` of a search result.

src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
```diff
@@ -76,7 +85,11 @@ public static SearchQuery fromSearchRequest(KaldbSearch.SearchRequest searchRequ
     // Set hits
     ArrayList<String> protoHits = new ArrayList<>(searchResult.hits.size());
     for (T hit : searchResult.hits) {
-      protoHits.add(JsonUtil.writeAsString(hit));
+      try {
+        protoHits.add(JsonUtil.writeAsString(hit));
+      } catch (JsonProcessingException e) {
+        throw new RuntimeException(e);
```
bryanlb, line 91: Should a single failed `searchResult` conversion be a runtime exception? Should we look at a null / optional approach here instead?

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
```diff
@@ -15,26 +25,73 @@
 
   public static List<String> servers = new ArrayList<>();
 
-  /**
-   * TODO 1. Cache the stub 2. Run in parallel 3. Integrate with ZK to update list 4. Aggregate the
-   * responses instead of only returning the result from the first server
-   */
-  public KaldbSearch.SearchResult distributedSearch(KaldbSearch.SearchRequest request) {
+  // public so that we can override in tests
+  public static int READ_TIMEOUT_MS = 15000;
```
bryanlb, line 29: This should probably be a configuration option?

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
```diff
@@ -15,26 +25,73 @@
...
+
+  public static final KaldbSearch.SearchResult error =
```
bryanlb, line 31: Is this being used? It appears the same as the `timeOutError` below.

src/main/java/com/slack/kaldb/server/KaldbTimeoutLocalSearcher.java
```diff
@@ -0,0 +1,31 @@
+package com.slack.kaldb.server;
```
bryanlb, line 1: This should probably be part of the `tests`, not `main`

bryanlb (changes requested):

src/main/java/com/slack/kaldb/chunk/ChunkManager.java
vthacker: fixing this as part of #52

src/main/java/com/slack/kaldb/chunk/ChunkManager.java
vthacker: that makes sense! We could use `Runtime.getRuntime().availableProcessors()` as default if the the value isn't configured

But I really want to stop using a thread pool here altogether - i.e go away from the chunk manager aggregating locally to making 1 gRPC call per chunk - it makes it easier to handle errors etc

So I will do that shortly after these PRs

src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
vthacker: that is a fair point! I think if we move to the 1 gPRC call per chunk we want to throw error?

But if we don't then what you are suggesting is the right way forward

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
vthacker: good call! Will fix

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
vthacker: I'll fix it! I had it different while I was prototyping an approach where I could distinguish the two with an error message field but decided against it for now atleast

src/test/java/com/slack/kaldb/server/KaldbQueryServiceTest.java
```diff
@@ -54,109 +59,176 @@
...
+    if (indexingServer1 != null) {
+      indexingServer1.stop().join();
     }
-    if (indexingServer != null) {
-      indexingServer.stop().join();
+    if (indexerMetricsRegistry2 != null) {
+      indexerMetricsRegistry2.close();
+    }
+    if (indexingServer2 != null) {
+      indexingServer2.stop().join();
```
mansu, line 199: Can we replace the join calls here with get with timeouts?

src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
```diff
@@ -39,28 +39,32 @@ public static SearchQuery fromSearchRequest(KaldbSearch.SearchRequest searchRequ
   }
 
   public static SearchResult<LogMessage> fromSearchResultProto(
-      KaldbSearch.SearchResult protoSearchResult) throws IOException {
-    List<LogMessage> hits = new ArrayList<>(protoSearchResult.getHitsCount());
+      KaldbSearch.SearchResult protoSearchResult) {
```
mansu, line 42: I think it may be better to keep the original fromSearchResultProto method as is and add a wrapper method that performs this operation. That way we can save a refactoring if we need the underlying method elsewhere.

src/main/java/com/slack/kaldb/logstore/search/SearchResult.java
```diff
@@ -34,6 +35,19 @@ public SearchResult() {
...
+        searchResult.tookMicros,
+        searchResult.totalCount,
+        searchResult.buckets,
+        searchResult.failedNodes,
+        searchResult.totalNodes + 1,
+        searchResult.totalSnapshots,
+        searchResult.snapshotsWithReplicas);
+  }
+
   // TODO: Make search result a protobuf?
```
mansu, line 51: Drop this TODO since it's a proto?

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
mansu: +1 to config option. Also, let's change the default time to 30 seconds or so for a start since we don't want to timeout queries too early until the system is optimal.

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
```diff
@@ -15,26 +25,70 @@
...
+  // TODO: ChunkManager#QUERY_TIMEOUT_SECONDS and this could be unified?
+  public static int READ_TIMEOUT_MS = 15000;
+
+  public static final KaldbSearch.SearchResult error =
+      KaldbSearch.SearchResult.newBuilder().setFailedNodes(1).setTotalNodes(0).build();
+
+  // TODO Cache the stub
+  // TODO Integrate with ZK to update list of servers
+
+  public CompletableFuture<List<KaldbSearch.SearchResult>> distributedSearch(
```
mansu, line 39: make this method private since it is used locally?

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
```diff
@@ -15,26 +25,70 @@
...
 
   @Override
   public void search(
       KaldbSearch.SearchRequest request,
       StreamObserver<KaldbSearch.SearchResult> responseObserver) {
 
-    KaldbSearch.SearchResult protoSearchResult = distributedSearch(request);
-    responseObserver.onNext(protoSearchResult);
-    responseObserver.onCompleted();
+    CompletableFuture<List<SearchResult<LogMessage>>> searchResults =
```
mansu, line 66: Please add function doc explaining the logic in search and distributedSearch functionality. Why do we need to call fromSearchResultAndIncrementNodeCount? Isn't the local searcher already incrementing the node count?

src/main/java/com/slack/kaldb/server/KaldbLocalSearcher.java
vthacker: I'm unsure what you are suggesting - all the change does is instead of throwing the IOException it returns empty.

Are you saying that we should move the returning default to another method?

src/main/java/com/slack/kaldb/server/KaldbQueryService.java
vthacker: done!

mansu (approved):

bryanlb (approved):

PR #52, opened 2021-07-15
Async Query Code

vthacker:
Make query code for a single node async. We need this for multi-node search since everything needs to be non-blocking from botton up


src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -74,6 +73,9 @@
   private final long rolloverFutureTimeoutMs;
   private ListenableFuture<Boolean> rolloverFuture;
 
+  public static final int LOCAL_QUERY_THREAD_POOL_SIZE = 4;
```
bryanlb, line 76: Maybe add a todo here that we want to move this to the config eventually.

bryanlb (approved):
Looks good! Just an optional comment you may want to add in the followup PR.

PR #55, opened 2021-07-20
Improve test execution time & reliability

bryanlb:
This ended up being an assortment of improvements to our test execution:

* Added maven dependency caching per [Github guide](https://docs.github.com/en/actions/guides/building-and-testing-java-with-maven#caching-dependencies). This results in ~ 20% faster test builds for our current use case.
* Fixed a test race condition with `CountingFatalErrorHandler` - also made this threadsafe as depending on the curator listener configuration can run in a separate threadpool.
*  Swapped test implementation of completable futures from `join()` to `get()`. This allows us to throw a checked exception if they do not return within the specified time. The `KaldbQueryServiceTest` appears to be intermittently hanging until the build job times out, and this is an attempt to identify where this is occurring.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,6 +22,12 @@ jobs:
       with:
         java-version: '11'
         distribution: 'adopt'
+    - name: Cache Maven packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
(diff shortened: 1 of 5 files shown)


mansu (approved):
The changes look great! Thanks!

vthacker:
nice!

PR #57, opened 2021-07-22
Improve shutdown procedure for tests

bryanlb:
This should address several issues we have in our tests with the following classes due to leftover threads and/or race conditions.

* `ZookeeperMetadataStoreImplTest`
* `KaldbIndexerTest`
* `KaldbQueryServiceTest`
* `KaldbKafkaConsumerTest`

There still seems to be an issue with the `ZookeeperMetadataStoreImplTest` occasionally failing waiting for the `countingFatalErrorHandler` to be incremented, and the `KaldbQueryServiceTest` occasionally fails when the `queryServer` does not to respond to query or shutdown.

### DIFF
--- a/src/test/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImplTest.java
+++ b/src/test/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImplTest.java
@@ -13,10 +13,13 @@ import com.slack.kaldb.util.CountingFatalErrorHandler;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import org.apache.curator.retry.RetryNTimes;
 import org.apache.curator.test.TestingServer;
+import org.apache.zookeeper.ClientCnxn;
+import org.apache.zookeeper.ZooKeeper;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -26,6 +29,7 @@ public class ZookeeperMetadataStoreImplTest {
   private ZookeeperMetadataStoreImpl metadataStore;
   private MeterRegistry meterRegistry;
   private CountingFatalErrorHandler countingFatalErrorHandler;
+  private ZooKeeper zooKeeper;
 
   @Before
   public void setUp() throws Exception {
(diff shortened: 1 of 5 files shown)


src/test/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumerTest.java
```diff
@@ -101,11 +104,20 @@ public void kafkaConsumerStartupShutdown() throws Exception {
...
 
     @After
-    public void tearDown() {}
+    public void tearDown() throws ExecutionException, InterruptedException, TimeoutException {
+      // KafkaConsumer.DEFAULT_CLOSE_TIMEOUT_MS is 30 * 1000, so wait that plus a little extra
+      with()
+          .pollInterval(1, TimeUnit.SECONDS)
+          .await()
+          .atMost(45, TimeUnit.SECONDS)
+          .until(() -> testConsumer.isShutdown());
```
mansu, line 119: Could the threads be hanging since we forgot to shutdown the consumer properly?

Is there a reason the tearDown function here is different than the one on line 71? Can we use the same method here? Also, do we need to add this logic for the above test also?

src/test/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImplTest.java
```diff
@@ -496,5 +504,30 @@ public void testFatalErrorHandlerInvocation() throws Exception {
 
     // The FatalErrorHandler is incremented async in a separate thread
     await().until(() -> countingFatalErrorHandler.getCount() == 1);
+
+    // close the underlying zookeeper connection to ensure it's correctly removed
+    closeZookeeperClientConnection();
```
mansu, line 509: Should we add this method in tearDown instead of selectively calling in specific tests?

src/test/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumerTest.java
bryanlb: In the timeout test we intentionally close the `kafkaServer` before the consumer to replicate a timeout. We do request to close the consumer immediately after the server (`testConsumer.triggerShutdown()`), but this is not a synchronous call. 

This change ensures we wait until the consumer has completely closed before continuing onto the next test. The other test does not require this, as we close in expected order and it closes cleanly without queuing a reconnection.

src/test/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImplTest.java
bryanlb: I only added this to the two tests that simulate zookeeper timeouts using the "incorrect" shutdown order, as this is not the desired way to handle the shutdown. If we were to add this to all tests there's a chance that we would be masking shutdown issues given the forcible nature of this approach.

mansu (approved):

PR #58, opened 2021-07-23
Add a SearchMetadataStore

mansu:
Added a search metadata store to store the metadata needed for search.

This diff contains the code for metadata store also. We will review it once that diff is merged.


src/main/java/com/slack/kaldb/metadata/core/CorruptMetadataNodeException.java
```diff
@@ -0,0 +1,11 @@
+package com.slack.kaldb.metadata.core;
+
+public class CorruptMetadataNodeException extends RuntimeException {
```
bryanlb, line 3: We probably don't need a custom exception here - per Effective Java we should favor standard exceptions (#72)  as it makes the API easier to learn, has a smaller memory footprint, and reduces the time loading. An `IllegalStateException` likely makes the most sense here.

src/main/java/com/slack/kaldb/metadata/core/CacheableMetadataStore.java
```diff
@@ -0,0 +1,101 @@
...
+  private final Optional<CachedMetadataStore<T>> cache;
+
+  private final List<KaldbMetadataStoreChangeListener> watchers;
+
+  public CacheableMetadataStore(
+      boolean shouldCache,
+      String snapshotStoreFolder,
+      MetadataStore metadataStore,
+      MetadataSerializer<T> metadataSerializer,
+      Logger logger)
```
bryanlb, line 33: Why are we passing in the logger instead of initializing it in this class? This makes invoking logging inconsistent (ie, some cases you need to do `LOG.error()` vs `logger.error()`.

bryanlb (changes requested):

bryanlb (approved):

src/main/proto/metadata.proto
```diff
@@ -22,3 +22,14 @@ message SnapshotMetadata {
...
+
+message SearchMetadata {
+  // Name of search metadata
+  string name = 1;
+
+  // snapshot name
+  string snapshotName = 2;
+
+  // url
+  string url = 3;
```
vthacker, line 34: is `url` always going to be `s3Path` ?

mansu: It can be any URI. The first part of the URI is a scheme which contains prefix like s3, hdfs, file etc..

vthacker (approved):

src/main/proto/metadata.proto
vthacker: got it!

PR #60, opened 2021-07-30
Add Guava services, service manager

bryanlb:
Rewrites Armeria, KaldbIndexer, KaldbKafkaConsumer, and ChunkManager as Guava services.

At a high level, this moves initialization of all of these classes into `Kaldb.class`, and then registers them with the Guava `ServiceManager`. All of these services are "started" at the same time, and use `awaitRunning()` to defer their own startup until their dependencies are running. At shutdown if a service has a requirement that a dependency is shutdown first, it issues a `stopAync()` to the service and then `awaitTerminated()` until it is safe to continue.

We currently are using `AbstractIdleService` and `AbstractExecutionThreadService`, depending on the specific class in question. https://github.com/google/guava/wiki/ServiceExplained 

```
[INFO ] 2021-07-30 15:50:14.088 [KaldbKafkaWriter] KaldbKafkaConsumer - Starting kafka consumer.
[INFO ] 2021-07-30 15:50:14.091 [kaldbIndexerService STARTING] KaldbIndexer - Starting indexing into Kaldb.
[INFO ] 2021-07-30 15:50:14.091 [ChunkManager STARTING] ChunkManager - Starting chunk manager
[INFO ] 2021-07-30 15:50:14.433 [armeriaQueryService STARTING] ArmeriaService - Started on port: 8081
[INFO ] 2021-07-30 15:50:14.435 [armeriaIndexService STARTING] ArmeriaService - Started on port: 8080

..

[INFO ] 2021-07-30 15:51:44.176 [armeriaQueryService STOPPING] ArmeriaService - Shutting down
[INFO ] 2021-07-30 15:51:44.176 [kaldbIndexerService STOPPING] KaldbIndexer - Shutting down Kaldb indexer.
[INFO ] 2021-07-30 15:51:44.176 [kaldbIndexerService STOPPING] KaldbIndexer - Waiting for Kafka consumer to close.
[INFO ] 2021-07-30 15:51:44.177 [ChunkManager STOPPING] ChunkManager - Closing chunk manager.
[INFO ] 2021-07-30 15:51:44.177 [armeriaIndexService STOPPING] ArmeriaService - Shutting down
[INFO ] 2021-07-30 15:51:44.177 [ChunkManager STOPPING] ChunkManager - Roll over future completed successfully.
[INFO ] 2021-07-30 15:51:44.178 [ChunkManager STOPPING] ChunkManager - Closed chunk manager.
[INFO ] 2021-07-30 15:51:44.221 [kaldbIndexerService STOPPING] KaldbIndexer - Closed Kafka consumer cleanly
[INFO ] 2021-07-30 15:51:44.221 [kaldbIndexerService STOPPING] KaldbIndexer - Kaldb indexer is closed.
[INFO ] 2021-07-30 15:51:46.308 [Thread-0] Kaldb - Shutting down log4j2

```

### DIFF
--- a/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
+++ b/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
@@ -18,15 +18,16 @@ import com.slack.kaldb.elasticsearchApi.searchResponse.EsSearchResponse;
 import com.slack.kaldb.elasticsearchApi.searchResponse.HitsMetadata;
 import com.slack.kaldb.elasticsearchApi.searchResponse.SearchResponseHit;
 import com.slack.kaldb.elasticsearchApi.searchResponse.SearchResponseMetadata;
-import com.slack.kaldb.logstore.LogMessage;
-import com.slack.kaldb.logstore.search.KaldbLocalQueryService;
 import com.slack.kaldb.proto.service.KaldbSearch;
+import com.slack.kaldb.proto.service.KaldbServiceGrpc;
+import io.grpc.stub.StreamObserver;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * Elasticsearch compatible API service, for use in Grafana
@@ -39,10 +40,10 @@ import java.util.Optional;
     "OptionalUsedAsFieldOrParameterType") // Per https://armeria.dev/docs/server-annotated-service/
 public class ElasticsearchApiService {
 
-  private final KaldbLocalQueryService<LogMessage> searcher;
+  private final KaldbServiceGrpc.KaldbServiceImplBase searcher;
   private final ObjectMapper objectMapper = new ObjectMapper();
 
-  public ElasticsearchApiService(KaldbLocalQueryService<LogMessage> searcher) {
+  public ElasticsearchApiService(KaldbServiceGrpc.KaldbServiceImplBase searcher) {
     this.searcher = searcher;
   }
 
@@ -72,7 +73,30 @@ public class ElasticsearchApiService {
     KaldbSearch.SearchRequest searchRequest = request.toKaldbSearchRequest();
     // TODO remove join when we move to query service
     List<SearchResponseHit> responseHits = new ArrayList<>();
-    KaldbSearch.SearchResult searchResult = searcher.doSearch(searchRequest).join();
+
+    CompletableFuture<KaldbSearch.SearchResult> searchResultFuture = new CompletableFuture<>();
+    StreamObserver<KaldbSearch.SearchResult> responseObserver =
+        new StreamObserver<>() {
+          private KaldbSearch.SearchResult searchResult;
+
+          @Override
+          public void onNext(KaldbSearch.SearchResult searchResult) {
+            this.searchResult = searchResult;
+          }
+
+          @Override
+          public void onError(Throwable throwable) {
+            searchResultFuture.completeExceptionally(throwable);
+          }
+
+          @Override
+          public void onCompleted() {
+            searchResultFuture.complete(searchResult);
+          }
+        };
+
+    searcher.search(searchRequest, responseObserver);
+    KaldbSearch.SearchResult searchResult = searchResultFuture.join();
 
     List<ByteString> hitsByteList = searchResult.getHitsList().asByteStringList();
     for (ByteString bytes : hitsByteList) {
--- a/src/main/java/com/slack/kaldb/server/ArmeriaService.java
+++ b/src/main/java/com/slack/kaldb/server/ArmeriaService.java
@@ -1,0 +1,107 @@
+package com.slack.kaldb.server;
+
+import com.google.common.util.concurrent.AbstractIdleService;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.grpc.GrpcMeterIdPrefixFunction;
+import com.linecorp.armeria.common.logging.LogLevel;
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.armeria.server.docs.DocService;
+import com.linecorp.armeria.server.grpc.GrpcService;
+import com.linecorp.armeria.server.grpc.GrpcServiceBuilder;
+import com.linecorp.armeria.server.healthcheck.HealthCheckService;
+import com.linecorp.armeria.server.logging.LoggingService;
+import com.linecorp.armeria.server.logging.LoggingServiceBuilder;
+import com.linecorp.armeria.server.management.ManagementService;
+import com.linecorp.armeria.server.metric.MetricCollectingService;
+import com.slack.kaldb.elasticsearchApi.ElasticsearchApiService;
+import com.slack.kaldb.proto.service.KaldbServiceGrpc;
+import io.micrometer.prometheus.PrometheusMeterRegistry;
+import io.netty.handler.codec.http.DefaultHttpHeaders;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ArmeriaService extends AbstractIdleService {
+  private final Logger LOG = LoggerFactory.getLogger(ArmeriaService.class);
+
+  private final PrometheusMeterRegistry prometheusMeterRegistry;
+  private final KaldbServiceGrpc.KaldbServiceImplBase searcher;
+  private final String serviceName;
+  private final Server server;
+  private final int serverPort;
+
+  public ArmeriaService(
+      int serverPort,
+      PrometheusMeterRegistry prometheusMeterRegistry,
+      KaldbServiceGrpc.KaldbServiceImplBase searcher) {
+    this(serverPort, prometheusMeterRegistry, searcher, null);
+  }
+
+  public ArmeriaService(
+      int serverPort,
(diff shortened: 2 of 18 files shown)


src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -0,0 +1,107 @@
...
+  private final KaldbServiceGrpc.KaldbServiceImplBase searcher;
+  private final String serviceName;
+  private final Server server;
+  private final int serverPort;
+
+  public ArmeriaService(
+      int serverPort,
+      PrometheusMeterRegistry prometheusMeterRegistry,
+      KaldbServiceGrpc.KaldbServiceImplBase searcher) {
+    this(serverPort, prometheusMeterRegistry, searcher, null);
```
vthacker, line 39: instead of null can we pass `armeria` or something?

bryanlb: This constructor isn't used currently, and when the name is null, it defaults to the class name which would be `ArmeriaService`.

I allowed changing the name here, so the logs are clearer if it's indexer armeria service or query armeria service (if both are running concurrently).

src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -72,7 +73,30 @@ private EsSearchResponse doSearch(EsSearchRequest request) throws IOException {
     KaldbSearch.SearchRequest searchRequest = request.toKaldbSearchRequest();
     // TODO remove join when we move to query service
     List<SearchResponseHit> responseHits = new ArrayList<>();
-    KaldbSearch.SearchResult searchResult = searcher.doSearch(searchRequest).join();
+
+    CompletableFuture<KaldbSearch.SearchResult> searchResultFuture = new CompletableFuture<>();
```
vthacker, line 77: nice!

src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -72,7 +73,30 @@ private EsSearchResponse doSearch(EsSearchRequest request) throws IOException {
...
+          }
+
+          @Override
+          public void onCompleted() {
+            searchResultFuture.complete(searchResult);
+          }
+        };
+
+    searcher.search(searchRequest, responseObserver);
+    KaldbSearch.SearchResult searchResult = searchResultFuture.join();
```
vthacker, line 99: Can we avoid the `join` here and then make everything from here to transforming the results async?

Or is that not part of the scope here?

src/main/java/com/slack/kaldb/server/ArmeriaService.java
vthacker: got it! thanks for explaining

src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -0,0 +1,107 @@
...
+  protected void startUp() throws Exception {
+    CompletableFuture<Void> serverFuture = server.start();
+    serverFuture.get(15, TimeUnit.SECONDS);
+    LOG.info("Started on port: {}", serverPort);
+  }
+
+  @Override
+  protected void shutDown() throws Exception {
+    LOG.info("Shutting down");
+    server.closeAsync().get(15, TimeUnit.SECONDS);
```
vthacker, line 97: I think `stop` also returns a CompletableFuture

Looking at the armeria code for `stop` vs `closeAsync` I think `stop` does more? 

I definitely see more usages of stop - https://github.com/line/armeria/search?q=server.stop

Not a 100% sure what the subtle difference if at all b/w the two are

src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -50,104 +47,108 @@ public Kaldb(Path configFilePath) throws IOException {
...
-        .requestHeadersSanitizer((ctx, headers) -> DefaultHttpHeaders.EMPTY_HEADERS);
+  public static ServiceManager.Listener getServiceManagerListener() {
+    return new ServiceManager.Listener() {
+      @Override
+      public void failure(Service service) {
+        LOG.error(
+            String.format(
+                "Service %s failed with cause %s",
+                service.getClass().toString(), service.failureCause().toString()));
+        // shutdown if any services enters failure state
```
vthacker, line 117: Maybe another log line here saying `shutting down Kaldb` might be useful ? nbd

src/main/java/com/slack/kaldb/server/KaldbIndexer.java
```diff
@@ -167,7 +132,14 @@ public void close() {
       LOG.warn("Failed to close kafka consumer cleanly because of an exception.", e);
     }
 
-    chunkManager.close();
+    chunkManager.stopAsync();
```
vthacker, line 135: nice! So this is where guava helps with shutdown sequencing - i.e before the indexer shuts down first ensure kafka is shutdown and the chunk has been written out

vthacker: I guess we might face a regular pod shutdown issue and if it's during writing a chunk to S3 15 seconds here might not be enough? Something for later I guess

vthacker (approved):
+1

Changes look good

None of my comments are blocking mostly just curious if `stop` vs `closeAsync`

src/main/java/com/slack/kaldb/server/KaldbIndexer.java
bryanlb: Yeah, the timing is something we can definitely tweak if we need to - the 15s limit is mostly because I have the overall shutdown limited to 30s currently.

src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: This is a good question, and I had to do a bit more digging to find the differences between the two - https://github.com/line/armeria/issues/3598. 

`.stop()` is expected to be deprecated in favor of `.close()`/`.closeAsync()`. The current difference between the two seems to be that close cannot be restarted, where stopped can be.

src/main/java/com/slack/kaldb/server/Kaldb.java
bryanlb: Done!

src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
bryanlb: For this PR I just did what was required to support both indexer & query use cases, as the query classes are currently under active development and are expected to change a bit more.

This use case is also a bit unique as this is all part of an http request/response cycle, which is why I also didn't put a timeout when resolving this future. I'm not sure having this written async is a huge benefit, given we need to return an http response at the end.

src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -50,104 +47,109 @@ public Kaldb(Path configFilePath) throws IOException {
...
-        .requestHeadersSanitizer((ctx, headers) -> DefaultHttpHeaders.EMPTY_HEADERS);
+  public static ServiceManager.Listener getServiceManagerListener() {
+    return new ServiceManager.Listener() {
+      @Override
+      public void failure(Service service) {
+        LOG.error(
+            String.format(
+                "Service %s failed with cause %s",
+                service.getClass().toString(), service.failureCause().toString()));
+        LOG.error("Shutting down Kaldb due to failed service");
```
mansu, line 117: I think it may be better to use the RuntimeHalter class here for shutdown. That way we would have a uniform way to handle fatals.

src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -50,104 +47,109 @@ public Kaldb(Path configFilePath) throws IOException {
...
-  private void setupSystemMetrics() {
+  public static void addShutdownHook(ServiceManager serviceManager) {
+    Runtime.getRuntime()
+        .addShutdownHook(
+            new Thread(
+                () -> {
+                  try {
+                    serviceManager.stopAsync().awaitStopped(30, TimeUnit.SECONDS);
+
+                    // then shut down log4j
```
mansu, line 132: Please add a comment about log4j shutdown is done here so we can log all the server shutdown logs. That way, we can have the needed context.

src/test/java/com/slack/kaldb/chunk/ChunkCleanerTaskTest.java
```diff
@@ -29,15 +30,15 @@
   private ChunkManagerUtil<LogMessage> chunkManagerUtil;
 
   @Before
-  public void setUp() throws IOException {
+  public void setUp() throws IOException, TimeoutException {
```
mansu, line 33: Do we want to change the ChunkCleanerTask to use the service manager also?

mansu (commented):
This diff looks great! Couple of minor changes and this should be good to go.

src/test/java/com/slack/kaldb/chunk/ChunkCleanerTaskTest.java
bryanlb: I actually am already working on this in a followup PR here - https://github.com/slackhq/kaldb/pull/64. I figured I would make all the assorted fixes we discussed as part of a standalone PR.

src/main/java/com/slack/kaldb/server/Kaldb.java
bryanlb: Done!

src/main/java/com/slack/kaldb/server/Kaldb.java
bryanlb: Done!

mansu (approved):

vthacker (approved):

PR #61, opened 2021-07-30
Add JMH Dependencies

vthacker:
Add JMH Dependencies

To run it we can use

Compile -  `mvn package -DskipTests=true -f benchmarks/pom.xml `
 
Run -  `java -jar benchmarks/target/benchmarks.jar IndexingBenchmark`

Format - `mvn fmt:format -f benchmarks/pom.xml`

or run it from IntelliJ directly https://plugins.jetbrains.com/plugin/7529-jmh-java-microbenchmark-harness

Example output

```
Result "com.slack.kaldb.IndexingBenchmark.measureIndexing":
  106102.831 ±(99.9%) 8195.356 ops/s [Average]
  (min, avg, max) = (79329.269, 106102.831, 128929.146), stdev = 10940.566
  CI (99.9%): [97907.475, 114298.187] (assumes normal distribution)


# Run complete. Total time: 00:09:07

REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
experiments, perform baseline and negative tests that provide experimental control, make sure
the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
Do not assume the numbers tell you what you want them to tell.

Benchmark                           Mode  Cnt       Score      Error  Units
IndexingBenchmark.measureIndexing  thrpt   25  106102.831 ± 8195.356  ops/s

```

Pro tip - `java -jar benchmarks/target/benchmarks.jar -h` - is a goldmine of options - for example dumping results into JSON, profiling the test run etc

### DIFF
--- a/benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
+++ b/benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
@@ -1,0 +1,63 @@
+package com.slack.kaldb;
+
+import com.slack.kaldb.logstore.LogMessage;
+import com.slack.kaldb.logstore.LuceneIndexStoreImpl;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+import java.util.stream.Stream;
+import org.openjdk.jmh.annotations.*;
+
+@State(Scope.Thread)
+public class IndexingBenchmark {
+
+  private final Duration commitInterval = Duration.ofSeconds(5 * 60);
+  private final Duration refreshInterval = Duration.ofSeconds(5 * 60);
+  String timestamp;
+
+  private Path tempDirectory;
+  private MeterRegistry registry;
+  LuceneIndexStoreImpl logStore;
+  private Random random;
+
+  @Setup(Level.Iteration)
+  public void createIndexer() throws IOException {
+    random = new Random();
+    timestamp = Instant.now().toString();
+    registry = new SimpleMeterRegistry();
+    tempDirectory =
+        Files.createDirectories(
+            Paths.get("jmh", String.valueOf(random.nextInt(Integer.MAX_VALUE))));
+    logStore =
+        LuceneIndexStoreImpl.makeLogStore(
+            tempDirectory.toFile(), commitInterval, refreshInterval, registry);
+  }
+
+  @TearDown(Level.Iteration)
+  public void tearDown() throws IOException {
+    logStore.close();
+    try (Stream<Path> walk = Files.walk(tempDirectory)) {
+      walk.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+    }
+    registry.close();
+  }
+
+  @Benchmark
+  public void measureIndexing() {
+    Map<String, Object> fieldMap = new HashMap<>();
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -537,60 +555,5 @@
...
-                            <debugForkedProcess>true</debugForkedProcess>
-                        -->
-                    </configuration>
-                </plugin>
-                <plugin>
-                    <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.8.1</version>
-                    <configuration>
-                        <source>11</source>
```
vthacker, line 589: pluginManagement is for multi module project - for now we will not be looking at that scenario so simplify by removing. for example here we had hardcoded 11 instead of `javac.version` and keeping it in sync is easy to miss

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+  public void tearDown() throws IOException {
+    logStore.close();
+    try (Stream<Path> walk = Files.walk(tempDirectory)) {
+      walk.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+    }
+    registry.close();
+  }
+
+  @Benchmark
+  public void measureIndexing() {
```
vthacker, line 55: just a simple benchmark that adds one document. We will expand on them in the coming PRs

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+  private Random random;
+
+  @Setup(Level.Iteration)
+  public void createIndexer() throws IOException {
+    random = new Random();
+    registry = new SimpleMeterRegistry();
+    tempDirectory =
+        Files.createDirectories(
+            Paths.get("jmh", String.valueOf(random.nextInt(Integer.MAX_VALUE))));
+    MeterRegistry registry = new SimpleMeterRegistry();
```
bryanlb, line 39: Why do we re-assign `registry`?

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -0,0 +1,63 @@
...
+      walk.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+    }
+    registry.close();
+  }
+
+  @Benchmark
+  public void measureIndexing() {
+    Map<String, Object> fieldMap = new HashMap<>();
+    fieldMap.put(LogMessage.ReservedField.TIMESTAMP.fieldName, Instant.now().toString());
+    String message = "Log Message";
```
bryanlb, line 58: Why is this string assigned a variable but the other arguments are not?

bryanlb (commented):
Should we moving some of the fields in the `measureIndexing` to final values instead of instantiating them in the loop? I suspect that requesting the system time for each iteration is probably skewing the results slightly.

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
vthacker: maybe the values is useful per iteration? We could dump them in the tear down code - Unsure if it's gonna be useful so we can make this a static as well . WDYT?

vthacker:
> Should we moving some of the fields in the `measureIndexing` to final values instead of instantiating them in the loop? I suspect that requesting the system time for each iteration is probably skewing the results slightly.

For sure! This test is just a stub in how I see it - I wanted to make sure the dependencies in this pom.xml is correct so that we can reference the main Kaldb code 

I haven't paid any close attention to the test yet - I'll do that tomorrow

bryanlb (approved):

PR #69, opened 2021-08-04
Fix flaky KaldbDistributedQueryServiceTest

bryanlb:
Converts a sleep to an await, as during high CPU contention this can fail to complete within 1 second.


mansu (approved):

vthacker:
there are a few other places in our code that could also use this? `Thread.sleep(1000)` shows more places where we are still using sleeps

vthacker (approved):

PR #70, opened 2021-08-04
Improve TestKakfaServer reliability

bryanlb:
Our `KaldbKafkaConsumer` does not currently provide a Kafka config for `auto.offset.reset`, which means [it defaults](https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html) to `latest`. This means there exists a race condition in our tests if you attempt to produce messages to the broker prior to the consumer being online. The recent Guava services refactor removed a bunch of `Thread.sleep` calls in our tests, which made this race condition even more likely.

This code adds a method you can now use to see if the consumer is connected before attempting to produce messages.

### DIFF
--- a/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
+++ b/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
@@ -237,10 +237,10 @@ public class KaldbDistributedQueryServiceTest {
                 .setBucketCount(2)
                 .build());
 
-    assertThat(searchResponse.getHitsCount()).isEqualTo(100);
-    assertThat(searchResponse.getTotalCount()).isEqualTo(200);
     assertThat(searchResponse.getTotalNodes()).isEqualTo(2);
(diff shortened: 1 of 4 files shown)


src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
```diff
@@ -237,10 +237,10 @@ public void testSearch() {
                 .setBucketCount(2)
                 .build());
 
-    assertThat(searchResponse.getHitsCount()).isEqualTo(100);
-    assertThat(searchResponse.getTotalCount()).isEqualTo(200);
     assertThat(searchResponse.getTotalNodes()).isEqualTo(2);
     assertThat(searchResponse.getFailedNodes()).isEqualTo(0);
+    assertThat(searchResponse.getTotalCount()).isEqualTo(200);
```
vthacker, line 242: changing the order is to understand why KaldbDistributedQueryServiceTest fails ( which is not addressed in this PR as we're still trying to debug it )

vthacker (approved):

mansu (approved):

PR #76, opened 2021-08-06
Add metadata store config to kaldb config file.

mansu:
Add metadata store config to kaldb config file. The metadata store config includes zk config.


kaldb/src/main/proto/kaldb_configs.proto
```diff
@@ -15,7 +15,8 @@ message KaldbConfig {
     S3Config s3Config = 2;
     IndexerConfig indexerConfig = 3;
     QueryServiceConfig queryConfig = 4;
-    repeated NodeRole nodeRoles = 5;
+    MetadataStoreConfig metadataStoreConfig = 5;
```
vthacker, line 18: I guess no big deal right now since we don't care - but interchanging `5` to be something else is a breaking back-compat change

So rolling deploys will fail to serialize etc

vthacker (approved):

kaldb/src/main/proto/kaldb_configs.proto
mansu: I picked the order for now to make more sense from a config perspective and also didn't think it would matter. Will keep that in mind from next time.

bryanlb (approved):

PR #80, opened 2021-08-06
Use dedicated event loops for test worker

bryanlb:
This establishes dedicated thread pools for each of the Armeria servers, instead of relying on the default behavior which uses a static instance shared across all three.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
@@ -9,6 +9,7 @@ import static org.awaitility.Awaitility.await;
 import com.adobe.testing.s3mock.junit4.S3MockRule;
 import com.github.charithe.kafka.EphemeralKafkaBroker;
 import com.linecorp.armeria.client.Clients;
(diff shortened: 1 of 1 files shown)


kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
```diff
@@ -177,12 +178,17 @@ private static Server newIndexingServer(
       KaldbTimeoutLocalQueryService wrapperService =
           new KaldbTimeoutLocalQueryService(service, waitForSearchMs);
       return Server.builder()
+          .workerGroup(
+              EventLoopGroups.newEventLoopGroup(4, "armeria-common-worker-indexer-delayed", true),
```
vthacker, line 182: instead of 4 should we do the formula?

bryanlb: I'd prefer if we keep it a constant, so it behaves more consistently between CI and localhost.

vthacker: fair :) We've been bitten by it.. haha

vthacker (approved):

mansu (approved):

PR #85, opened 2021-08-09
Use named pool and better error handling

vthacker:
Use named pool and better error handling

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
@@ -4,12 +4,7 @@ import static com.slack.kaldb.util.ArgValidationUtils.ensureNonNullString;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.AbstractIdleService;
-import com.google.common.util.concurrent.FutureCallback;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -152,7 +147,9 @@ public ChunkManager(
      One day we will have to think about rate limiting/backpressure and we will revisit this so it could potentially reject threads if the pool is full
   */
   private static ExecutorService queryThreadPool() {
-    return Executors.newFixedThreadPool(LOCAL_QUERY_THREAD_POOL_SIZE);
+    return Executors.newFixedThreadPool(
```
vthacker, line 150: Bryan's fix!

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -301,11 +298,12 @@ public void rollOverActiveChunk() {
                     CompletableFuture.supplyAsync(() -> chunk.query(query), queryExecutorService)
                         // TODO: this will not cancel lucene query. Use ExitableDirectoryReader in
                         // the future and pass this timeout
-                        .completeOnTimeout(errorResult, QUERY_TIMEOUT_SECONDS, TimeUnit.SECONDS))
+                        .orTimeout(QUERY_TIMEOUT_SECONDS, TimeUnit.SECONDS))
```
vthacker, line 301: let this throw a timeout exception since we catch it in exceptionally - and now we will log it as well

mansu (dismissed):

mansu (approved):

bryanlb (approved):

PR #91, opened 2021-08-13
Add initital tracing integration

bryanlb:
Adds support for reporting Zipkin compatible traces, using Brave via the Armeria integration. Also adds a first-pass of instrumentation for the query path.

Also supports log4j2 integration:

> [INFO ] 2021-08-13 12:06:54.352 [armeria-common-worker-nio-2-2] **[trace=c93ded37a0cf9253/f168fe4bb68bc187]** Flags - com.linecorp.armeria.useOpenSsl: true (default)

Screen Shot 2021-08-13 at 3 11 09 PM
Screen Shot 2021-08-13 at 3 11 22 PM
Screen Shot 2021-08-13 at 3 11 38 PM

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
@@ -3,8 +3,11 @@ package com.slack.kaldb.chunk;
 import static com.slack.kaldb.util.ArgValidationUtils.ensureNonNullString;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
+import brave.ScopedSpan;
+import brave.Tracing;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.*;
+import com.linecorp.armeria.common.RequestContext;
 import com.slack.kaldb.blobfs.s3.S3BlobFs;
 import com.slack.kaldb.blobfs.s3.S3BlobFsConfig;
 import com.slack.kaldb.config.KaldbConfig;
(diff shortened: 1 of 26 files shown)


bryanlb:
Some details around the example traces for context:
* All spans are currently instrumented with the class + method name of where they're located
* `KaldbDistributedQueryService.distributedSearch` is currently instantiating a new GRPC stub for each call, which is why this is regularly taking excessive amounts of time. This also does not wrap the outbound call, as the method only sets up the GRPC stubs and returns a list of completable futures.
* I added instrumentation around the conversions to and from the proto objects, as we are interested how the serialization performance is. From my local testing, this seems to take between 50 - 100ms per request.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -296,7 +299,40 @@ public void rollOverActiveChunk() {
...
+                                  "dataEndTimeEpochSecs",
+                                  String.valueOf(chunk.info().getDataEndTimeEpochSecs()));
+                              span.tag(
+                                  "chunkSnapshotTimeEpochSecs",
+                                  String.valueOf(chunk.info().getChunkSnapshotTimeEpochSecs()));
+                              span.tag("numDocs", String.valueOf(chunk.info().getNumDocs()));
+                              span.tag("chunkSize", String.valueOf(chunk.info().getChunkSize()));
+                              span.tag("readOnly", String.valueOf(chunk.isReadOnly()));
+
+                              LOG.info("test");
```
vthacker, line 328: leftover from testing?

bryanlb: Whoops, indeed. Fixed.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkManager.java
```diff
@@ -296,7 +299,39 @@ public void rollOverActiveChunk() {
...
+                              span.tag("chunkSize", String.valueOf(chunk.info().getChunkSize()));
+                              span.tag("readOnly", String.valueOf(chunk.isReadOnly()));
+                              
+                              try {
+                                return chunk.query(query);
+                              } finally {
+                                span.finish();
+                              }
+                            },
+                            RequestContext.makeContextPropagating(queryExecutorService))
```
vthacker, line 334: I can't comment on the snippet below ( since nothing has changed ) but can we add a span in the exceptionally block ?

kaldb/src/main/resources/log4j2.xml
```diff
@@ -4,7 +4,7 @@
 
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
-            <PatternLayout  pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c{1} - %msg%n" />
+            <PatternLayout  pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %notEmpty{[trace=%X{traceId}/%X{spanId}]} %c{1} - %msg%n" />
```
vthacker, line 7: nice!

vthacker (dismissed):

vthacker (approved):

PR #94, opened 2021-08-18
Fix histogram initalization in SearchResultUtils

bryanlb:
Fixes an issue where histograms were not initialized with a value from the proto results, causing histograms to appear empty from the query server.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -27,7 +27,7 @@ s3Config:
   s3Bucket: ${S3_BUCKET:-test-s3-bucket}
 
 tracingConfig:
-  zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-http://localhost:9411/api/v2/spans}
+  zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-}
 
 queryConfig:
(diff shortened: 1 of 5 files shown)


config/config.yaml
```diff
@@ -27,7 +27,7 @@ s3Config:
   s3Bucket: ${S3_BUCKET:-test-s3-bucket}
 
 tracingConfig:
-  zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-http://localhost:9411/api/v2/spans}
+  zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-}
```
bryanlb, line 30: This will disable zipkin tracing by default for localhost

docker-compose.yml
```diff
@@ -40,18 +40,6 @@ services:
       GF_PATHS_PLUGINS: "/var/lib/grafana/plugins"
       GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS: "slack-kaldb-app,slack-kaldb-app-backend-datasource"
 
-  grafana7:
```
bryanlb, line 43: No longer used

mansu (approved):

vthacker (approved):

PR #96, opened 2021-08-19
Cache GRPC stub

bryanlb:
Puts the GRPC stub initialization in a loading cache to speed up query time

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -3,6 +3,9 @@ package com.slack.kaldb.logstore.search;
 import brave.ScopedSpan;
 import brave.Tracing;
 import brave.grpc.GrpcTracing;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.linecorp.armeria.client.Clients;
 import com.slack.kaldb.logstore.LogMessage;
 import com.slack.kaldb.proto.service.KaldbSearch;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -24,6 +27,17 @@
 
   public static List<String> servers = new ArrayList<>();
 
+  private final LoadingCache<String, KaldbServiceGrpc.KaldbServiceFutureStub> stubLoadingCache =
```
mansu, line 30: Does this map need data from ZK? If so, we need to initialize this field via a method.

bryanlb: This does not - it just provides a Function that is invoked when the loading cache has a miss. I refactored this slightly to make this clearer by breaking out the GRPC client initialization into a separate method.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -48,13 +58,13 @@
...
-      KaldbServiceGrpc.KaldbServiceFutureStub stub =
-          Clients.newClient(server, KaldbServiceGrpc.KaldbServiceFutureStub.class)
-              .withInterceptors(
-                  GrpcTracing.newBuilder(Tracing.current()).build().newClientInterceptor())
-              .withDeadlineAfter(READ_TIMEOUT_MS, TimeUnit.MILLISECONDS);
-      queryServers.add(ListenableFuturesExtra.toCompletableFuture(stub.search(request)));
+      queryServers.add(
+          ListenableFuturesExtra.toCompletableFuture(
+                  stubLoadingCache
+                      .getUnchecked(server)
```
vthacker, line 64: curious why we use `getUnchecked` and not `get` ?

vthacker (approved):
Pr looks good for now!

I guess we'll end up using https://curator.apache.org/curator-recipes/curator-cache.html instead of the Guava cache once we integrate with Zk

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
bryanlb: Per the doc block, `getUnchecked` should be used when checked exceptions are not thrown by the cache loader. Since `getKaldbServiceGrpcClient` doesn't throw a checked exception this should be safe for us to use.

PR #97, opened 2021-08-23
Cleaner `mvn package`

mansu:
Currently, the `mvn package` command emits a lot of warnings since there are duplicate MANIFEST files. In this diff, we do a first pass to  update the build files for a cleaner mvn package command. 

Also, replaced the deprecated javax.annotation package with jakarta.annotation package to clear some warnings.

To have a warning free mvn package step, we still need to resolve overlapping netty package issues.

### DIFF
--- a/benchmarks/pom.xml
+++ b/benchmarks/pom.xml
@@ -96,7 +96,21 @@
                                 </transformer>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
                             </transformers>
+
                             <filters>
+                                <filter>
+                                    <!--
+                                        Shading signed JARs will fail without this.
+                                        http://stackoverflow.com/questions/999489/invalid-signature-file-when-attempting-to-run-a-jar
+                                    -->
+                                    <artifact>*:*</artifact>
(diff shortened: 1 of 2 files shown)


benchmarks/pom.xml
```diff
@@ -96,15 +96,30 @@
...
+                                </filter>
                             </filters>
                         </configuration>
                     </execution>
                 </executions>
             </plugin>
             <plugin>
                 <artifactId>maven-deploy-plugin</artifactId>
-                <version>3.0.0-M1</version>
+                <version>2.8.1</version>
```
bryanlb, line 122: This should remain on `3.0.0-M1`, as we target Maven 3.6.3 in our build pipeline. The `2.x` releases are for Maven 2 (https://maven.apache.org/plugins/maven-deploy-plugin/, https://issues.apache.org/jira/browse/MNG-6169)

bryanlb (changes requested):

benchmarks/pom.xml
mansu: Removed maven deploy plugin since we don't use it.

bryanlb (approved):

PR #101, opened 2021-08-26
Add a MetadataStoreService.

mansu:
Add a MetadataStoreService and related Zk config.


kaldb/src/main/java/com/slack/kaldb/server/SnapshotMetadataStoreService.java
```diff
@@ -0,0 +1,31 @@
...
+  protected void startUp() throws Exception {
+    metadataStoreService.awaitRunning(15, TimeUnit.SECONDS);
+    snapshotStore =
+        new SnapshotMetadataStore(
+            metadataStoreService.getMetadataStore(), SNAPSHOT_METADATA_PATH, shouldCache);
+  }
+
+  @Override
+  protected void shutDown() throws Exception {
+    snapshotStore.close();
```
bryanlb, line 29: The `CacheableMetadataStore.close()` is currently non-reentrant, so that will need reworked in order to support this.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -72,11 +72,14 @@ public void setup() {
     serviceManager.startAsync();
   }
 
-  public static Set<Service> getServices() {
+  public static Set<Service> getServices() throws Exception {
```
bryanlb, line 75: This isn't needed, no exception is currently thrown.

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImpl.java
```diff
@@ -51,6 +54,20 @@
   public static final String METADATA_WRITE_COUNTER = "metadata.write";
   public static final String METADATA_READ_COUNTER = "metadata.read";
 
+  private static final int ZK_RETRY_COUNT = 3;
```
bryanlb, line 57: This should probably be part of the ZkConfig

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZookeeperMetadataStoreImpl.java
mansu: I left in as a hard coded value intentionally. If we failed thrice for a session reconnect, we usually terminate the VM, so this is fine for now.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
mansu: Fixed.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -51,7 +51,7 @@ public Kaldb(Path configFilePath) throws IOException {
     KaldbConfig.initFromFile(configFilePath);
   }
 
-  public static void main(String[] args) throws IOException {
+  public static void main(String[] args) throws Exception {
```
bryanlb, line 54: This can be reverted back to an `IOException` correct? I believe this was leftover from the previous exception change.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -61,7 +61,7 @@ public static void main(String[] args) throws IOException {
     kalDb.setup();
   }
 
-  public void setup() {
+  public void setup() throws Exception {
```
bryanlb, line 64: This no longer throws a checked exception.

bryanlb (changes requested):
Looks like there's still a few leftover exceptions from the previous version of the code.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
mansu: Yeah, but once I add the pending changes, this will be an Exception. So, left it in there. Reverted the exception.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
mansu: Fixed.

bryanlb (approved):

PR #107, opened 2021-09-02
Add gzip compression to responses

bryanlb:
Also switches response to `JSON_UTF_8`, as that is supported by the EncodingService by default.


vthacker:
is this faster?

bryanlb:
It is, reduces the size on the wire from 1.3Mb to ~ 150kb. Doesn't fix our underlying connectivity issue, but it does speed things up considerably. 

image

vthacker:
Hmm! the top 500 records + histogram data is 1.3MB ? 

Changes makes sense then!

vthacker (approved):

PR #110, opened 2021-09-07
Register Indexer in ChunkManager

mansu:
Register Indexer during ChunkManager initialization.

ChunkManager initialization is not the ideal place to register the indexer, but doing so in this diff since it involves quite a few changes in the code. Once this diff is committed, will move the indexer registration closer to the snapshot registration. This diff also helps us make parallel progress on distributed search.

I have also refactored the code so most unit tests don't have a dependency on `KaldbConfigUtil.initEmptyIndexerConfig` anymore. This results in cleaner seperation of code and initialization of tests via ChunkManagerUtil.


kaldb/src/main/java/com/slack/kaldb/chunk/SearchContext.java
```diff
@@ -0,0 +1,26 @@
...
+public class SearchContext {
+  public static SearchContext fromConfig(KaldbConfigs.ServerConfig serverConfig) {
+    return new SearchContext(serverConfig.getServerAddress(), serverConfig.getServerPort());
+  }
+
+  public final String hostname;
+  public final int port;
+
+  public SearchContext(String hostname, int port) {
+    checkState(hostname != null && !hostname.isEmpty(), "hostname field can't be null or empty");
```
bryanlb, line 16: These should be `checkArgument` instead of `checkState`

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
```diff
@@ -23,6 +23,9 @@
 public class KaldbConfig {
   public static Duration DEFAULT_START_STOP_DURATION = Duration.ofSeconds(15);
 
+  public static String SEARCH_METADATA_STORE_PATH = "/search";
```
bryanlb, line 26: Is there a reason we don't expose these as actual configs?

kaldb/src/main/java/com/slack/kaldb/metadata/core/CacheableMetadataStore.java
```diff
@@ -75,13 +75,10 @@ public void start() throws Exception {
     else cache.get().start();
   }
 
+  // Close is idempotent and also works when cache is disabled.
```
bryanlb, line 78: Ohh nice. I was waiting for this.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkManagerTest.java
```diff
@@ -67,24 +75,46 @@ public void setUp() throws InvalidProtocolBufferException {
...
+  public void tearDown() throws TimeoutException, IOException {
     metricsRegistry.close();
+    metadataStoreService.stopAsync();
+    metadataStoreService.awaitTerminated(DEFAULT_START_STOP_DURATION);
     if (chunkManager != null) {
       chunkManager.stopAsync();
       chunkManager.awaitTerminated(DEFAULT_START_STOP_DURATION);
     }
     s3Client.close();
+    if (localZkServer != null) {
```
bryanlb, line 105: I don't think we need this null check. The `setUp` guarantees that it has been initialized, and none of the other services that are initialized in `@Before` (`s3Client`, `metricsRegistry`, `metadataStoreService`) have null checks. We only are using null checks for `chunkManager` since that's actually initialized in the tests.

kaldb/src/test/java/com/slack/kaldb/testlib/ChunkManagerUtil.java
```diff
@@ -74,6 +108,11 @@ public void close() throws IOException, TimeoutException {
     if (s3Client != null) {
       s3Client.close();
     }
+    metadataStoreService.stopAsync();
+    metadataStoreService.awaitTerminated(DEFAULT_START_STOP_DURATION);
+    if (localZkServer != null) {
```
bryanlb, line 113: Shouldn't need the null check here either?

kaldb/src/test/java/com/slack/kaldb/testlib/KaldbConfigUtil.java
```diff
@@ -5,6 +5,7 @@
 import com.slack.kaldb.proto.config.KaldbConfigs;
 
 public class KaldbConfigUtil {
+  // TODO: Drop this method since we can no longer initialize the chunk manager using this.
```
bryanlb, line 8: Can we add a `@Deprecated` annotation to this? That will flag it in `IntelliJ`, so folks don't inadvertently add it to new tests before we can remove it.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
mansu: Since these are internal paths to KalDb, making these paths configurable is not very valuable. The zk path prefix is the only user exposed config at the moment.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkManagerTest.java
mansu: We employ null checks here since some unit tests kill terminate the ZK server and set this variable to null. It is a convention I adopted for tests, so the teardown code can know this to be the case.

kaldb/src/test/java/com/slack/kaldb/testlib/ChunkManagerUtil.java
mansu: We employ null checks here since some unit tests kill terminate the ZK server and set this variable to null. It is a convention I adopted for tests, so the teardown code can know this to be the case.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkManagerTest.java
bryanlb: For this and the following test however, the `localZkServer` variable is private,  never re-assigned to null, and never passed by reference - making the null check unnecessary.

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
bryanlb: Ah, I forgot these would be behind the prefix as well. 👍

kaldb/src/test/java/com/slack/kaldb/testlib/KaldbConfigUtil.java
mansu: good point. Done.

kaldb/src/main/java/com/slack/kaldb/chunk/SearchContext.java
mansu: Damn autocomplete. Done.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkManagerTest.java
mansu: That's true and we don't close ZK in these unit tests for now. However, that's a pattern we used in other tests. Since it's a simple null check, I would like to keep it so I maintain the pattern. Is there any downside of having these null checks?

bryanlb: There's a few - it's inconsistent with the other services in the same method (ie, `s3Client`, `metadataStoreService`). It also makes the code less readable - it's suggesting that the client can be set to null when it's impossible with the current code.

Even if the code were refactored at some point in the future such that it could be null - these are tests, so the impact of an NPE would be quick to identify and resolve.

mansu: Ok. Removed the checks now. I feel more checks don't hurt though.

bryanlb (approved):

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
bryanlb: These probably should be suffixed with `...STORE_ZK_PATH` instead of just `...STORE_PATH`.

https://github.com/slackhq/kaldb/pull/117#discussion_r710507760

mansu: Changed.

bryanlb (approved):

PR #111, opened 2021-09-07
Indexing Benchmark For API LOG

vthacker:


benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -40,6 +42,9 @@
   private LogMessage logMessage;
   private Document luceneDocument;
 
+  private BufferedReader reader;
+  private static SimpleDateFormat df = new SimpleDateFormat("yyyy-mm-ddHH:mm:ss.SSSzzz");
```
bryanlb, line 46: Seems like this might be something we'll (eventually) make configurable.

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -92,6 +100,64 @@ public void tearDown() throws IOException {
       walk.sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
     }
     registry.close();
+    if (reader != null) {
+      reader.close();
+    }
+  }
+
+  @Benchmark
+  public void measureAPILogIndexing() throws IOException {
```
mansu, line 109: Can you please move this benchmark into it's own file and a slack internal package? That way, it will be much easier to remove this code, once we no longer need it?

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -83,6 +88,9 @@ public void createIndexer() throws Exception {
     DocumentBuilder<LogMessage> documentBuilder = LogDocumentBuilderImpl.build(false);
 
     luceneDocument = documentBuilder.fromMessage(logMessage);
+
+    String apiLogFile = System.getProperty("jmh.api.log.file", "api_logs.txt");
```
bryanlb, line 92: Can we get an updated README for benchmarks with this? Maybe an example valid log line, and how to run with a specific file?

bryanlb (changes requested):

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
vthacker: Isn't it the same as removing this method and some instance variables?

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
vthacker: I guess so? The current murron output was this format so I didn't feel like it needs to be made configurable yet?

benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
mansu: True, but that is much harder than removing a class file.

vthacker: Also having it in the same file makes it easier for debug actions

bryanlb (dismissed):

mansu (approved):

PR #114, opened 2021-09-09
Fix ZK race condition with listener executions

bryanlb:
There currently exists a race condition with our metadata listeners, where events can still be executing after the metadata store has closed. This ensures that the listeners have exited prior to closing curator, so listeners that are attempting to read curator will not fail.


kaldb/src/test/java/com/slack/kaldb/metadata/zookeeper/CachedMetadataStoreImplTest.java
```diff
@@ -100,6 +109,67 @@ public void tearDown() throws IOException {
     return cachedMetadataStore;
   }
 
+  @Test
+  public void raceCondition() throws Exception {
```
mansu, line 113: Please use a more descriptive name for this test like testRaceConditionDuringClose.

mansu (approved):

kaldb/src/test/java/com/slack/kaldb/metadata/zookeeper/CachedMetadataStoreImplTest.java
bryanlb: Ah, I overlooked going back and fixing that. It's now `raceConditionOnMetadataStoreCloseListeners`

mansu (approved):

PR #116, opened 2021-09-13
Add a MetadataStoreService

bryanlb:
Reintroduces the `MetadataStoreService` originally proposed in #101, and rolled back in #115.

This includes fixes for the missing configs, updates the tests to assert that the correct curator strings are provided, and adds an integration test to assert that `KalDb` can successfully start and pass a simple healthcheck.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -36,3 +36,11 @@ queryConfig:
   serverConfig:
     serverPort: ${KALDB_QUERY_SERVER_PORT:-8081}
     serverAddress: ${KALDB_QUERY_SERVER_ADDRESS:-localhost}
+
+metadataStoreConfig:
+  zookeeperConfig:
+    zkConnectString: "127.0.0.1:2181"
+    zkPathPrefix: "kaldb"
+    zkSessionTimeoutMs: 1000
+    zkConnectionTimeoutMs: 1500
+    sleepBetweenRetriesMs: 500
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -39,8 +39,8 @@ queryConfig:
 
 metadataStoreConfig:
   zookeeperConfig:
-    zkConnectString: "127.0.0.1:2181"
-    zkPathPrefix: "kaldb"
-    zkSessionTimeoutMs: 1000
-    zkConnectionTimeoutMs: 1500
-    sleepBetweenRetriesMs: 500
+    zkConnectString: ${KALDB_ZK_CONNECTION_STRING:-localhost:2181}
+    zkPathPrefix: ${KALDB_ZK_PATH_PREFIX:-kaldb}
+    zkSessionTimeoutMs: ${KALDB_ZK_SESSION_TIMEOUT_MS:-5000}
+    zkConnectionTimeoutMs: ${KALDB_ZK_CONNECT_TIMEOUT_MS:-500}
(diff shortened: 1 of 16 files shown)


kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class KalDbIntegrationTest {
+  private static final Logger LOG = LoggerFactory.getLogger(KalDbIntegrationTest.class);
+
+  private Kaldb kaldb;
+  private final ObjectMapper om = new ObjectMapper();
+
+  @Before
+  public void start() throws IOException {
+    KaldbConfig.reset();
```
bryanlb, line 31: This is the same approach we use on `KaldbConfigTest.java`. Unfortunately we seem to have a significant amount of classes now that set state into `KaldbConfig.java` and don't cleanup on exit - including helper methods (ie, `KaldbConfigUtil.initEmptyIndexerConfig`). 

We may want to reconsider how we're making the config values available, as this will continue to be an issue.

config/config.yaml
```diff
@@ -36,3 +36,11 @@ queryConfig:
   serverConfig:
     serverPort: ${KALDB_QUERY_SERVER_PORT:-8081}
     serverAddress: ${KALDB_QUERY_SERVER_ADDRESS:-localhost}
+
+metadataStoreConfig:
+  zookeeperConfig:
+    zkConnectString: "127.0.0.1:2181"
+    zkPathPrefix: "kaldb"
+    zkSessionTimeoutMs: 1000
```
mansu, line 44: Ideal ZK session timeout values are: 
minSessionTimeout = ticktime * 2,  maxSessionTimeout=ticktime * 20.

Also, the session timeout should be higher than connection timeout. If we don't know the ZK tick time, a placeholder ZK session timeout to be 5 secs for now. Also, the connection timeout of 1.5 sec seems high. 500ms may be more appropriate with 100ms between reconnects.

kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
```diff
@@ -0,0 +1,84 @@
...
+    kaldb = new Kaldb(Path.of("../config/config.yaml"));
+    LOG.info("Starting kalDb with the resolved configs: {}", KaldbConfig.get().toString());
```
mansu, line 33: Should we move this statement into KaldbConfig when it gets initialized? It would be very useful to print the config when an application starts.

mansu (changes requested):

config/config.yaml
bryanlb: Originally I just copied over the example from the tests, but then I forgot to go back and edit these. I set the values you suggested, and made these all settable via ENV vars as well.

kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
bryanlb: The difficulty in putting this into the main application is ensuring that we don't print out any secrets into the logs. As we currently don't have a way to mark specific values as secret, we should keep this behavior to the tests.

mansu (approved):

PR #120, opened 2021-09-15
Add ZK, Kafka to integration tests

bryanlb:
This now starts ZK & Kafka prior to running the integration tests, so any initialization code that depends on these dependencies can properly boot.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
@@ -4,10 +4,14 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.charithe.kafka.EphemeralKafkaBroker;
 import com.slack.kaldb.config.KaldbConfig;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.HashMap;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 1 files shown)


kaldb/src/test/java/com/slack/kaldb/server/KalDbIntegrationTest.java
```diff
@@ -24,10 +28,16 @@
...
   private Kaldb kaldb;
+  private TestingServer testingServer;
+  private EphemeralKafkaBroker broker;
   private final ObjectMapper om = new ObjectMapper();
 
   @Before
-  public void start() throws IOException {
+  public void start() throws Exception {
+    testingServer = new TestingServer(2181);
+    broker = EphemeralKafkaBroker.create(9092);
```
mansu, line 38: Should we use random ports or do you think using standard ports is fine here?

bryanlb: These need to be the same as what is provided in the `config/config.yaml`, as that's where we're reading the connection strings from. If we want to use dynamic values we'd need to set the ENV vars, which I felt was less optimal than using the standard ports.

mansu: Oh, I meant would it be better to use a random port like 3333 instead of 2181. We can update the config also with this information. Not dynamic port numbers from tests.

bryanlb: Ah, in that case I think leaving them to the default ports is fine. If they're static there's not a significant advantage to changing them from the defaults.

mansu (approved):

PR #121, opened 2021-09-17
Add initial cache node registration

bryanlb:
Adds initial node type, run config, associated configs, and README updates.

Split from https://github.com/slackhq/kaldb/pull/117

PR #117, opened 2021-09-14
Add initial Cache node registration

bryanlb:
Adds Cache node configs, run config, and initial cache service registration.

PR #56, opened 2021-07-22
Single node async 

vthacker:
Make query code for a single node async. We need this for multi-node search since everything needs to be non-blocking from botton up

Attempt 2 where original PR was https://github.com/slackhq/kaldb/pull/52

PR #122, opened 2021-09-17
Refactor ChunkManager

bryanlb:
* Splits the existing `ChunkManager` into an abstract base class `ChunkManager` and two implementations - `IndexingChunkManager`, `CachingChunkManager`
* Stubs out functionality for the new `CachingChunkManager`, including a basic lifecycle management
* Fixes an issue with `KaldbMetadataStore` that prevented instantiating two metadata stores at the same path, and re-introduces the indexer registration from #110 that was disabled in #121 (f6974c3fbe570d04be1b6da801aa11b327932a41)


kaldb/src/test/java/com/slack/kaldb/chunk/ChunkCleanerTaskTest.java
```diff
@@ -30,13 +33,15 @@
 
   @Before
   public void setUp() throws Exception {
+    KaldbConfigUtil.initEmptyIndexerConfig();
```
bryanlb, line 36: This was removed in #110, but the test currently still requires this config to be initialized. 

It initially passed tests with this removed because another test that didn't cleanup its config was running before this one. With this refactor the tests are running in a slightly different order, causing this test to now fail.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+    LOG.info("Created a new read only chunk {}", chunkId);
   }
 
-  @Override
-  public void addMessage(T message) {
-    throw new ReadOnlyChunkInsertionException(
-        String.format("Chunk %s is a read only chunk", chunkInfo));
+  private KaldbMetadataStoreChangeListener cacheNodeListener() {
+    return () -> {
+      CacheSlotMetadata cacheSlotMetadata = cacheSlotMetadataStore.getCached().get(0);
```
vthacker, line 82: looks like everywhere we do a `get(0)` ? I guess that's because we have to conform to how the `CachedMetadataStore` hierarchy is?

So like every cache node gets it's znode and whenever the first entry  under it updates to evict/assigned etc we do something

looks fine i guess

vthacker: will there be N top level znodes for every chunk within a cache node?

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
...
+   * 1. topK results sorted by timestamp
+   * 2. histogram over a fixed time range
+   * We will not aggregate locally for future use-cases that have complex group by etc
+   */
+  public CompletableFuture<SearchResult<T>> query(SearchQuery query) {
+
+    SearchResult<T> errorResult =
+        new SearchResult<>(new ArrayList<>(), 0, 0, new ArrayList<>(), 0, 0, 1, 0);
+
+    List<CompletableFuture<SearchResult<T>>> queries =
```
vthacker, line 63: this might qualify for a long chain award in terms of LOC :) Anyways this was always there

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: The current design is such that the `CachedMetadataStoreImpl` caches all nested nodes. As we know there exist only a single nested node (`SLOT` in this case) we can safely just get the first item in the list. 

We have a todo to refactor the current metadata store constructors, such that we can initialize the store with data on the top znode, and not have to put the data on a nested node. That will simplify access to these znodes once that's refactored.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
bryanlb: https://github.com/slackhq/kaldb/pull/56/files#diff-cf8f5116a98917009ffaee10b4534d5695a235497100e64eab54d90937543d40R283 😄

kaldb/src/main/java/com/slack/kaldb/chunk/manager/caching/CachingChunkManager.java
```diff
@@ -0,0 +1,68 @@
...
+    this.metadataStoreService = metadataStoreService;
+    this.cacheConfig = cacheConfig;
+  }
+
+  @Override
+  protected void startUp() throws Exception {
+    LOG.info("Starting caching chunk manager");
+    metadataStoreService.awaitRunning(DEFAULT_START_STOP_DURATION);
+
+    for (int i = 0; i < cacheConfig.getSlotsPerInstance(); i++) {
```
bryanlb, line 37: I ended up refactoring this from the original [IntStream on the prior PR](https://github.com/slackhq/kaldb/pull/117/files#diff-0e217760e60aa380caf8feb80e8ab69c44231b5d36a6ebe627fe7d6822d5ee46R57), as the `ReadOnlyChunkImpl` can throw an exception on instantiation. In this specific case we likely would prefer to fail startup of the application instead of allowing some number of slots less than our target to instantiate, as this would never be retried.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ChunkManager<T> extends AbstractIdleService {
```
mansu, line 30: Please add a class doc for this class.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
...
+                    chunk.containsDataInTimeRange(
+                        query.startTimeEpochMs / 1000, query.endTimeEpochMs / 1000))
+            .map(
+                (chunk) ->
+                    CompletableFuture.supplyAsync(
+                            () -> {
+                              ScopedSpan span =
+                                  Tracing.currentTracer()
+                                      .startScopedSpan("ReadWriteChunkImpl.query");
+                              span.tag("chunkId", chunk.info().chunkId);
```
mansu, line 78: I don't think it is valuable to add all the chunk metadata to the tag. I think just adding the chunkId is fine. We can look up metadata store if needed for additional info and add tags as needed. For now, let's drop the rest of the tags. 

Is the query encoded in the trace any where?

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
...
+                              try {
+                                return chunk.query(query);
+                              } finally {
+                                span.finish();
+                              }
+                            },
+                            RequestContext.makeContextPropagating(queryExecutorService))
+                        // TODO: this will not cancel lucene query. Use ExitableDirectoryReader in
+                        // the future and pass this timeout
+                        .orTimeout(QUERY_TIMEOUT_SECONDS, TimeUnit.SECONDS))
```
mansu, line 107: On this timeout would we call span.finish? Other wise the span will be flushed after 30 seconds and the span duration will be off.

Also, prefer the duration variant if available.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/indexing/IndexingChunkManager.java
```diff
@@ -1,77 +1,56 @@
...
- * addMessage api is used by writers where as the query API is used by the readers.
- *
- * <p>Internally the chunk manager maintains a list of chunks. All chunks except one is considered
- * active. The chunk manager writes the message to the currently active chunk. Once a chunk reaches
- * a roll over point(defined by a roll over strategy), the current chunk is marked as read only. At
- * that point a new chunk is created which becomes the active chunk.
- */
-public class ChunkManager<T> extends AbstractIdleService {
-  private static final Logger LOG = LoggerFactory.getLogger(ChunkManager.class);
+public class IndexingChunkManager<T> extends ChunkManager<T> {
```
mansu, line 39: Please add a class doc for this class.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/caching/CachingChunkManager.java
```diff
@@ -0,0 +1,68 @@
...
+import com.slack.kaldb.logstore.LogMessage;
+import com.slack.kaldb.proto.config.KaldbConfigs;
+import com.slack.kaldb.server.MetadataStoreService;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.io.IOException;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CachingChunkManager<T> extends ChunkManager<T> {
```
mansu, line 16: Please add a class doc for this class.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
```diff
@@ -64,57 +61,4 @@ public void setUp() throws IOException {
   public void tearDown() {
     registry.close();
   }
-
```
mansu, line 67: I think it's best to wait until the next diff to merge this class along with it's tests. For now we can remove it.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
```diff
@@ -424,24 +421,24 @@ public void testSnapshotToS3UsingChunkApi() throws Exception {
...
-          new ReadOnlyChunkImpl<>(
-              newLocalFolderPath.getAbsoluteFile().toPath(),
-              new ChunkInfo("testDataSet2", 0),
-              registry);
-      SearchResult<LogMessage> newChunkResults = readOnlyChunk.query(searchQuery);
-      assertThat(newChunkResults.hits.size()).isEqualTo(1);
-      readOnlyChunk.close();
-      assertThat(FileUtils.listFiles(newLocalFolderPath, null, true).size())
-          .isEqualTo(s3Files.length);
+      //      File newLocalFolderPath = localDownloadFolder.newFolder();
```
mansu, line 426: Is there a reason to comment out this part of the unit test?

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbMetadataStoreTest.java
```diff
@@ -1393,4 +1394,70 @@ public void testCrudOperationsAndCache() throws ExecutionException, InterruptedE
       assertThat(deleteEx.getCause()).isInstanceOf(NoNodeException.class);
     }
   }
+
+  public static class TestAbstractMetadataStore {
```
mansu, line 1398: Good test!

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+    LOG.info("Closed chunk {}", chunkId);
+    cleanup();
   }
 
-  @Override
-  public boolean isReadOnly() {
-    return readOnly;
+  /** Deletes the log store data from local disk. */
+  public void cleanup() {
+    // TODO: Implement chunk state cleanup
```
mansu, line 183: I think we should implement this when we add the caching store.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/Chunk.java
```diff
@@ -1,76 +1,31 @@
...
+  /** Metadata about the loaded chunk if available, else null. */
   ChunkInfo info();
 
   /**
-   * A chunk will be available for writes initially. But once we no longer need to write any data to
-   * it can be turned into a read only chunk.
+   * Returns search results for the provided query. If no chunk data exists will return an empty
+   * result.
    */
-  boolean isReadOnly();
```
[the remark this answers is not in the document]
bryanlb: This method is still available on the `ReadWriteChunkImpl`, just not on the `Chunk` interface as this doesn't make sense for the `ReadOnlyChunkImpl`.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
```diff
@@ -157,40 +153,30 @@ public boolean snapshotToS3(String bucket, String prefix, S3BlobFs s3BlobFs) {
...
 
-  @Override
   public void postSnapshot() {
     LOG.info("Post snapshot operation completed for RW chunk {}", chunkInfo);
   }
 
-  /** Deletes the log store data from local disk. Should be called after close(). */
-  @Override
+  /** Deletes the log store data from local disk. */
   public void cleanup() {
```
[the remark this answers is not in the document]
bryanlb: This shouldn't be a separate command from close, as closing needs to automatically handle cleanup. Per the closeable spec it is required to release all system resources as part of invoking close.

I've refactored this so it's a bit clearer in the `close` method.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+  public void close() throws IOException {
+    if (logSearcher != null) {
+      logSearcher.close();
+    }
+
+    executorService.shutdown();
+    cacheSlotMetadataStore.close();
+
+    LOG.info("Closed chunk {}", chunkId);
+    cleanup();
```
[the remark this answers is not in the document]
bryanlb: Cleanup has to happen at the very end of the `ReadOnlyChunkImpl`, prior to exiting. We need to first shut down the logsearch, executor service, and deregister our node from ZK prior to deleting the data off of disk.

This ensures that we stop any new inbound queries, and stop any pending queries before removing the lucene index files.

kaldb/src/main/java/com/slack/kaldb/chunk/Chunk.java
```diff
@@ -1,76 +1,31 @@
...
 
-  /* Commit the data to disk */
-  void commit();
-
-  // Return a map of stats about the index.
-  // NOTE: With metrics baked in this may be redundant or return a metrics object?
-  // Map<String, Object> getStats();
-
   /** Close the chunk. */
   void close() throws IOException;
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
+package com.slack.kaldb.chunk.manager;
```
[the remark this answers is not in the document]
bryanlb: I ended up removing one layer of nesting, such that all `ChunkManager` related classes are in a package, and all `Chunk` related classes are in a package.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+    String slotId = String.format("%s-%s", serverAddress, chunkId);
+
+    String cacheSlotPath = String.format("%s/%s", CACHE_SLOT_STORE_ZK_PATH, slotId);
+    cacheSlotMetadataStore =
+        new CacheSlotMetadataStore(metadataStoreService.getMetadataStore(), cacheSlotPath, true);
+    cacheSlotMetadataStore.addListener(cacheNodeListener());
+
+    CacheSlotMetadata cacheSlotMetadata =
+        new CacheSlotMetadata(
+            METADATA_SLOT_NAME, Metadata.CacheSlotState.FREE, "", Instant.now().toEpochMilli());
```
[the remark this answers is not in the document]
bryanlb: Let's address the metadata store refactor in a separate diff, as this one is already getting quite large.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
bryanlb: These were added as part of #91, and were exceptionally useful when troubleshooting the performance issue we had with the Armeria threadpools.

You can see various screenshots on that PR of the tags we added to the traces. The chunkID is helpful when troubleshooting as that directly maps to the blob store directory.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: This was addressed as part of a fix above - https://github.com/slackhq/kaldb/pull/122#discussion_r715809724

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
bryanlb: Ah yes, I've now removed this. This test used the previous stub of `ReadOnlyChunkImpl`, which allowed instantiation of a `ReadOnlyChunkImpl` against a directory on disk. This is no longer possible as the `ReadOnlyChunkImpl` is now wired against ZK events.

All of the tests for verifying the `ReadOnlyChunkImpl` can hydrate / load / query assets from S3 were moved into the `ReadOnlyChunkImplTest` as part of #124 - (https://github.com/slackhq/kaldb/blob/6f578681ae137b90654ee10cdd47745fbc8e1d61/kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java)

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
bryanlb: Removed

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+      CacheSlotMetadata cacheSlotMetadata = cacheSlotMetadataStore.getCached().get(0);
+      LOG.debug("Change on chunk {} - {}", chunkId, cacheSlotMetadata.toString());
+
+      if (cacheSlotMetadata.cacheSlotState.equals(Metadata.CacheSlotState.ASSIGNED)) {
+        LOG.info(String.format("Chunk %s - ASSIGNED received", chunkId));
+        executorService.submit(
+            () -> {
+              try {
+                handleChunkAssignment();
+              } catch (Exception e) {
```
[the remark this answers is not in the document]
bryanlb: This has already been addressed as part of #124 - https://github.com/slackhq/kaldb/pull/124/files#diff-08ae1b7f6d010091ac19c82949597708014c27300d7dc2352525ba07e7011c24R150

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
   @Override
   public SearchResult<T> query(SearchQuery query) {
-    return logSearcher.search(
-        query.indexName,
-        query.queryStr,
-        query.startTimeEpochMs,
-        query.endTimeEpochMs,
-        query.howMany,
-        query.bucketCount);
+    if (logSearcher != null) {
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+  public ChunkInfo info() {
+    return chunkInfo;
   }
 
   @Override
-  @VisibleForTesting
-  public LogIndexSearcher<T> getLogSearcher() {
-    return logSearcher;
+  public boolean containsDataInTimeRange(long startTs, long endTs) {
+    if (chunkInfo != null) {
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+    this.executorService =
+        Executors.newSingleThreadExecutor(
+            new ThreadFactoryBuilder()
+                .setNameFormat(String.format("readonly-chunk-%s-%%d", chunkId))
+                .build());
+
+    String serverAddress = cacheConfig.getServerConfig().getServerAddress();
+    String slotId = String.format("%s-%s", serverAddress, chunkId);
+
+    String cacheSlotPath = String.format("%s/%s", CACHE_SLOT_STORE_ZK_PATH, slotId);
```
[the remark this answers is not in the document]
bryanlb: This has been addressed in #124 to use `SearchContext`.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -0,0 +1,166 @@
...
+import org.slf4j.LoggerFactory;
+
+public abstract class ChunkManager<T> extends AbstractIdleService {
+  private static final Logger LOG = LoggerFactory.getLogger(ChunkManager.class);
+
+  protected final Map<String, Chunk<T>> chunkMap = new ConcurrentHashMap<>();
+
+  // TODO: We want to move this to the config eventually
+  // Less than KaldbDistributedQueryService#READ_TIMEOUT_MS
+  public static final int QUERY_TIMEOUT_SECONDS = 10;
```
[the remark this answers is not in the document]
bryanlb: Let's do this as a separate PR, as this is existing code that wasn't changed in this diff. I don't want to expand the scope with other refactors as this is already quite large.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
+
+    LOG.info("Created a new read only chunk {}", chunkId);
   }
 
-  @Override
-  public void addMessage(T message) {
-    throw new ReadOnlyChunkInsertionException(
-        String.format("Chunk %s is a read only chunk", chunkInfo));
+  private KaldbMetadataStoreChangeListener cacheNodeListener() {
+    return () -> {
```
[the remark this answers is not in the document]
bryanlb: Let's discuss this more in #124 as that has a working implementation reacting to state transitions.

I think reacting on the state is significantly cleaner when considering our fail states - at any time the cluster manager can attempt an eviction; the cluster manager also can try to repair stuck nodes with the `NodeMonitorService`; and the individual node can detect fail states and attempt internal recovery as well.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -27,115 +37,169 @@
...
  * <p>TODO: Is chunk responsible for maintaining it's own metadata?
- *
- * <p>TODO: Move common operations between RO and RW chunk stores into a base class.
  */
 public class ReadOnlyChunkImpl<T> implements Chunk<T> {
 
   private static final Logger LOG = LoggerFactory.getLogger(ReadOnlyChunkImpl.class);
 
-  private final ChunkInfo chunkInfo;
+  private ChunkInfo chunkInfo;
```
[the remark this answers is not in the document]
bryanlb: Let's leave this in for this PR, as #124 is already good to go once this is merged.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
bryanlb: This was reworked per feedback here - https://github.com/slackhq/kaldb/pull/122#discussion_r715819975. 

The `CompletableFuture` does not take a duration as a timeout so this approach is required.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkRollOverException.java
```diff
@@ -1,7 +0,0 @@
-package com.slack.kaldb.chunk;
```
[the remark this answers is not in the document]
bryanlb: Split into separate PR #126 which is ready for review

kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/caching/CachingChunkManager.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/indexing/IndexingChunkManager.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/manager/IndexingChunkManager.java
```diff
@@ -1,77 +1,65 @@
...
   // TODO: ChunkDataPrefix can be moved to KaldbConfig?
   private final String chunkDataPrefix;
 
-  // ConcurrentHashMap here guarantees we will not have race conditions when updating and reading
-  // the chunkMap. There is a possibility when reading that the values will be stale, so care
-  // should be taken to ensure that the possibility of the chunk having already been removed is
-  // considered.
-  private final Map<String, Chunk<T>> chunkMap = new ConcurrentHashMap<>();
-
-  // TODO: Pass a reference to BlobFS instead of S3BlobFS.
```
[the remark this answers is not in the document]
bryanlb: Ah, good call. I had originally fixed this and then aborted due to the PR getting a bit out of hand.

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -52,7 +51,7 @@ public KaldbMetadataStore(
     // a path at the same time, ignore the exception if node already exists.
     try {
       metadataStore.create(storeFolder, "", true).get();
-    } catch (NodeExistsException e) {
+    } catch (ExecutionException ignored) {
```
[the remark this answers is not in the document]
bryanlb: Done.

PR #123, opened 2021-09-21
add trace benchmark

vthacker:


benchmarks/src/main/java/com/slack/kaldb/IndexTraceLogBenchmark.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.lucene.store.Directory;
+import org.openjdk.jmh.annotations.*;
+
+@State(Scope.Thread)
+public class IndexTraceLogBenchmark {
+
+  private static final String INDEX_NAME = "spans";
+
+  private Random random;
+  private final Duration commitInterval = Duration.ofSeconds(5 * 60);
```
bryanlb, line 34: `Duration.ofMinutes()`?

benchmarks/src/main/java/com/slack/kaldb/IndexTraceLogBenchmark.java
```diff
@@ -0,0 +1,185 @@
...
+  //        skipCount++;
+  //      }
+  //      System.out.println("Unable to parse " + skipCount + " lines");
+  //    }
+  //  }
+
+  @Setup(Level.Iteration)
+  public void createIndexer() throws Exception {
+
+    //        reader = Files.newBufferedReader(Path.of("trace_logs.txt"),
```
bryanlb, line 94: Left over?

benchmarks/src/main/java/com/slack/kaldb/IndexTraceLogBenchmark.java
```diff
@@ -0,0 +1,185 @@
...
+
+  @TearDown(Level.Iteration)
+  public void tearDown() throws IOException {
+    Directory directory = logStore.getIndexWriter().getDirectory();
+    String[] segmentFiles = directory.listAll();
+    long indexedBytes = 0;
+    for (String segmentFile : segmentFiles) {
+      indexedBytes += directory.fileLength(segmentFile);
+    }
+    System.out.println(
```
bryanlb, line 121: Are we using the SLF4J logger for the tests? If so we should probably use it here. If not, why do we have a [benchmark specific SLF4J config](https://github.com/slackhq/kaldb/blob/master/benchmarks/log4j2.xml)?

bryanlb (dismissed):

benchmarks/src/main/java/com/slack/kaldb/IndexTraceLogBenchmark.java
vthacker: very good point! haha I had forgotten

```
[INFO ] 2021-09-22 13:28:46.221 [com.slack.kaldb.IndexTraceLogBenchmark.measureTraceLogIndexingSlingshotMode-jmh-worker-1]  IndexTraceLogBenchmark - Indexed = 100 Skipped = 0 Index size = 47 KB

```

using the logger and output gets printed to console

benchmarks/src/main/java/com/slack/kaldb/IndexAPILog.java
```diff
@@ -69,15 +73,15 @@ public void tearDown() throws IOException {
...
+      LOG.info(
           "Indexed = "
               + indexCount
               + " Skipped = "
               + skipCount
               + " Index size = "
               + FileUtils.byteCountToDisplaySize(indexedBytes));
     } else {
-      System.out.println(
+      LOG.info((
```
bryanlb, line 84: Looks like this is causing compilation to fail

bryanlb (changes requested):
Looks like it's failing compilation

PR #125, opened 2021-09-24
Add basic cluster manager

bryanlb:
Adds basic cluster manager node type, first cluster manager service (ReplicaCreator).


bryanlb:
Moved to #127

PR #128, opened 2021-09-30
Add a unit test to understand field conflicts.

mansu:
A unit test to understand field conflicts and search in Kaldb.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
@@ -2,6 +2,7 @@ package com.slack.kaldb.logstore;
 
 import static com.slack.kaldb.testlib.TemporaryLogStoreAndSearcherRule.findAllMessages;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 import brave.Tracing;
 import com.slack.kaldb.testlib.MessageUtil;
@@ -67,13 +68,20 @@ public class FieldConflictsTest {
 
     strictLogStore.logStore.commit();
     strictLogStore.logStore.refresh();
-    Thread.sleep(1000);
 
     final String queryByHost = "hostname:host1-dc2.abc.com";
-    Collection<LogMessage> resultsByHost =
-        findAllMessages(
-            strictLogStore.logSearcher, MessageUtil.TEST_INDEX_NAME, queryByHost, 1000, 1);
-    assertThat(resultsByHost.size()).isEqualTo(2);
+    await()
+        .untilAsserted(
+            () ->
+                assertThat(
+                        findAllMessages(
+                                strictLogStore.logSearcher,
+                                MessageUtil.TEST_INDEX_NAME,
+                                queryByHost,
+                                1000,
+                                1)
+                            .size())
+                    .isEqualTo(2));
 
     final String conflictingTypeByNumber = conflictingFieldName + ":1";
     Collection<LogMessage> searchByInt =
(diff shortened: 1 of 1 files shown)


kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
```diff
@@ -0,0 +1,227 @@
...
+                "foo-bar",
+                LogMessage.ReservedField.HOSTNAME.fieldName,
+                "host1-dc2.abc.com",
+                conflictingFieldName,
+                1));
+    strictLogStore.logStore.addMessage(msg2);
+
+    strictLogStore.logStore.commit();
+    strictLogStore.logStore.refresh();
+    Thread.sleep(1000);
```
bryanlb, line 70: This should be converted to an `await()` - we have metrics on the log store that should be emitted once the commit and refresh are complete.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
```diff
@@ -0,0 +1,227 @@
...
+                "foo-bar",
+                LogMessage.ReservedField.HOSTNAME.fieldName,
+                "host1-dc2.abc.com",
+                conflictingFieldName,
+                200));
+    strictLogStore.logStore.addMessage(msg2);
+
+    strictLogStore.logStore.commit();
+    strictLogStore.logStore.refresh();
+    Thread.sleep(1000);
```
bryanlb, line 159: Same as above, switch to an `await()`

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
```diff
@@ -0,0 +1,227 @@
...
+            MessageUtil.TEST_INDEX_NAME,
+            conflictingTypeExactMatch,
+            1000,
+            1);
+    assertThat(searchByNumber.size()).isEqualTo(2);
+  }
+
+  @Test
+  public void testFieldConflictingFieldTypeWithDifferentValue() throws InterruptedException {
+    final String conflictingFieldName = "conflictingField";
```
bryanlb, line 101: We should use a different conflicting field names for each test. We don't reset the log store so there's a chance there's shared state across each of these tests. Asserts that are in the subsequent test can pass due to state initialized in the prior test.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
```diff
@@ -0,0 +1,227 @@
...
+  @BeforeClass
+  public static void beforeClass() throws Exception {
+    Tracing.newBuilder().build();
+  }
+
+  @Rule
+  public TemporaryLogStoreAndSearcherRule strictLogStore =
+      new TemporaryLogStoreAndSearcherRule(false);
+
+  public FieldConflictsTest() throws IOException {}
```
bryanlb, line 26: Unused constructor?

bryanlb (changes requested):

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
mansu: Done.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
mansu: Done.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
mansu: The @Rule annotation resets the store for every test run. Further, we check the total number of messages for each test, and we can see that the unit tests are cleared.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
mansu: Need since we throw a checked exception in @Rule.

kaldb/src/test/java/com/slack/kaldb/logstore/FieldConflictsTest.java
bryanlb: Ah, I misread this as a ClassRule. Carry on. 👍

bryanlb (approved):

PR #130, opened 2021-10-01
Improve ReadWriteChunkImpl close method

bryanlb:
Now with more graceful shutdown handling when failing to cleanup files

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
@@ -102,8 +102,14 @@ public class ReadWriteChunkImpl<T> implements Chunk<T> {
     logSearcher.close();
     logStore.close();
     LOG.info("Closed chunk {}", chunkInfo);
-    logStore.cleanup();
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java
```diff
@@ -102,8 +102,14 @@ public void close() throws IOException {
...
     LOG.info("Closed chunk {}", chunkInfo);
-    logStore.cleanup();
-    LOG.info("Cleaned up chunk {}", chunkInfo);
+
+    try {
+      logStore.cleanup();
+      LOG.info("Cleaned up chunk {}", chunkInfo);
+    } catch (Exception e) {
+      // this will allow the service to still close successfully when failing to cleanup the file
+      LOG.error("Failed to cleanup logstore for chunk {}", chunkInfo);
```
mansu, line 111: Should we also log the exception or is it logged already?

bryanlb: Yep, fixed. 👍

mansu (approved):

PR #131, opened 2021-10-05
Refactor ChunkInfo to support millisecond time granularity

bryanlb:


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -66,77 +66,84 @@ public void setChunkSize(long chunkSize) {
   // Size of the chunk
   private long chunkSize;
 
-  public ChunkInfo(String chunkId, long chunkCreationTimeEpochSecs) {
+  public ChunkInfo(String chunkId, long chunkCreationTimeEpochMs) {
```
mansu, line 69: There is a call site for this constructor that should also be changed: https://github.com/slackhq/kaldb/blob/master/kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunkImpl.java#L63

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -66,77 +66,84 @@ public void setChunkSize(long chunkSize) {
...
+    return chunkLastUpdatedTimeEpochMs;
   }
 
-  public void setChunkLastUpdatedTimeSecsEpochSecs(long chunkLastUpdatedTimeSecsEpochSecs) {
-    this.chunkLastUpdatedTimeSecsEpochSecs = chunkLastUpdatedTimeSecsEpochSecs;
+  public void setChunkLastUpdatedTimeEpochMs(long chunkLastUpdatedTimeEpochMs) {
+    this.chunkLastUpdatedTimeEpochMs = chunkLastUpdatedTimeEpochMs;
+  }
+
+  public void setDataStartTimeEpochMs(long dataStartTimeEpochMs) {
```
mansu, line 112: Can we drop these set methods since they are unused? Also, I think removing the fields also makes these values set only which may be an advantage.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
```diff
@@ -130,43 +130,36 @@ public void testAddAndSearchChunkInTimeRange() {
       assertThat(getCount(REFRESHES_COUNTER, registry)).isEqualTo(1);
       assertThat(getCount(COMMITS_COUNTER, registry)).isEqualTo(1);
 
-      final long expectedStartTimeEpochSecs = messageStartTime / 1000;
-      final long expectedEndTimeEpochSecs = expectedStartTimeEpochSecs + 99;
+      final long expectedEndTimeEpochMs = messageStartTime + (99 * 1000);
```
mansu, line 133: Please rename this variable messageStartTimeMs so the units are more obvious.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
```diff
@@ -130,43 +130,36 @@ public void testAddAndSearchChunkInTimeRange() {
...
-          (expectedStartTimeEpochSecs + 1) * 1000,
-          0);
+      searchChunk("Message100", messageStartTime, messageStartTime + 1000, 0);
 
       // Add more messages in other time range and search again with new time ranges.
       final List<LogMessage> newMessages =
           MessageUtil.makeMessagesWithTimeDifference(
               1, 100, 1000, startTime.plus(2, ChronoUnit.DAYS));
-      final long newMessageStartTimeEpochSecs = newMessages.get(0).timeSinceEpochMilli / 1000;
+      final long newMessageStartTimeEpochSecs = newMessages.get(0).timeSinceEpochMilli;
```
mansu, line 162: I think it's also better to rename this variable to use EpochMs as it's suffix since the calculation has changed.

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
bryanlb: Done.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
bryanlb: Done.

PR #132, opened 2021-10-06
Refactor ChunkManager to use list instead of map

bryanlb:


kaldb/src/main/java/com/slack/kaldb/chunk/manager/ChunkManager.java
```diff
@@ -34,7 +33,7 @@
 public abstract class ChunkManager<T> extends AbstractIdleService {
   private static final Logger LOG = LoggerFactory.getLogger(ChunkManager.class);
 
-  protected final Map<String, Chunk<T>> chunkMap = new ConcurrentHashMap<>();
+  protected final List<Chunk<T>> chunkList = new CopyOnWriteArrayList<>();
```
mansu, line 36: Please add a comment on why we use CopyOnWriteArrayList here. I think this choice will be useful when reading this code in the future.

bryanlb: Added

PR #134, opened 2021-10-08
Updated CachedMetadataImpl to work for single node.

mansu:
Renamed `CachedMetadataImpl` to `ZooKeeperCachedMetadataImpl` to clarify that this is a ZK class and not a class in Kaldb metadata. 

Rewrote the CachedMetadataImpl class to also work with caching a node and its children. This was a bigger change than expected since we need to a different curator API to get access both parent and child node notifications.

I also fixed a couple of exceptions and a bug in this diff. 

Currently, when we create the cache, we already expect the node to be present. If not, the node is created with an empty data. However, with the new notification API, this causes a protobuf parse exception, since the new node is created with an empty string. So, I updated all the unit tests to create a node first. If needed, we can update the API with the initial value of the node, but none of our use cases need that right now.

Currently, we pass in executor service to the curator cache builder but that is a deprecated API that results in a warning. So, I removed the code that passes in executor service to the cache which simplifies the code. This makes our code more dependent on ZK 3.6.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZookeeperCachedMetadataStoreImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZookeeperCachedMetadataStoreImpl.java
@@ -227,6 +227,8 @@ public class ZookeeperCachedMetadataStoreImpl<T extends KaldbMetadata>
   }
 
   private void maybeNotify() {
+    // TODO: getCount when used this way could return a stale value and we may have a race
+    //  condition here. Update this code to not use getCount.
     if (initializedLatch.getCount() == 0) {
       listenerContainer.forEach(
           listener -> {


kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbMetadataStoreTest.java
```diff
@@ -33,6 +35,12 @@
 
 @RunWith(Enclosed.class)
 public class KaldbMetadataStoreTest {
+
+  private static final String SNAPSHOTS_PATH = "/snapshots";
```
bryanlb, line 39: This should probably reuse the config value of `SNAPSHOT_METADATA_STORE_ZK_PATH` instead of redeclaring it here with the same value.

kaldb/src/test/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImplTest.java
```diff
@@ -31,18 +32,26 @@
...
   private ZooKeeper zooKeeper;
+  private int expectedCacheErrorCount = 0;
+
+  public ZooKeeperCachedMetadataStoreImplTest() throws InvalidProtocolBufferException {
+    defaultRootStr = serDe.toJsonStr(defaultRoot);
+  }
 
   static class CountingCachedMetadataListener implements CachedMetadataStoreListener {
     private int cacheChangedCounter = 0;
+    // TODO: Add a test with state changed counter.
```
bryanlb, line 54: Is there a reason we're not including this test now?

kaldb/src/test/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImplTest.java
```diff
@@ -368,62 +382,76 @@ public void stateChanged(CuratorFramework client, ConnectionState newState) {
...
-    assertThat(cache.getInstances()).isEmpty();
+    assertThat(cache.getInstances()).containsOnly(defaultRoot);
 
     cache.close();
+    await()
+        .untilAsserted(
+            () ->
+                assertThat(((ZooKeeperCachedMetadataStoreImpl<SnapshotMetadata>) cache).isStopped())
+                    .isTrue());
+    // TODO: Get instances after cache close works?
```
bryanlb, line 428: This todo can be removed, as this behavior is was intentional - see `ZooKeeperCachedMetadataStoreImplTest.watchCachePersistentTreeTest()`

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImpl.java
```diff
@@ -25,32 +22,30 @@
...
-import org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent;
 import org.apache.curator.utils.CloseableUtils;
 import org.apache.curator.utils.ZKPaths;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * A CachedMetadataStoreImpl uses a curator path cache to cache all the nodes under a given node. In
- * addition, this class also accepts a metadata serializer/de-serializer objects, so we only
+ * A CachedMetadataStoreImpl uses a curator path cache to cache a node and all the child nodes under
```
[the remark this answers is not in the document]
mansu: Done.

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbMetadataStoreTest.java
mansu: This code was written before those variables are defined. Agree on reusing. Updated the code.

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImpl.java
```diff
@@ -254,7 +221,30 @@ public boolean isStarted() {
...
   }
 
-  private void initialized() {
+  @VisibleForTesting
+  public boolean isStopped() {
+    return state.get().equals(State.STOPPED);
+  }
+
+  private void mayBeNotify() {
+    if (initializedLatch.getCount() == 0) {
```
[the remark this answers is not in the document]
mansu: Just looked at this code to use AtomicBoolean and an initializedLatch is better since we `await` in `start` method. In this case, while not ideal, using `getCount` is ok since we only wait till the cache is initialized and it's always true after that. 

Also, this is existing code, so resolving it for now. If needed we can do this in a separate diff.

kaldb/src/test/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImplTest.java
mansu: I added the TODO assuming we were missing the test. After digging in I realized that this method is not called in current implementation.I have a diff to update the interface, but then realized having this interface now is cheaper than adding it later. So leaving it in for now. If you feel strongly, I can remove it for now.

mansu:
Diff is ready for re-review.

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImpl.java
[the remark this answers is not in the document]
bryanlb: Let's add a todo that this should be fixed. 

As it exists now there is a race condition where we can miss firing the `cacheChanged` event. Since the `countDown()` is being executed in a different thread, `getCount()` can return a stale value which would result in failing to notify listeners.

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/ZooKeeperCachedMetadataStoreImpl.java
```diff
@@ -254,7 +221,30 @@ public boolean isStarted() {
     return state.get().equals(State.STARTED);
   }
 
-  private void initialized() {
+  @VisibleForTesting
+  public boolean isStopped() {
+    return state.get().equals(State.STOPPED);
+  }
+
+  private void mayBeNotify() {
```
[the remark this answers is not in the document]
mansu: I may have lost this change during a merge. Fixed it now.

bryanlb (approved):

PR #137, opened 2021-10-12
Add sync version of calls for KaldbMetadataStore API.

mansu:
Currently the KaldbMetadataStore API is asynchronous. Add a sync version of these API so we can enforce a uniform timeout everywhere and make the code cleaner.


kaldb/src/main/java/com/slack/kaldb/metadata/core/EphemeralMutableMetadataStore.java
```diff
@@ -35,6 +39,11 @@ public EphemeralMutableMetadataStore(
     }
   }
 
+  public Object createSync(T metadataNode)
+      throws ExecutionException, InterruptedException, TimeoutException {
```
bryanlb, line 43: What if these returned an `Optional<T>?` That would allow us to avoid casting the returned object and individually handling these exceptions.

kaldb/src/main/java/com/slack/kaldb/metadata/core/EphemeralMutableMetadataStore.java
```diff
@@ -35,11 +41,29 @@ public EphemeralMutableMetadataStore(
...
+
   public ListenableFuture<?> update(T metadataNode) {
     if (!updatable) {
       throw new UnsupportedOperationException("Can't update store at path " + storeFolder);
     }
 
     return super.update(metadataNode);
   }
+
+  public void updateSync(T metadataNode) {
```
bryanlb, line 61: Let's convert these to optionals, as this API design ends up being pretty cumbersome to use for common use cases. Here's a few examples, first using the current implementation and then if it were using optionals:

#### Try to set a value, and if it fails use a default value
```java
SnapshotMetadata result1 = null;
try {
  store.createSync(snapshot1);
  result1 = snapshot1;
} catch (InternalMetadataStoreException internalMetadataStoreException) {
  result1 = DEFAULT_VALUE;
}
```

```java
SnapshotMetadata result2 = store.createSyncOptional(snapshot1).orElse(DEFAULT_VALUE);
```

#### Try to set a value with two attempts tracking if it is successful
```java
int retryCount = 0;
int maxRetries = 2;
boolean success = false;
while(retryCount < maxRetries && !success) {
  try {
    store.createSync(snapshot1);
    success = true;
  } catch (InternalMetadataStoreException internalMetadataStoreException) {
    retryCount++;
  }
}
```

```
boolean success = store.createSyncOptional(snapshot1)
    .or(() -> store.createSyncOptional(snapshot1))
    .isPresent();
```

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/metadata/core/EphemeralMutableMetadataStore.java
mansu: I think we should the API as is for now for the following reasons:

* There is a built in curator retry for all of the operations. So, we don't need to retry any of these operations. If an operation fails, it generally means ZK is unavailable in which case, we currently can't do any further operations and terminate the process.
* An Optional API for a method that doesn't return a value is confusing 
* An optional API hides the underlying error from the caller.
* If we need method chaining, we can also use the async API.

I think we should try using the sync API as is for now and see how it works before making the semantics more complex. I used this API in unit tests and in index registration and quite like the fire and forget nature of it.

kaldb/src/main/java/com/slack/kaldb/metadata/core/EphemeralMutableMetadataStore.java
```diff
@@ -35,11 +41,29 @@ public EphemeralMutableMetadataStore(
     }
   }
 
+  public void createSync(T metadataNode) {
```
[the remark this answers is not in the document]
mansu: Done.

PR #138, opened 2021-10-18
Index registration

mansu:
Register the indexer with snapshot store and search metadata store in Kaldb. 

Until now, we don't pass in all the parameters needed to construct a proper snapshot into Kaldb.  So, as part of this diff, we pass in those arguments also. 

Added new unit tests where needed and updated existing tests.


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -16,16 +16,46 @@
  * code into multiple, classes.
  *
  * <p>TODO: Have a read only chunk info for read only chunks so we don't accidentally update it.
- * TODO: Add a state machine for a chunk?
+ *
+ * <p>TODO: Add a minOffset field to chunk and snapshot metadata.
  */
 public class ChunkInfo {
+  public static final long MAX_FUTURE_TIME = Instant.ofEpochSecond(253402329599L).toEpochMilli();
```
bryanlb, line 23: Let's use `Instant.MAX` here instead.

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -208,12 +209,14 @@ private void doRollover(ReadWriteChunkImpl<T> currentChunk) {
             @Override
             public void onSuccess(Boolean success) {
               if (success == null || !success) {
+                LOG.warn("roll over failed.");
```
bryanlb, line 212: We should use a consistent approach on these log messages to make them more readable - most currently are sentence cased without punctuation. I'm not particular about which format as long as we are consistent with one.

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -231,6 +234,7 @@ public void onFailure(Throwable t) {
    * When the ChunkManager is being closed, roll over active chunk so we can be sure that it's data is persisted in
    * a remote store.
    */
+  @VisibleForTesting
```
bryanlb, line 237: This annotation should be removed, as recovery nodes will need to manually rollover the active chunk as well.

kaldb/src/main/java/com/slack/kaldb/chunkManager/RollOverChunkTask.java
```diff
@@ -40,25 +40,30 @@ public RollOverChunkTask(
...
       rolloversCompletedCounter.increment();
       chunk.info().setChunkSnapshotTimeEpochMs(Instant.now().toEpochMilli());
       LOG.info("Finished chunk roll over {}", chunk.info());
       return true;
-    } else {
+    } catch (RuntimeException e) {
       rolloversFailedCounter.increment();
-      LOG.info("Failed chunk roll over {}", chunk.info());
-      return false;
+      LOG.warn("Failed chunk roll over {}", chunk.info(), e);
```
bryanlb, line 65: This should probably be an error, not a warn.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkCleanerServiceTest.java
```diff
@@ -43,16 +62,19 @@ public void setUp() throws Exception {
 
   @After
   public void tearDown() throws IOException, TimeoutException {
-    KaldbConfig.reset();
-    metricsRegistry.close();
+    LOG.info("tear down");
```
bryanlb, line 65: Not sure why we have this log message?

kaldb/src/test/java/com/slack/kaldb/chunk/IndexingChunkManagerTest.java
```diff
@@ -145,16 +161,16 @@ private void initChunkManager(
 
     // TODO: This is temporary, move this registration closer to logic.
     // Test search end point registration.
-    await().until(() -> chunkManager.getSearchMetadataStore().list().get().size() == 1);
-    SearchMetadata searchMetadata = chunkManager.getSearchMetadataStore().list().get().get(0);
-    assertThat(searchMetadata.name).isEqualTo("localhost");
-    assertThat(searchMetadata.snapshotName).isEqualTo(SearchMetadata.LIVE_SNAPSHOT_NAME);
-    assertThat(searchMetadata.url).contains(HOSTNAME);
-    assertThat(searchMetadata.url).contains(String.valueOf(port));
+    // await().until(() -> chunkManager.getSearchMetadataStore().list().get().size() == 1);
```
bryanlb, line 164: This should probably be removed now correct?

kaldb/src/test/java/com/slack/kaldb/chunk/ReadWriteChunkImplTest.java
```diff
@@ -271,13 +352,17 @@ public void testCleanupOnOpenChunk() throws IOException {
       assertThat(results.hits.size()).isEqualTo(1);
 
       chunk.close();
+      closeChunk = false;
+      // TODO: Missing next steps in test?
```
bryanlb, line 356: This needs a bit more context to explain what is expected to be resolved here.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -16,16 +16,46 @@
  * code into multiple, classes.
  *
  * <p>TODO: Have a read only chunk info for read only chunks so we don't accidentally update it.
- * TODO: Add a state machine for a chunk?
+ *
+ * <p>TODO: Add a minOffset field to chunk and snapshot metadata.
```
bryanlb, line 20: We should do this now at the same time we're adding the max offset, or remove the todo if it's no longer needed.

kaldb/src/main/java/com/slack/kaldb/chunkManager/RollOverChunkTask.java
```diff
@@ -40,25 +40,30 @@ public RollOverChunkTask(
     rolloversFailedCounter = meterRegistry.counter(ROLLOVERS_FAILED);
   }
 
+  // TODO: Is the return value needed?.
```
bryanlb, line 43: Not sure of the purpose of this todo? The only way we communicate a runtime exception with the refactored code is via this boolean, which is checked in the IndexingChunkManager `onSuccess`

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
bryanlb: Since we're already refactoring this class to add `maxOffset` we should add the `minOffset `at the same time instead of a todo. It's easier to add it now, and it should only be a few additional lines. Then we won't need to come back to it later or track the todo.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -16,16 +16,47 @@
...
  */
 public class ChunkInfo {
+  // Use a future time of Dec 31, 3000 instead of INSTANT.MAX since it overflows.
+  public static final long MAX_FUTURE_TIME = Instant.ofEpochSecond(32535158400L).toEpochMilli();
```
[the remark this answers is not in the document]
bryanlb: Ah yeah - since we're converting it to milliseconds that would cause an overflow - `Instant.MAX` uses the max long in seconds. In that case we should just use the `Long.MAX_VALUE`.

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
bryanlb: The recovery node work is already in progress, so this would have to be immediately removed. Instead of adding it to just remove it later this week let's not include it here. 

The current expectation is that this method can and will be called externally, so we shouldn't introduce an annotation that indicates otherwise.

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkCleanerServiceTest.java
```diff
@@ -117,36 +206,54 @@ public void testDeleteStateDataOn2Chunks() throws IOException {
...
 
+    await().until(() -> getCount(RollOverChunkTask.ROLLOVERS_COMPLETED, metricsRegistry) == 1);
     assertThat(chunkManager.getChunkList().size()).isEqualTo(2);
     assertThat(getCount(MESSAGES_RECEIVED_COUNTER, metricsRegistry)).isEqualTo(11);
     assertThat(getCount(MESSAGES_FAILED_COUNTER, metricsRegistry)).isEqualTo(0);
     assertThat(getCount(RollOverChunkTask.ROLLOVERS_INITIATED, metricsRegistry)).isEqualTo(1);
     assertThat(getCount(RollOverChunkTask.ROLLOVERS_FAILED, metricsRegistry)).isEqualTo(0);
     assertThat(getCount(RollOverChunkTask.ROLLOVERS_COMPLETED, metricsRegistry)).isEqualTo(1);
 
+    List<SnapshotMetadata> snapshots1 = fetchSnapshots(chunkManager);
```
bryanlb, line 236: Can we use a more descriptive names for these variables? It makes following the test a bit difficult as it's not clear what snapshots are expected to be in this list. (ie, `List<SnapshotMetadata> snapshotsBeforeFoo`)

This applies throughout this class where we have quite a few cases of variables with numbers appended (Ie, `snapshots4`).

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
mansu: Dropped the TODO for now since we may not even need this field.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -135,34 +187,27 @@ public boolean containsDataInTimeRange(long startTimeMs, long endTimeMs) {
    * Update the max and min data time range of the chunk given a new timestamp.
    */
   public void updateDataTimeRange(long messageTimeStampMs) {
-    if (dataStartTimeEpochMs == 0 || dataEndTimeEpochMs == 0) {
-      dataStartTimeEpochMs = messageTimeStampMs;
-      dataEndTimeEpochMs = messageTimeStampMs;
+    if (dataEndTimeEpochMs == MAX_FUTURE_TIME) {
+      dataStartTimeEpochMs = Math.min(dataStartTimeEpochMs, messageTimeStampMs);
+      dataEndTimeEpochMs = Math.min(dataEndTimeEpochMs, messageTimeStampMs);
```
bryanlb, line 192: I'm not sure this makes sense - since we already checked for equality of `dataEndTimeEpochMs` to that of `MAX_FUTURE_TIME` this is in effect:

```
dataEndTimeEpochMs = Math.min(MAX_FUTURE_TIME, messageTimeStampMs);
```
When would we expect `messageTimeStampMs` to be larger than that of `MAX_FUTURE_TIME`? Shouldn't this just be:
```
dataEndTimeEpochMs = messageTimeStampMs;
```

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -16,16 +15,45 @@
  * code into multiple, classes.
  *
  * <p>TODO: Have a read only chunk info for read only chunks so we don't accidentally update it.
- * TODO: Add a state machine for a chunk?
  */
 public class ChunkInfo {
+  // Use a future time of Dec 31, 3000 instead of INSTANT.MAX since it overflows.
```
bryanlb, line 20: This comment should be updated.

kaldb/src/test/java/com/slack/kaldb/chunk/IndexingChunkManagerTest.java
```diff
@@ -142,19 +157,10 @@ private void initChunkManager(
...
-    SearchMetadata searchMetadata = chunkManager.getSearchMetadataStore().list().get().get(0);
-    assertThat(searchMetadata.name).isEqualTo("localhost");
-    assertThat(searchMetadata.snapshotName).isEqualTo(SearchMetadata.LIVE_SNAPSHOT_NAME);
-    assertThat(searchMetadata.url).contains(HOSTNAME);
-    assertThat(searchMetadata.url).contains(String.valueOf(port));
   }
 
   @Test
-  @Disabled
+  @Ignore
```
bryanlb, line 163: Comment for future reference - `@Ignore` is Junit4, `@Disabled` is Junit5. We currently use Junit4.

bryanlb (dismissed):
Might want to update that comment, otherwise this looks good.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
mansu: Removed.

bryanlb (approved):

PR #141, opened 2021-10-28
Refactor cache node ZK path

bryanlb:
Removes unnecessary additional node depth, and moves path configs internal to appropriate stores.

Previous:
```
/cacheNode/{slotName}/SLOT
```

Now:
```
/cacheSlot/{slotName}
```

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -1,8 +1,6 @@
 package com.slack.kaldb.chunk;
 
-import static com.slack.kaldb.config.KaldbConfig.CACHE_SLOT_STORE_ZK_PATH;
 import static com.slack.kaldb.logstore.BlobFsUtils.copyFromS3;
-import static com.slack.kaldb.metadata.cache.CacheSlotMetadata.METADATA_SLOT_NAME;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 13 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -99,22 +99,21 @@ public ReadOnlyChunkImpl(
...
-    cacheSlotMetadataStore.addListener(cacheNodeListener());
-
     CacheSlotMetadata cacheSlotMetadata =
         new CacheSlotMetadata(
-            METADATA_SLOT_NAME, Metadata.CacheSlotState.FREE, "", Instant.now().toEpochMilli());
-    cacheSlotMetadataStore.create(cacheSlotMetadata);
+            slotName, Metadata.CacheSlotState.FREE, "", Instant.now().toEpochMilli());
+    cacheSlotMetadataStore.createSync(cacheSlotMetadata);
+
+    CacheSlotMetadataStore cacheSlotNodeMetadataStore =
```
mansu, line 114: I think updating the classdoc to explain the difference between `cacheSlotNodeMetadataStore` and `cacheSlotMetadataStore` would make it easier to read the code at a later time.

Nit: The names are too close. Consider using a different name for `cacheSlotNodeMetadataStore` like `slotListener`.

mansu (changes requested):
The code looks great! Thanks for the changes. One class doc is added this is good to go.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: Classdoc added for each constructor explaining the structure and intended use case for each, and renamed the `cacheSlotNodeMetadataStore` variable to better reflect it's for use as a listener only.

mansu (approved):

PR #142, opened 2021-11-03
Enforce that the node name in ZK is same as snapshotId.

mansu:
Currently, the node name can be different from snapshotId but this flexibility introduces bugs that are really hard to debug. Since we don't need this flexibility for now, enforce this restriction by modifying the constructor.


kaldb/src/main/java/com/slack/kaldb/metadata/snapshot/SnapshotMetadata.java
```diff
@@ -29,7 +52,7 @@ public SnapshotMetadata(
     checkState(startTimeUtc > 0, "start time should be greater than zero.");
     checkState(endTimeUtc > 0, "end time should be greater than zero.");
     checkState(endTimeUtc > startTimeUtc, "start time should be greater than  end time.");
-    checkState(maxOffset > 0, "max offset should be greater than zero.");
+    checkState(maxOffset >= 0, "max offset should be greater than zero.");
```
bryanlb, line 55: We need to update the error message as well

kaldb/src/main/java/com/slack/kaldb/util/SnapshotUtil.java
```diff
@@ -4,6 +4,6 @@
 
 public class SnapshotUtil {
```
bryanlb, line 5: This should be moved to the test package, as this is only for use in tests and not production code.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/util/SnapshotUtil.java
mansu: Good call. Moved.

kaldb/src/main/java/com/slack/kaldb/metadata/snapshot/SnapshotMetadata.java
mansu: Done.

bryanlb (approved):

PR #143, opened 2021-11-03
Add support for distributed query via zk discovery

vthacker:
Add support for distributed query via zk discovery

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -155,8 +155,10 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
 
   private void registerSearchMetadata(String snapshotName)
       throws ExecutionException, InterruptedException, TimeoutException {
+    // for "name" we pass url so that we can run two JVMs on the same host and not have znode path
+    // collision.
     this.searchMetadata =
-        new SearchMetadata(searchContext.hostname, snapshotName, searchContext.toUrl());
+        new SearchMetadata(searchContext.toName(), snapshotName, searchContext.toUrl());
     searchMetadataStore.create(searchMetadata).get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
   }
 
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -5,8 +5,6 @@ import brave.Tracing;
 import brave.grpc.GrpcTracing;
 import com.linecorp.armeria.client.Clients;
 import com.slack.kaldb.logstore.LogMessage;
-import com.slack.kaldb.metadata.core.KaldbMetadataStoreChangeListener;
-import com.slack.kaldb.metadata.search.SearchMetadata;
 import com.slack.kaldb.metadata.search.SearchMetadataStore;
 import com.slack.kaldb.proto.service.KaldbSearch;
 import com.slack.kaldb.proto.service.KaldbServiceGrpc;
@@ -17,6 +15,7 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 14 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -150,8 +150,10 @@ private KaldbMetadataStoreChangeListener cacheNodeListener() {
 
   private void registerSearchMetadata(String snapshotName)
       throws ExecutionException, InterruptedException, TimeoutException {
+    // for "name" we pass url so that we can run two JVMs on the same host and not have znode path
+    // collision.
     this.searchMetadata =
-        new SearchMetadata(searchContext.hostname, snapshotName, searchContext.toUrl());
+        new SearchMetadata(searchContext.toUrl(), snapshotName, searchContext.toUrl());
```
mansu, line 156: I think URL could contain protocol info going forward which can contain slashes that ZK doesn't like. So, it's better to use a different method to construct the node name from search method.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -44,39 +37,105 @@
...
+        // add new servers
+        latestSearchServers.forEach(
+            server -> {
+              if (!stubs.containsKey(server)) {
+                stubs.put(server, getKaldbServiceGrpcClient(protocol + server));
+              }
+            });
+
+        // invalidate old servers that no longer exist
+        stubs
```
mansu, line 83: I think it would be valuable to log the number of stubs added and removed as part of a run. Further, adding a counter for the changes and a guage to count the number of stubs would be very valuable to monitor this queue. 

In addition, a debug level message for the changes would help debugging immensely.

mansu (changes requested):
Leaving a couple of comments.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
vthacker: I don't think the protocol stored anywhere today? 

If we were to store protocol in `kaldb_configs.proto` we would need to define how a query node can talk to an indexing/cache/recovery node 

If that sounds something we want we can do that in a separate PR

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: Added everything except for the gauge! The reason for not adding the gauge was the number of  URLs we query is different from the number of search metadata objects. So the guage might give misleading info?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -25,16 +25,17 @@
...
+  private final Counter searchMetadataTotalChangeCounter;
 
-  private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String server) {
-    return Clients.newClient(server, KaldbServiceGrpc.KaldbServiceFutureStub.class)
-        .withCompression("gzip")
-        .withInterceptors(GrpcTracing.newBuilder(Tracing.current()).build().newClientInterceptor());
-  }
+  private Map<String, KaldbServiceGrpc.KaldbServiceFutureStub> stubs = new ConcurrentHashMap<>();
+
+  private static String protocol = "gproto+http://";
```
mansu, line 38: Please move this to toURL into the `SeachContext`.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -44,39 +45,119 @@
...
-            LOG.error("Node failed to respond ", ex);
-            return emptyResult;
-          });
-    } finally {
-      span.finish();
+      List<CompletableFuture<KaldbSearch.SearchResult>> queryServers =
+          new ArrayList<>(stubs.size());
+
+      for (KaldbServiceGrpc.KaldbServiceFutureStub stub : stubs.values()) {
+        // With the deadline we are keeping a high limit on how much time can each CompletableFuture
```
mansu, line 135: Reformat this comment?

mansu (changes requested):
This looks great! Some minor updates and this should be good to go.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -121,12 +122,13 @@ public void start() throws Exception {
     }
 
     if (roles.contains(KaldbConfigs.NodeRole.QUERY)) {
-      KaldbDistributedQueryService searcher = new KaldbDistributedQueryService();
-      KaldbDistributedQueryService.servers =
-          new ArrayList<>(KaldbConfig.get().getQueryConfig().getTmpIndexNodesList());
+      searchMetadataStore = new SearchMetadataStore(metadataStore, true);
```
vthacker, line 125: this is kinda ugly.. To all other "services" ( guava ) we can get away by passing `metadataStore` and then within the service we can create RecoveryNodeMetadataStore or SearchMetadataStore or some other store

vthacker: But here since `KaldbDistributedQueryService` is a GRPC service so SearchMetadataStore needs to be created in `Kaldb` so that we can use the shutdown hook to close it

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -121,12 +126,14 @@ public void start() throws Exception {
     }
 
     if (roles.contains(KaldbConfigs.NodeRole.QUERY)) {
-      KaldbDistributedQueryService searcher = new KaldbDistributedQueryService();
-      KaldbDistributedQueryService.servers =
-          new ArrayList<>(KaldbConfig.get().getQueryConfig().getTmpIndexNodesList());
+      SearchMetadataStore searchMetadataStore = new SearchMetadataStore(metadataStore, true);
+      MetadataStoreLifecycleManager.addMetadataStore((MetadataStore) searchMetadataStore);
```
vthacker, line 130: TODO: verify if typecast is the right thing here?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: Could you explain how a gauge will help ? There can be many search requests being made and a gauge won't help tell us on a per search basis the cardinality ? 

I think tracing would be a better fit no?

kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
```diff
@@ -258,10 +312,100 @@ public void testSearch() {
...
     assertThat(searchResponse.getTotalCount()).isEqualTo(200);
     assertThat(searchResponse.getHitsCount()).isEqualTo(100);
+
+    verifyAddServerToZK();
+    testSearchWithOneShardTimeout();
   }
 
-  @Test
-  public void testSearchWithOneShardTimeout() {
+  // TODO: Verify if we respect setIndexName(MessageUtil.TEST_INDEX_NAME)
```
mansu, line 320: Can we remove these TODOs?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
mansu: I think a guage would help us detect if too many stubs are added or removed. Tracing is not a great fit here since this logic is not in the request path.

I think for now let's just add how many stubs are added and removed on the log message on line 114.

mansu (commented):
Couple of minor nits and this is good to go.

kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
vthacker: yup! removing it now

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -62,65 +61,58 @@ public KaldbDistributedQueryService(
...
         this.meterRegistry.counter(SEARCH_METADATA_TOTAL_CHANGE_COUNTER);
-    this.searchMetadataStore.addListener(searchServersListener());
-    initStubs();
+    this.searchMetadataStore.addListener(this::updateStubs);
+
+    // first time call this function manually so that we initialize stubs
+    updateStubs();
   }
 
-  private void initStubs() {
```
vthacker: unified it to one method called `updateStubs`

kaldb/src/test/java/com/slack/kaldb/chunk/ChunkCleanerServiceTest.java
```diff
@@ -53,6 +54,8 @@ public void setUp() throws Exception {
     metricsRegistry = new SimpleMeterRegistry();
     chunkManagerUtil =
         new ChunkManagerUtil<>(S3_MOCK_RULE, metricsRegistry, 10 * 1024 * 1024 * 1024L, 10L);
+    chunkManagerUtil.chunkManager.startAsync();
```
vthacker, line 57: We changed the semantics of `ChunkManagerUtil` ( a test helper ) to not start chunk manager

Because of this we need to start it manually everywhere

The reason for the change was in `KaldbDistributedQueryServiceTest` we need to put chunk manager in a service manager ( which will not allow a running service to be put in the SM )

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
vthacker: fixed

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: added

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -44,39 +45,119 @@
...
+    try {
+      searchMetadataTotalChangeCounter.increment();
+      searchMetadataNodes = searchMetadataStore.list().get();
+      searchMetadataNodes.forEach(
+          searchMetadata ->
+              stubs.put(
+                  searchMetadata.url, getKaldbServiceGrpcClient(protocol + searchMetadata.url)));
+    } catch (Exception e) {
+      LOG.error("Cannot initStubs stubs");
+      // throw runtime exception?
```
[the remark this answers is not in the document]
vthacker: done

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: done

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: done

mansu (approved):

PR #146, opened 2021-11-09
Add recovery task assignment service

bryanlb:
Adds the recovery task assigner service


bryanlb:
This will be slightly changed after #150 is merged, closing for now.

PR #149, opened 2021-11-12
Make ReadWriteChunkImpl an abstract base class.

mansu:
Refactored `ReadWriteChunkImpl` into 2 classes: an abstract base class `ReadWriteChunk` and `IndexingReadWriteChunkImpl` containing the logic related to index registration. 

This separation makes room for implementing the recovery indexer.


kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
```diff
@@ -188,68 +179,49 @@ public void commit() {
...
   public void preSnapshot() {
-    LOG.info("Started RW chunk pre-snapshot {}", chunkInfo);
+    logger.info("Started RW chunk pre-snapshot {}", chunkInfo);
     setReadOnly(true);
     commit();
-    LOG.info("Finished RW chunk pre-snapshot {}", chunkInfo);
+    logger.info("Finished RW chunk pre-snapshot {}", chunkInfo);
   }
 
+  public abstract void postSnapshot();
```
bryanlb, line 190: Each of the abstract methods should have a javadoc explaining their use as well in addition to the overall class doc. IntelliJ surfaces this info (if available) when you go to extend the use, which makes it easier to reference.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
```diff
@@ -25,65 +25,67 @@
...
  *
- * <p>A ReadWriteChunk goes through the following life cycle.
+ * <p>A ReadWriteChunk goes through the following life cycle. The hooks into the life cycle are
+ * implemented as abstract base methods so derived classes can take custom action on those stages.
  *
  * <p>When a chunk is created it is open for both reads and writes. Since a ReadWriteChunk is
- * ingesting live data, a cluster manager doesn't manage it. Instead, when a chunk in created, it
- * registers a live snapshot and a live search node.
+ * ingesting live data, a cluster manager doesn't manage it. The register method is called after the
+ * chunk is created and deRegister method is called just before chunk close.
```
bryanlb, line 44: These methods should be named for when they occur, and not what the current implementation does today. For example `afterChunkCreation` and `beforeChunkClose` is better than `register` and `deRegister`, as registration is a concept specific to `IndexingReadWriteChunkImpl`. Future `ReadWriteChunks` will want to take actions at these lifecycle stages that have nothing to do with registration.

This also makes these two methods consistent with `postSnapshot` which already follows this pattern.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/IndexingReadWriteChunkImpl.java
```diff
@@ -0,0 +1,93 @@
...
+            chunkInfo.getDataEndTimeEpochMs(),
+            chunkInfo.getMaxOffset(),
+            chunkInfo.getKafkaPartitionId());
+    snapshotMetadataStore.updateSync(updatedSnapshotMetadata);
+    liveSnapshotMetadata = updatedSnapshotMetadata;
+
+    LOG.info("Post snapshot operation completed for RW chunk {}", chunkInfo);
+  }
+
+  public void register() {
```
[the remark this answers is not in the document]
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
mansu: Good call. Reamed to `postCreate` and `preClose` to be consistent with the snapshot names.

bryanlb (approved):

PR #151, opened 2021-11-18
add metadata store lifecycle manager

vthacker:
Add metadata store lifecycle manager. The code has example an example usage for the query service. 

in this design every service creates it's own MetadataStoreLifecycleManager and adds the list of stores that it needs. Then the current framework will do the job of closing the stores on shutdown

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/MetadataStoreLifecycleManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/MetadataStoreLifecycleManager.java
@@ -1,0 +1,36 @@
+package com.slack.kaldb.metadata.zookeeper;
+
+import com.google.common.util.concurrent.AbstractIdleService;
+import com.slack.kaldb.metadata.core.CacheableMetadataStore;
+import com.slack.kaldb.proto.config.KaldbConfigs;
+import java.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Each instance of MetadataStoreLifecycleManager holds all the metadata stores required for each
+ * node role MetadataStoreLifecycleManager is a managed guava service. The caller needs to add the
+ * created instance to a service manager which would then take care of the object lifecycle
+ */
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/MetadataStoreLifecycleManager.java
```diff
@@ -0,0 +1,22 @@
+package com.slack.kaldb.metadata.zookeeper;
+
+import com.google.common.util.concurrent.AbstractIdleService;
+import com.slack.kaldb.metadata.core.CacheableMetadataStore;
+import java.util.*;
+
+public class MetadataStoreLifecycleManager extends AbstractIdleService {
```
bryanlb, line 7: We should add a classdoc that explains the intended use of this class.

kaldb/src/main/java/com/slack/kaldb/metadata/zookeeper/MetadataStoreLifecycleManager.java
```diff
@@ -0,0 +1,22 @@
...
+
+  public MetadataStoreLifecycleManager(List<CacheableMetadataStore<?>> metadataStores) {
+    this.metadataStores = metadataStores;
+  }
+
+  @Override
+  protected void startUp() throws Exception {}
+
+  @Override
+  protected void shutDown() throws Exception {
```
bryanlb, line 19: Probably should add a log message when the shutdown is complete.

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -52,6 +52,7 @@
       new PrometheusMeterRegistry(PrometheusConfig.DEFAULT);
   protected ServiceManager serviceManager;
   protected MetadataStore metadataStore;
+  protected MetadataStoreLifecycleManager metadataStoreLifecycleManager;
```
bryanlb, line 55: This isn't needed anymore.

bryanlb (changes requested):

bryanlb (approved):

PR #154, opened 2021-11-29
Cache slot assignment service

bryanlb:
Assigns available cache slots to replias awaiting assignment


bryanlb:
Closed in favor of #159

PR #160, opened 2021-12-06
add thread name to all spans

vthacker:
This info is useful to debug certain scenarios where we can tell on which thread pool is my async code running.

For example right now when I deployed it I see `threadName=armeria-eventloop-epoll-4-7` as the span tag for `KaldbDistributedQueryService.distributedSearch` span and any other query service span

I was expecting this to be the blocking thread pool executor... I think ...

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
@@ -118,6 +118,7 @@ public class ArmeriaService extends AbstractIdleService {
                     .addScopeDecorator(ThreadContextScopeDecorator.get())
                     .build())
             .addSpanHandler(spanHandler)
+            .addSpanHandler(new KaldbSpanHandler())
             .build();
     sb.decorator(BraveService.newDecorator(tracing));
   }
--- a/kaldb/src/main/java/com/slack/kaldb/server/KaldbSpanHandler.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/KaldbSpanHandler.java
@@ -1,0 +1,15 @@
+package com.slack.kaldb.server;
+
+import brave.handler.MutableSpan;
+import brave.handler.SpanHandler;
+import brave.propagation.TraceContext;
+
+/** Add common tags that we want to annotate on every request */
+public class KaldbSpanHandler extends SpanHandler {
+
+  @Override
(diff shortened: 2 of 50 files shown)


kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -118,6 +120,15 @@ private void addTracing(ServerBuilder sb) {
                     .addScopeDecorator(ThreadContextScopeDecorator.get())
                     .build())
             .addSpanHandler(spanHandler)
+            .addSpanHandler(
+                new SpanHandler() {
+                  @Override
+                  public boolean begin(
+                      TraceContext context, MutableSpan span, TraceContext parent) {
+                    span.tag("threadName", Thread.currentThread().getName());
+                    return super.begin(context, span, parent);
```
bryanlb, line 129: I believe this should return a boolean instead of invoking the super, per the documentation?

What do we think about moving this to a separate class? Will we want to expand the available common tags over time?

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
vthacker: Hmm! here we do `return super.begin(...)` which should return a bool?

vthacker: moved to a separate class

kaldb/src/main/java/com/slack/kaldb/server/KaldbSpanHandler.java
```diff
@@ -0,0 +1,15 @@
...
+import brave.handler.MutableSpan;
+import brave.handler.SpanHandler;
+import brave.propagation.TraceContext;
+
+/** Add common tags that we want to annotate on every request */
+public class KaldbSpanHandler extends SpanHandler {
+
+  @Override
+  public boolean begin(TraceContext context, MutableSpan span, TraceContext parent) {
+    span.tag("threadName", Thread.currentThread().getName());
```
mansu, line 12: `thread` would also work.

mansu (approved):

vthacker:
JSON Logging adds this by default

PR #164, opened 2021-12-09
Distributed Query Service: Parallelize more work

vthacker:
Parallelize the call to `SearchResultUtils::fromSearchResultProtoOrEmpty` by moving it right after the individual gRPC call finish instead of waiting for all gRPC calls to finish

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResult.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResult.java
@@ -2,12 +2,15 @@ package com.slack.kaldb.logstore.search;
 
 import com.google.common.base.Objects;
 import com.slack.kaldb.histogram.HistogramBucket;
+import com.slack.kaldb.logstore.LogMessage;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResult.java
```diff
@@ -2,12 +2,15 @@
...
 import com.slack.kaldb.histogram.HistogramBucket;
+import com.slack.kaldb.logstore.LogMessage;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class SearchResult<T> {
 
-  private static final SearchResult EMPTY = new SearchResult<>(null, 0, 0, null, 1, 1, 0, 0);
+  private static final SearchResult EMPTY =
```
vthacker, line 12: `Collections.emptyList()` is needed instead of null to avoid NPE

vthacker, line 12: Collections.emptyList()

mansu (approved):

PR #165, opened 2021-12-09
Refactor chunk, metadata IDs

bryanlb:
Fixes consistency issues, as well as collisions issues on chunk IDs.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -155,10 +155,9 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
 
   private void registerSearchMetadata(String snapshotName)
       throws ExecutionException, InterruptedException, TimeoutException {
-    // for "name" we pass url so that we can run two JVMs on the same host and not have znode path
-    // collision.
     this.searchMetadata =
-        new SearchMetadata(searchContext.toName(), snapshotName, searchContext.toUrl());
+        new SearchMetadata(
+            snapshotName + "_" + searchContext.hostname, snapshotName, searchContext.toUrl());
     searchMetadataStore.create(searchMetadata).get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
   }
 
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -157,7 +157,9 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
       throws ExecutionException, InterruptedException, TimeoutException {
     this.searchMetadata =
         new SearchMetadata(
-            snapshotName + "_" + searchContext.hostname, snapshotName, searchContext.toUrl());
+            SearchMetadata.getSnapshotName(snapshotName, searchContext.hostname),
+            snapshotName,
+            searchContext.toUrl());
     searchMetadataStore.create(searchMetadata).get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
   }
 
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
@@ -98,7 +98,7 @@ public abstract class ReadWriteChunk<T> implements Chunk<T> {
     this.kafkaPartitionId = kafkaPartitionId;
     chunkInfo =
         new ChunkInfo(
-            chunkDataPrefix + "_" + chunkCreationTime.toEpochMilli() + "_" + UUID.randomUUID(),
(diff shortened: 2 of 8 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/snapshot/SnapshotMetadataStore.java
```diff
@@ -6,7 +6,7 @@
 import org.slf4j.LoggerFactory;
 
 public class SnapshotMetadataStore extends PersistentMutableMetadataStore<SnapshotMetadata> {
-  public static final String SNAPSHOT_METADATA_STORE_ZK_PATH = "/snapshots";
+  public static final String SNAPSHOT_METADATA_STORE_ZK_PATH = "/snapshot";
```
bryanlb, line 9: This is the only ZK path that is plural, the rest are singular (search, replica, cacheSlot, etc)

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -155,10 +155,8 @@ private KaldbMetadataStoreChangeListener cacheNodeListener() {
 
   private void registerSearchMetadata(String snapshotName)
       throws ExecutionException, InterruptedException, TimeoutException {
-    // for "name" we pass url so that we can run two JVMs on the same host and not have znode path
-    // collision.
     this.searchMetadata =
-        new SearchMetadata(searchContext.toName(), snapshotName, searchContext.toUrl());
+        new SearchMetadata(snapshotName + "_" + searchContext.hostname, snapshotName, searchContext.toUrl());
```
bryanlb, line 159: Prevents collisions when a single node needs to register multiple search metadata (ie, cache nodes)

kaldb/src/main/java/com/slack/kaldb/chunk/ReadWriteChunk.java
```diff
@@ -121,7 +123,7 @@ protected ReadWriteChunk(
   public abstract void preClose();
 
   private SearchMetadata toSearchMetadata(String snapshotName, SearchContext searchContext) {
-    return new SearchMetadata(snapshotName, snapshotName, searchContext.toUrl());
+    return new SearchMetadata(snapshotName + "_" + searchContext.hostname, snapshotName, searchContext.toUrl());
```
bryanlb, line 126: Adds hostname for easier debugging and consistency with the ReadOnlyChunk

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -155,10 +155,9 @@ private KaldbMetadataStoreChangeListener cacheNodeListener() {
 
   private void registerSearchMetadata(String snapshotName)
       throws ExecutionException, InterruptedException, TimeoutException {
-    // for "name" we pass url so that we can run two JVMs on the same host and not have znode path
-    // collision.
     this.searchMetadata =
-        new SearchMetadata(searchContext.toName(), snapshotName, searchContext.toUrl());
+        new SearchMetadata(
+            snapshotName + "_" + searchContext.hostname, snapshotName, searchContext.toUrl());
```
mansu, line 160: I think it's best to update the toName() method in SearchContext or add a new method in SearchMetadata to keep all the name building logic in the same place. This logic is already repeated twice here. 

May be make it static so we can use it in unit tests also.

kaldb/src/main/java/com/slack/kaldb/metadata/snapshot/SnapshotMetadataStore.java
mansu: +1 for consistency.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
```diff
@@ -153,7 +153,7 @@ public void shouldHandleChunkLivecycle() throws Exception {
     assertThat(searchMetadataStore.getCached().get(0).snapshotName).isEqualTo(snapshotId);
     assertThat(searchMetadataStore.getCached().get(0).url)
         .isEqualTo("gproto+http://localhost:8080");
-    assertThat(searchMetadataStore.getCached().get(0).name).isEqualTo("localhost:8080");
+    assertThat(searchMetadataStore.getCached().get(0).name).isEqualTo(snapshotId + "_localhost");
```
mansu, line 156: I think it's best to call the toName method here to check the name.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: Done.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
bryanlb: Done.

mansu (approved):

PR #169, opened 2021-12-10
Create codeql-analysis.yml

mansu:
Enable codeQL analysis on code. Fix some security issues found by the tool.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '39 10 * * 2'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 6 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
```
vthacker, line 20: so we will run it on every commit and as a cron?

mansu: Looks like it runs on every commit. Not sure about the cron bit.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultAggregatorImpl.java
```diff
@@ -39,7 +39,7 @@ public SearchResultAggregatorImpl(SearchQuery searchQuery) {
     int totalNodes = 0;
     int totalSnapshots = 0;
     int snapshpotReplicas = 0;
-    int totalCount = 0;
+    long totalCount = 0;
```
bryanlb, line 42: What's the reason behind changing these? I'm not sure why we're moving some of the counters to doubles or long values, when these will never contain decimal values, or exceed the int overflow. It seems like perhaps the reason these are being flagged is due to issues elsewhere?

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultAggregatorImpl.java
mansu: The tool flagged them as high risk, so changed them. These are legitimate changes though since now there are no down conversions and all conversions are explicit.

mansu:
Revisit at a later time.

PR #175, opened 2021-12-14
Use JSON Logging

vthacker:
Makes looking at logs in kibana much easier.

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -64,6 +64,11 @@
             <artifactId>log4j-core</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-layout-template-json</artifactId>
+            <version>${log4j.version}</version>
+        </dependency>
 
         <dependency>
             <groupId>com.lmax</groupId>
@@ -487,7 +492,14 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.4.1</version>
+                <version>3.2.4</version>
+                <dependencies>
+                    <dependency>
+                        <groupId>com.github.edwgiz</groupId>
+                        <artifactId>maven-shade-plugin.log4j2-cachefile-transformer</artifactId>
+                        <version>2.15.0</version>
+                    </dependency>
+                </dependencies>
                 <executions>
                     <execution>
                         <phase>package</phase>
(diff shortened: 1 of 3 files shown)


kaldb/pom.xml
```diff
@@ -487,7 +492,14 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.4.1</version>
+                <version>3.2.4</version>
+                <dependencies>
+                    <dependency>
```
mansu, line 497: Do we need these external dependencies? Can we make this change without these additional dependencies?

mansu: How hard would it be to emit log messages in span json format?

kaldb/src/main/resources/log4j2.xml
```diff
@@ -4,7 +4,9 @@
 
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
-            <PatternLayout  pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %notEmpty{[trace=%X{traceId}/%X{spanId}]} %c{1} - %msg%n" />
+            <!-- https://github.com/apache/logging-log4j2/blob/master/log4j-layout-template-json/src/main/resources/EcsLayout.json -->
+            <!-- MDC keys will get their own field as per EcsLayout.json  -->
+            <JsonTemplateLayout eventTemplateUri="classpath:EcsLayout.json" />
```
mansu, line 9: How hard would it be to emit log messages in span json format and use that as a template?

kaldb/pom.xml
vthacker: nope :) It won't compile. I've linked the github issue

kaldb/src/main/resources/log4j2.xml
vthacker: What's a span JSON format? Also we have to then see if that format can be ingested by our internal logging pipeline.

kaldb/src/main/resources/log4j2.xml
```diff
@@ -4,7 +4,8 @@
 
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
-            <PatternLayout  pattern="[%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %notEmpty{[trace=%X{traceId}/%X{spanId}]} %c{1} - %msg%n" />
+            <!-- https://github.com/apache/logging-log4j2/blob/master/log4j-layout-template-json/src/main/resources/EcsLayout.json -->
```
bryanlb, line 7: Should this comment be updated/removed?

bryanlb (approved):

kaldb/src/main/resources/log4j2.xml
vthacker: I kinda removed it then kept it.. It helped me checking out examples from the project.

PR #176, opened 2021-12-16
Cleanup stale indexer nodes and recovery task creation. 

mansu:
Currently, when an indexer restarts we leave stale metadata in ZK. This class cleans up stale indexer metadata. In addition, there is data loss for the data that is indexed but not uploaded to S3. 

In this diff, we will implement stale live snapshot cleanup. In addition, we will determine the last offset in the durable metadata store and optionally create a recovery task if we can't catch up.


kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is responsible for the indexer startup operations like stale live snapshot cleanup.
+ * determining the start indexing offset and create Recovery task etc..
+ */
+public class RecoveryTaskFactory {
```
bryanlb, line 17: Instead of `RecoveryTaskFactory` this might be better as something like a `StartupOperationsUtils`. This class doesn't implement a factory pattern, and we already have expanded this to include non-recovery tasks, such as "stale live snapshot" cleanup.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,81 @@
...
+    this.partitionId = partitionId;
+  }
+
+  public List<SnapshotMetadata> getStaleLiveSnapshots(List<SnapshotMetadata> snapshots) {
+    List<SnapshotMetadata> snapshotsForPartition =
+        snapshots
+            .stream()
+            .filter(snapshotMetadata -> snapshotMetadata.partitionId.equals(partitionId))
+            .collect(Collectors.toUnmodifiableList());
+    return snapshotsForPartition
```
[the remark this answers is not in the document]
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/util/SnapshotsUtil.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A metadata class that contains useful utilities for working with Snapshots. */
+public class SnapshotsUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(SnapshotsUtil.class);
+  private static final int SNAPSHOT_OPERATION_TIMEOUT_SECS = 10;
+
+  public static int deleteSnapshots(
```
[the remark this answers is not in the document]
mansu: Agree. Let's refactor this code along with other cluster manger operations.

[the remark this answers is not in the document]
mansu: Agree. Made this a private method for now. We can refactor this code out when refactor this logic for cluster manager services.

kaldb/src/main/java/com/slack/kaldb/util/SnapshotsUtil.java
```diff
@@ -0,0 +1,60 @@
...
+                })
+            .collect(Collectors.toUnmodifiableList());
+
+    ListenableFuture<?> futureList = Futures.successfulAsList(deletionFutures);
+    try {
+      futureList.get(SNAPSHOT_OPERATION_TIMEOUT_SECS, TimeUnit.SECONDS);
+    } catch (Exception e) {
+      futureList.cancel(true);
+    }
+
```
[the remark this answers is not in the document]
mansu: Added.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
mansu: I renamed this class `StartupOperationsUtils` but that was even more vague than `RecoveryTaskFactory`. So, left the name as is. Let's change the name when we integrate this code with the indexer.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+  public List<SnapshotMetadata> getStaleLiveSnapshots(List<SnapshotMetadata> snapshots) {
+    return snapshots
+        .stream()
+        .filter(snapshotMetadata -> snapshotMetadata.partitionId.equals(partitionId))
+        .filter(SnapshotMetadata::isLive)
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  @VisibleForTesting
+  public List<SnapshotMetadata> deleteStaleLiveSnapshots(List<SnapshotMetadata> snapshots) {
```
vthacker, line 72: Why not just reuse `deleteSnapahots` and the caller can call `getStaleLiveSnapshots` to get the list of stale snapshots if that's what it wants to do?

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+      LOG.warn("Failed to delete {} live snapshots", failedDeletes);
+      throw new IllegalStateException("Failed to delete stale live snapshots");
+    }
+
+    return staleSnapshots;
+  }
+
+  // Get the highest offset for which data is durable for a partition.
+  @VisibleForTesting
+  public long getHighestDurableOffsetForPartition(
```
vthacker, line 88: The PR description says this PR only cleans up old indexer stale state. Is this method needed for that? Or for the followup PR ?

mansu: This method is used for creating the recovery task.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+          currentHeadOffsetForPartition,
+          maxOffsetDelay);
+      recoveryTaskMetadataStore.createSync(
+          new RecoveryTaskMetadata(
+              recoveryTaskName,
+              partitionId,
+              nextOffsetForPartition,
+              currentHeadOffsetForPartition - 1,
+              creationTimeEpochMs));
+      LOG.info(
```
bryanlb, line 200: We probably need some additional metrics across this method - I think knowing the amount of recovery tasks created and the offsets/size (as tags?) would be helpful. That would allow us to see how often we're creating recovery tasks, and how big they are, etc.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    // The head offset for Kafka partition is the offset of the next message to be indexed. We
+    // assume that offset is passed into this function. The highest durable offset is the partition
+    // offset of the message that is indexed. Hence, the offset is incremented by 1 to get the
+    // next message.
+    long nextOffsetForPartition = highestDurableOffsetForPartition + 1;
+
+    // Create a recovery task if needed.
+    if (currentHeadOffsetForPartition - highestDurableOffsetForPartition > maxOffsetDelay) {
```
bryanlb, line 184: We should have a minimum threshold before a recovery task will be created - if we're only lagging say a few hundred messages an indexer can easily get caught up to that. If it's a few hundred thousand we'll need a recovery task.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+      final long creationTimeEpochMs = Instant.now().toEpochMilli();
+      final String recoveryTaskName = "recoveryTask_" + partitionId + "_" + creationTimeEpochMs;
```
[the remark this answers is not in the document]
mansu: Changed.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
```diff
@@ -0,0 +1,260 @@
...
+    this.recoveryTaskMetadataStore = recoveryTaskMetadataStore;
+    this.partitionId = partitionId;
+    this.maxOffsetDelay = maxOffsetDelay;
+
+    snapshotDeleteSuccess = meterRegistry.counter(STALE_SNAPSHOT_DELETE_SUCCESS);
+    snapshotDeleteFailed = meterRegistry.counter(STALE_SNAPSHOT_DELETE_FAILED);
+  }
+
+  @VisibleForTesting
+  public List<SnapshotMetadata> getStaleLiveSnapshots(List<SnapshotMetadata> snapshots) {
```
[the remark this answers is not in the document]
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskFactory.java
mansu: Done.

bryanlb (approved):
🚀

PR #177, opened 2021-12-16
chunk cleaner service should never delete active chunk

vthacker:


kaldb/src/test/java/com/slack/kaldb/chunk/ChunkCleanerServiceTest.java
```diff
@@ -96,6 +96,9 @@ public void testDeleteStaleDataOn1Chunk()
 
     testBasicSnapshotMetadata(chunkManager, creationTime);
 
+    // try to delete active chunk
+    assertThat(chunkCleanerService.deleteStaleData(Instant.now().plusSeconds(100))).isEqualTo(0);
```
vthacker, line 100: test will fail without the fix

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkCleanerService.java
```diff
@@ -48,7 +48,8 @@ public int deleteStaleChunksPastCutOff(long staleDataCutOffMs) {
 
     List<Chunk<T>> staleChunks = new ArrayList<>();
     for (Chunk<T> chunk : (Iterable<Chunk<T>>) chunkManager.chunkList) {
-      if (chunk.info().getChunkSnapshotTimeEpochMs() <= staleDataCutOffMs) {
+      if (chunk.isSnapshotted()
```
bryanlb, line 51: It seems like we should be able to simplify this, and keep all of this logic contained to the `ChunkCleanerService` instead of adding it up to the `Chunk` interface. The concept of a snapshot is specific to the `ReadWriteChunk`, so it's a bit awkward to add this to the interface.

We already have to check the the `snapshotTime` against the `staleDataCutoff` - why not just do something like the following?

```
private boolean chunkIsStale(ChunkInfo chunkInfo, long staleDataCutoffMs) {
  return chunkInfo.getChunkSnapshotTimeEpochMs() > 0 && chunkInfo.getChunkSnapshotTimeEpochMs() <= staleDataCutoffMs;
}
```

We shouldn't need to check `readOnly` - if something has been snapshotted it should have a snapshot time populated, else it should be zero.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkCleanerService.java
vthacker: This would work as well. I like this better. Lemme change the PR

bryanlb (approved):

PR #179, opened 2021-12-16
Add replica eviction service

bryanlb:
Marks expired replicas that are currently assigned to a cache slot for eviction.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
@@ -1,0 +1,164 @@
+package com.slack.kaldb.clusterManager;
+
+import static com.google.common.util.concurrent.Futures.addCallback;
+import static com.slack.kaldb.config.KaldbConfig.DEFAULT_ZK_TIMEOUT_SECS;
+import static com.slack.kaldb.util.FutureUtils.successCountingCallback;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.AbstractScheduledService;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.slack.kaldb.metadata.cache.CacheSlotMetadata;
+import com.slack.kaldb.metadata.cache.CacheSlotMetadataStore;
+import com.slack.kaldb.metadata.replica.ReplicaMetadata;
+import com.slack.kaldb.metadata.replica.ReplicaMetadataStore;
+import com.slack.kaldb.proto.config.KaldbConfigs;
+import com.slack.kaldb.proto.metadata.Metadata;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Marks expired replicas that are currently assigned to a cache slot for eviction. This service
+ * does not delete or mutate replicas, but only marks the cache slot as pending eviction. Once the
+ * cache node has evicted an expired replica, a separate service will delete expired, unassigned
+ * replicas.
+ */
+public class ReplicaEvictionService extends AbstractScheduledService {
+  private static final Logger LOG = LoggerFactory.getLogger(ReplicaEvictionService.class);
+
+  private final CacheSlotMetadataStore cacheSlotMetadataStore;
+  private final ReplicaMetadataStore replicaMetadataStore;
+  private final KaldbConfigs.ManagerConfig managerConfig;
+  private final MeterRegistry meterRegistry;
+
+  @VisibleForTesting protected int futuresListTimeoutSecs = DEFAULT_ZK_TIMEOUT_SECS;
+
+  public static final String REPLICA_MARK_EVICT_SUCCEEDED = "replica_mark_evict_succeeded";
+  public static final String REPLICA_MARK_EVICT_FAILED = "replica_mark_evict_failed";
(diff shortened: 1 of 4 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Marks expired replicas that are currently assigned to a cache slot for eviction. This service
+ * does not delete or mutate replicas, but only marks the cache slot as pending eviction. Once the
+ * cache node has evicted an expired replica, a separate service will delete expired, unassigned
```
mansu, line 34: Why do we need a separate service for replica deletion? Since, the logic to evict replicas is here, is this the best place to know what replicas to delete?

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
```diff
@@ -0,0 +1,164 @@
...
+  protected int markReplicasForEviction() {
+    Timer.Sample evictionTimer = Timer.start(meterRegistry);
+
+    Map<String, ReplicaMetadata> replicaMetadataByReplicaId =
+        replicaMetadataStore
+            .getCached()
+            .stream()
+            .collect(Collectors.toUnmodifiableMap(ReplicaMetadata::getName, Function.identity()));
+
+    long expireAfterNow = Instant.now().toEpochMilli();
```
mansu, line 109: If the `expireAfterNow` time is passed in as an argument from the caller function, it makes this function deterministic and easier to unit test.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
```diff
@@ -0,0 +1,164 @@
...
+    AtomicInteger successCounter = new AtomicInteger(0);
+    List<ListenableFuture<?>> replicaEvictions =
+        cacheSlotMetadataStore
+            .getCached()
+            .stream()
+            // get all the slots that are currently assigned, but have an expiration in the past
+            .filter(
+                cacheSlotMetadata ->
+                    cacheSlotMetadata.cacheSlotState.equals(
+                            Metadata.CacheSlotMetadata.CacheSlotState.ASSIGNED)
```
mansu, line 119: We should be evicting all snapshots which are in LIVE, not ASSIGNED. Since the node stays assigned for a short amount of time, it's probably ok to let it go to loading. Should we also evict snapshots in LOADING state?

Also, extract the logic for the filter into a predicate so the we can test this logic easily.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
```diff
@@ -0,0 +1,164 @@
...
+
+                  addCallback(
+                      future,
+                      successCountingCallback(successCounter),
+                      MoreExecutors.directExecutor());
+                  return future;
+                })
+            .collect(Collectors.toUnmodifiableList());
+
+    ListenableFuture<?> futureList = Futures.successfulAsList(replicaEvictions);
```
mansu, line 142: Since this code is repetitive across services, let's refactor it into a utility function.

kaldb/src/test/java/com/slack/kaldb/clusterManager/ReplicaEvictionServiceTest.java
```diff
@@ -0,0 +1,814 @@
...
+              UUID.randomUUID().toString(),
+              Instant.now().toEpochMilli(),
+              Instant.now().plusSeconds(60).toEpochMilli());
+      replicas.add(replicaMetadata);
+      replicaMetadataStore.create(replicaMetadata);
+    }
+
+    List<CacheSlotMetadata> cacheSlots = new ArrayList<>();
+    for (int i = 0; i < 3; i++) {
+      CacheSlotMetadata cacheSlotMetadata =
```
mansu, line 162: It may be better to add a unit test with mixed cache slot states and make sure those replicas are unaffected. Alternatively, it may be simpler to refactor the filter function into a predicate and add a unit test for that.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
bryanlb: As discussed, we're keeping these as separate services for two reasons - one is that eviction can take a while to complete, and the second is we want replica eviction and replica deletion to be independently fault tolerant. We could combine these into a single service / class, but as they're doing independent operations it makes sense to keep them separate for now.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
bryanlb: Ah, good call out - these should all be checking for LIVE, not ASSIGNED. Also extracted the logic for the filter.

I think leaving the ones marked LOADING is fine - the only time we would expect a node to be LOADING an expired snapshot is if it had somehow gotten "stuck" for a prolonged period of time - which the monitoring service should address.

kaldb/src/test/java/com/slack/kaldb/clusterManager/ReplicaEvictionServiceTest.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaEvictionService.java
bryanlb: As discussed, I agree that we should come up with a common util. Let's address this as a separate PR once the pending ones are merged as a single pass. See https://github.com/slackhq/kaldb/pull/176#discussion_r779696717 for discussion about where I'm thinking we should go with this.

mansu (approved):

PR #180, opened 2021-12-17
query span improvements

vthacker:


kaldb/src/main/java/com/slack/kaldb/util/SpanUtils.java
```diff
@@ -0,0 +1,18 @@
+package com.slack.kaldb.util;
+
+import brave.ScopedSpan;
+import com.slack.kaldb.logstore.search.SearchResult;
+
+public class SpanUtils {
```
bryanlb, line 6: Probably should have a comment for the class explaining the purpose

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbLocalQueryService.java
```diff
@@ -20,9 +21,19 @@ public KaldbLocalQueryService(ChunkManager<T> chunkManager) {
     ScopedSpan span = Tracing.currentTracer().startScopedSpan("KaldbLocalQueryService.doSearch");
     SearchQuery query = SearchResultUtils.fromSearchRequest(request);
     CompletableFuture<SearchResult<T>> searchResult = chunkManager.query(query);
+    searchResult.whenComplete(
+        (res, ex) -> {
+          SpanUtils.addSearchResultMeta(span, res);
```
bryanlb, line 26: Since this hooks into the `whenComplete`, we expect [either the result, or the exception](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#whenComplete-java.util.function.BiConsumer-). As such we should probably cover the scenario where this completes exceptionally, as this would cause an NPE in the `addSearchResultMeta` call.

bryanlb (changes requested):

PR #182, opened 2021-12-17
don't swallow exception

vthacker:
A log line like this is no use for us

```
Service class com.slack.kaldb.chunkManager.IndexingChunkManager failed with cause com.slack.kaldb.metadata.zookeeper.InternalMetadataStoreException: Failed to delete node: SearchMetadata{...object_datails}
 
```

We have no idea what caused it. So let the logger print the entire exception

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/core/UpdatableCacheableMetadataStore.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/core/UpdatableCacheableMetadataStore.java
@@ -40,8 +40,7 @@ public abstract class UpdatableCacheableMetadataStore<T extends KaldbMetadata>
     try {
       update(metadataNode).get(KaldbConfig.DEFAULT_ZK_TIMEOUT_SECS, TimeUnit.SECONDS);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
-      logger.error("Error updating znode: " + metadataNode, e);
-      throw new InternalMetadataStoreException("Error updating znode: " + metadataNode, e);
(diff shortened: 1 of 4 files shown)


bryanlb (dismissed):

kaldb/src/main/java/com/slack/kaldb/metadata/core/UpdatableCacheableMetadataStore.java
```diff
@@ -40,7 +40,8 @@ public void updateSync(T metadataNode) {
     try {
       update(metadataNode).get(KaldbConfig.DEFAULT_ZK_TIMEOUT_SECS, TimeUnit.SECONDS);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
-      throw new InternalMetadataStoreException("Error updating node: " + metadataNode, e);
+      logger.error("Error updating znode: " + metadataNode, e);
```
bryanlb, line 43: Can we do this without the additional logs before throwing the exceptions? When every place does this up the stack trace it makes it difficult to understand, as there's a lot of duplicate error logs. It also skews the metrics as we count these multiple times.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/metadata/core/UpdatableCacheableMetadataStore.java
vthacker: I'll just revert the last commit - I guess I was okay with duplicate logs because not all callers can be audited to make sure the exception is being propagated but seeing multiple errors is inconvenient

bryanlb (approved):

PR #183, opened 2021-12-17
Rename CacheSlotAssignment > ReplicaAssignment

bryanlb:
This ensures consistency in naming now for all replica-related services (replica create / assign / evict / delete).


vthacker (approved):
+1 to the name change

PR #188, opened 2021-12-21
Snapshot deletion service

bryanlb:
Deletes snapshots and their associated blob storage objects that both exceed their configured lifespan, and have no associated replicas configured.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
@@ -10,6 +10,7 @@ import com.google.common.util.concurrent.AbstractScheduledService;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.blobfs.s3.S3BlobFs;
 import com.slack.kaldb.metadata.replica.ReplicaMetadataStore;
@@ -39,9 +40,13 @@ import org.slf4j.LoggerFactory;
  * have a corresponding snapshot object - removal of orphaned objects will be removed by a separate
  * service.
  */
+@SuppressWarnings("UnstableApiUsage")
 public class SnapshotDeletionService extends AbstractScheduledService {
   private static final Logger LOG = LoggerFactory.getLogger(SnapshotDeletionService.class);
+
   private static final int THREAD_POOL_SIZE = 10;
+  private static final int MAXIMUM_DELETES_PER_SECOND = 100;
+
   private final KaldbConfigs.ManagerConfig managerConfig;
 
   private final ReplicaMetadataStore replicaMetadataStore;
@@ -63,6 +68,7 @@ public class SnapshotDeletionService extends AbstractScheduledService {
       Executors.newFixedThreadPool(
           THREAD_POOL_SIZE,
           new ThreadFactoryBuilder().setNameFormat("snapshot-deletion-service-%d").build());
+  private final RateLimiter rateLimiter = RateLimiter.create(MAXIMUM_DELETES_PER_SECOND);
 
   public SnapshotDeletionService(
       ReplicaMetadataStore replicaMetadataStore,
@@ -164,7 +170,13 @@ public class SnapshotDeletionService extends AbstractScheduledService {
                   ListenableFuture<?> future =
                       Futures.submit(
                           () -> {
-                            // we first delete the snapshot reference - if for some reason we fail
+                            // These futures are rate-limited so that we can more evenly distribute
+                            // the load to the downstream services (metadata, s3). There is no
+                            // urgency to complete the deletes, so limiting the maximum rate allows
+                            // us to avoid unnecessary spikes.
+                            rateLimiter.acquire();
+
+                            // We first delete the snapshot reference - if for some reason we fail
                             // to delete the object from the object storage, the orphaned object
                             // cleanup service will attempt to delete it later
                             snapshotMetadataStore.deleteSync(snapshotMetadata);
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
@@ -1,0 +1,212 @@
+package com.slack.kaldb.clusterManager;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.util.concurrent.Futures.addCallback;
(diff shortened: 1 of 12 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -0,0 +1,212 @@
...
+   * would see that respected throughout the duration of their lifespan.
+   *
+   * <p>Deletion of expired snapshots should gracefully handle the scenario where immediately prior
+   * to reaching the end of the configured snapshot lifespan, an on-demand request is received. In
+   * this scenario a new replica would have been created, and the object would be currently
+   * downloading for servicing the request. The expectation from the end user would be that the
+   * request would still be served, as when the request was made it was still "in compliance" with
+   * the lifespan configuration.
+   *
+   * <p>When creating on-demand replicas, care should be taken to not create replicas for snapshots
```
mansu, line 131: To minimize the race conditions, I think it's best to delete an object a few more hours after expiration. This value can be hardcoded for now to 6 hours. In future, we can leave a TODO to make this value configurable.

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -0,0 +1,212 @@
...
+                        && !snapshotIdsWithReplicas.contains(snapshotMetadata.name))
+            .map(
+                snapshotMetadata -> {
+                  ListenableFuture<?> future =
+                      Futures.submit(
+                          () -> {
+                            // we first delete the snapshot reference - if for some reason we fail
+                            // to delete the object from the object storage, the orphaned object
+                            // cleanup service will attempt to delete it later
+                            snapshotMetadataStore.deleteSync(snapshotMetadata);
```
mansu, line 170: I think it may be better to delete this logic in 2 separate steps. Delete the metadata and then delete the S3 assets. Deleting assets in parallel may be subject to some rate limits, so it may be better to do it sequentially.

kaldb/src/test/java/com/slack/kaldb/clusterManager/SnapshotDeletionServiceTest.java
```diff
@@ -0,0 +1,798 @@
...
+
+    SnapshotDeletionService snapshotDeletionService =
+        new SnapshotDeletionService(
+            replicaMetadataStore, snapshotMetadataStore, s3BlobFs, managerConfig, meterRegistry);
+
+    int deletes = snapshotDeletionService.deleteExpiredSnapshotsWithoutReplicas();
+    assertThat(deletes).isEqualTo(0);
+
+    assertThat(snapshotMetadataStore.getCached()).containsExactlyInAnyOrder(snapshotMetadata);
+    verify(s3BlobFs, times(0)).delete(any(), anyBoolean());
```
mansu, line 326: Please add an additional check to ensure that the s3 path and the data exists?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
bryanlb: This isn't needed, as snapshots already have a configurable expiration (`snapshotLifespanMins: 10080`) that is significantly longer than that of replicas (`replicaLifespanMins: 1440`). This note is simply to call out that new replicas should not be created for expired snapshots that are pending deletion. 

This is something we should check for when making an on-demand request for a snapshot (ie, snapshot is not expired) - since this code isn't written yet though I went ahead and added a note here.

kaldb/src/test/java/com/slack/kaldb/clusterManager/SnapshotDeletionServiceTest.java
bryanlb: Done. Added additional checks for the s3BlobFs for every test that attempts to invoke delete with existing files. This shouldn't be strictly necessary since we verified `delete()` wasn't called, but in the event something were to change and an overloaded `delete()` implementation were to be called this should double check the file integrity.

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
bryanlb: As discussed, I went ahead and added the rate limiting to this so that we can reduce spikes on calls to the metadata and s3 services. This logic is fault tolerant of s3 failures, as the orphaned object cleanup will ensure that any failed deletes to s3 are retried until they are brought into compliance.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -0,0 +1,224 @@
...
+                            // These futures are rate-limited so that we can more evenly distribute
+                            // the load to the downstream services (metadata, s3). There is no
+                            // urgency to complete the deletes, so limiting the maximum rate allows
+                            // us to avoid unnecessary spikes.
+                            rateLimiter.acquire();
+
+                            // We first delete the snapshot reference - if for some reason we fail
+                            // to delete the object from the object storage, the orphaned object
+                            // cleanup service will attempt to delete it later
+                            snapshotMetadataStore.deleteSync(snapshotMetadata);
```
mansu, line 182: Let's delete the snapshot metadata after deleting the data on S3 so we can reduce the need for object clean up. That way on the next run, we can delete the missed files.

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -0,0 +1,224 @@
...
+ * value greater than the replica lifespan. This service only delete objects from blob storage that
+ * have a corresponding snapshot object - removal of orphaned objects will be removed by a separate
+ * service.
+ */
+@SuppressWarnings("UnstableApiUsage")
+public class SnapshotDeletionService extends AbstractScheduledService {
+  private static final Logger LOG = LoggerFactory.getLogger(SnapshotDeletionService.class);
+
+  private static final int THREAD_POOL_SIZE = 10;
+  private static final int MAXIMUM_DELETES_PER_SECOND = 100;
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
bryanlb: Done.

mansu (approved):

PR #190, opened 2021-12-22
Add list directory functionality to blobfs

bryanlb:
Adds functionality to list directories at a given path, utilizing the common prefix functionality of the AWS S3 sdk.

PR #193, opened 2022-01-06
Update s3mock dependency

bryanlb:
Updates s3mock to newer version. This was originally motivated out of https://github.com/slackhq/kaldb/pull/190#discussion_r774200759, as the currently used version has several mocks that do not function like the AWS api (getS3Object, prefixes, multipart uploads).

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -351,7 +351,7 @@
         <dependency>
             <groupId>com.adobe.testing</groupId>
             <artifactId>s3mock-junit4</artifactId>
-            <version>2.1.24</version>
+            <version>2.2.3</version>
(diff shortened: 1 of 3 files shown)


kaldb/pom.xml
```diff
@@ -351,7 +351,7 @@
         <dependency>
             <groupId>com.adobe.testing</groupId>
             <artifactId>s3mock-junit4</artifactId>
-            <version>2.1.24</version>
+            <version>2.2.3</version>
```
bryanlb, line 354: Newest is 2.3.3, but that requires us to update several other dependencies including micrometer.

mansu (approved):

PR #195, opened 2022-01-10
Refactory capacity reporting

bryanlb:
Report available slot capacity as a guage instead of a counter. Combine reporting for excess/insuficient capacity.


kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaAssignmentService.java
```diff
@@ -51,13 +51,13 @@
 
   public static final String REPLICA_ASSIGN_SUCCEEDED = "replica_assign_succeeded";
   public static final String REPLICA_ASSIGN_FAILED = "replica_assign_failed";
-  public static final String REPLICA_ASSIGN_INSUFFICIENT_CAPACITY =
-      "replica_assign_insufficient_capacity";
+  public static final String REPLICA_ASSIGN_AVAILABLE_CAPACITY =
+      "replica_assign_available_capacity";
```
vthacker, line 55: should we start adding `kaldb_` prefix?

bryanlb: I was thinking all as a single pass?

vthacker (approved):

PR #196, opened 2022-01-11
fix log message

vthacker:
tiny PR to ensure log message stack trace gets captured correctly.


bryanlb (commented):
We probably also want to fix the other occurrences on this in this file as well? Looks like line 154 does the same thing.

vthacker:
Closing in favour of a more long term fix -https://github.com/slackhq/kaldb/pull/197

PR #197, opened 2022-01-12
add forbiddenapis

vthacker:
https://github.com/policeman-tools/forbidden-apis will help us ban APIs that we consider error prone.

For example ban `LOG.error(String)` since every error log should propagate an exception

A lot of APIs are copied from lucene/solr codebase
see https://github.com/apache/lucene/tree/main/gradle/validation/forbidden-apis
and https://github.com/apache/solr/tree/main/gradle/validation/forbidden-apis

LoggingInfoStream was also added ( copied from Solr ) since we were currently printing to sys out without a logger which means it won't respect JSON logging, file rotations etc

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -11,6 +11,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <javac.target>11</javac.target>
+        <maven.compiler.target>11</maven.compiler.target>
         <uberjar.name>kaldb</uberjar.name>
         <http.client.version>4.5.13</http.client.version>
         <http.core.version>4.4.14</http.core.version>
@@ -451,6 +452,45 @@
             </extension>
         </extensions>
         <plugins>
+            <plugin>
+                <groupId>de.thetaphi</groupId>
+                <artifactId>forbiddenapis</artifactId>
+                <version>3.2</version>
+                <configuration>
+                    <!--
+                      if the used Java version is too new,
+                      don't fail, just do nothing:
+                    -->
+                    <failOnUnsupportedJava>false</failOnUnsupportedJava>
+                    <bundledSignatures>
+                        <!--
+                          This will automatically choose the right
+                          signatures based on 'maven.compiler.target':
+                        -->
+                        <bundledSignature>jdk-unsafe</bundledSignature>
+                        <bundledSignature>jdk-deprecated</bundledSignature>
+                        <!-- disallow undocumented classes like sun.misc.Unsafe: -->
+                        <bundledSignature>jdk-non-portable</bundledSignature>
+                        <!-- don't allow unsafe reflective access: -->
+                        <bundledSignature>jdk-reflection</bundledSignature>
+                        <bundledSignature>jdk-system-out</bundledSignature>
+                    </bundledSignatures>
+                    <signaturesFiles>
+                        <signaturesFile>./forbiddenapis/defaults.kaldb.txt</signaturesFile>
+                        <signaturesFile>./forbiddenapis/com.google.guava.guava.all.txt</signaturesFile>
+                        <signaturesFile>./forbiddenapis/commons-codec.commons-codec.all.txt</signaturesFile>
+                        <signaturesFile>./forbiddenapis/defaults.logging.txt</signaturesFile>
+                    </signaturesFiles>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>check</goal>
+                            <goal>testCheck</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
--- a/kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3BlobFs.java
+++ b/kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3BlobFs.java
@@ -17,6 +17,7 @@ import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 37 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -173,7 +174,10 @@ public boolean containsDataInTimeRange(long startTimeMs, long endTimeMs) {
     ensureTrue(
         endTimeMs - startTimeMs >= 0,
         String.format(
-            "end timestamp %d can't be less than the start timestamp %d.", endTimeMs, startTimeMs));
+            Locale.ROOT,
```
mansu, line 177: What are the advantages of using Locale.ROOT over leaving it null and let the system pick the default?

kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3BlobFs.java
```diff
@@ -227,7 +228,7 @@ private boolean copyFile(URI srcUri, URI dstUri) throws IOException {
   public boolean mkdir(URI uri) throws IOException {
     LOG.info("mkdir {}", uri);
     try {
-      Preconditions.checkNotNull(uri, "uri is null");
+      Objects.requireNonNull(uri, "uri is null");
```
mansu, line 231: Per guava [documentation](https://github.com/google/guava/wiki/PreconditionsExplained) it seems Preconditions.checkNotNull is still recommended. Do we still want to ban it?

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -187,18 +183,20 @@ public void addMessage(LogMessage message) {
...
         | UnSupportedPropertyTypeException
         | IllegalArgumentException e) {
-      LOG.error(String.format("Indexing message %s failed with error:", message), e);
+      LOG.error(String.format(Locale.ROOT, "Indexing message %s failed with error:", message), e);
       messagesFailedCounter.increment();
     } catch (IOException e) {
       // TODO: For now crash the program on IOException since it is likely a serious issue.
       // In future may need to handle this case more gracefully.
-      e.printStackTrace();
+      LOG.error("Error adding document ", e);
```
mansu, line 199: TODO should be updated to match the error.

kaldb/pom.xml
```diff
@@ -451,6 +452,45 @@
             </extension>
         </extensions>
         <plugins>
+            <plugin>
+                <groupId>de.thetaphi</groupId>
+                <artifactId>forbiddenapis</artifactId>
```
mansu, line 457: Is there a reason to prefer forbiddenapis library over a the more general checkstyle library?

kaldb/src/main/java/com/slack/kaldb/logstore/LogMessage.java
```diff
@@ -165,13 +167,13 @@ public boolean equals(Object o) {
...
-        && Objects.equal(id, that.id);
+        && Objects.equals(getIndex(), that.getIndex())
+        && Objects.equals(getType(), that.getType())
+        && Objects.equals(id, that.id);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hashCode(timeSinceEpochMilli, getIndex(), getMillisecondsSinceEpoch(), id);
+    return Objects.hash(timeSinceEpochMilli, getIndex(), getMillisecondsSinceEpoch(), id);
```
mansu, line 177: Can we please make sure that the there a unit test testing this logic and that the test passes after this change,

mansu (changes requested):

kaldb/pom.xml
vthacker: I think this library is configurable so we can add our own rules.

For example, I added a rule saying any `LOG.error` without a stack trace is banned - in our JSON logging setup we rely on the stack trace and index it onto a separate field . 

There was an instance while debugging prod restarts , the log was missing a stack trace. With this rule we'll fix that

Also I am pretty sure our current setup of google error prone is broken - there were talks to add this sort of feature to google error prone but it was never implemented - https://github.com/google/error-prone/issues/1211

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
vthacker: That's a good question

As per the jdocs

```
Useful constant for the root locale. The root locale is the locale whose language, country, and variant are empty ("") strings. This is regarded as the base locale of all locales, and is used as the language/country neutral locale for the locale sensitive operations.
Since:
1.6
```

and if we don't provide a locale `The locale used is the default locale for formatting for this instance of the Java virtual machine.` will get picked

So unsure, but this is the convention that lucene/solr and other projects follow so I kinda went along

kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3BlobFs.java
vthacker: Hmm! Not sure? We use both with just one message argument.

Anyways if you think this rule is unnecessary I can remove this rule and back out the change

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
vthacker: sorry I don't follow? What error should this match?

kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3BlobFs.java
bryanlb: Per the guava docs:
> Simple, varargs "printf-style" exception messages. (This advantage is also why we recommend continuing to use checkNotNull over Objects.requireNonNull)

mansu: Let's stick with the Preconditions API for now. Also, ban Objects.requireNotNull if it's easy to do.

mansu (changes requested):

PR #198, opened 2022-01-13
Add a DataType to all the messages.

mansu:
Add a type field to all messages so we can use embed other types easily in future.


bryanlb:
Per weekly team discussion we decided to defer this for now, and will keep the branch for reference. We will revisit this once we're ready to introduce our next data type.

PR #199, opened 2022-01-17
Include time unit in metadata proto fields.

mansu:
Include time unit in metadata proto fields so it's easier to add time in a consistent format everywhere.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
@@ -23,11 +23,11 @@ public class ChunkInfo {
   public static ChunkInfo fromSnapshotMetadata(SnapshotMetadata snapshotMetadata) {
     return new ChunkInfo(
         snapshotMetadata.snapshotId,
-        snapshotMetadata.startTimeUtc,
-        snapshotMetadata.endTimeUtc,
-        snapshotMetadata.startTimeUtc,
-        snapshotMetadata.endTimeUtc,
-        snapshotMetadata.endTimeUtc,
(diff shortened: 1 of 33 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -23,11 +23,11 @@
   public static ChunkInfo fromSnapshotMetadata(SnapshotMetadata snapshotMetadata) {
     return new ChunkInfo(
         snapshotMetadata.snapshotId,
-        snapshotMetadata.startTimeUtc,
-        snapshotMetadata.endTimeUtc,
-        snapshotMetadata.startTimeUtc,
-        snapshotMetadata.endTimeUtc,
-        snapshotMetadata.endTimeUtc,
+        snapshotMetadata.startTimeEpochMsUtc,
```
bryanlb, line 26: If we're going with "epoch" as part of the name we don't need to include "UTC" as well. 

I do like including the units in the name so it's abundantly clear.

bryanlb (changes requested):
I'd like to see us remove the "UTC" portion of the names as this is redundant with "epoch." 
We also should include the other proto definitions as well in `kladb_search` as they have yet another time unit format.

kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
mansu: Good call. I removed the utc from the names.

mansu:
This code is ready for re-review. Let's handle the other protos in a separate diff.

bryanlb (approved):

PR #202, opened 2022-01-20
increase chunk manager thread pool size

vthacker:
increase chunk manager thread pool - We're unable to add metrics https://github.com/slackhq/kaldb/pull/201 so just bumping up the number

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
@@ -15,11 +15,7 @@ import com.slack.kaldb.proto.config.KaldbConfigs;
 import com.spotify.futures.CompletableFutures;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -40,16 +36,19 @@
   // TODO: We want to move this to the config eventually
   // Less than KaldbDistributedQueryService#READ_TIMEOUT_MS
   public static final int QUERY_TIMEOUT_SECONDS = 10;
-  public static final int LOCAL_QUERY_THREAD_POOL_SIZE = 4;
```
mansu, line 43: Let's leave the static variables in place and them `MIN_QUERY_THREAD_POOL_SIZE`  and `MAX_QUERY_THREAD_POOL_SIZE`.

mansu (changes requested):

mansu (approved):

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -40,16 +36,21 @@
...
   */
   private static ExecutorService queryThreadPool() {
-    return Executors.newFixedThreadPool(
-        LOCAL_QUERY_THREAD_POOL_SIZE,
+    return new ThreadPoolExecutor(
+        MIN_QUERY_THREAD_POOL_SIZE,
+        MAX_QUERY_THREAD_POOL_SIZE,
+        60L,
+        TimeUnit.SECONDS,
+        new LinkedBlockingQueue<>(),
```
bryanlb, line 53: This should probably use a `SynchronousQueue`, or alternatively just use the `Executors.newCachedThreadPool`. Using an unbounded queue to back this executor will make it harder to debug, can be susceptible to lockups. 

The recommended default queuing strategy is to use a `SynchronousQueue` unless there is a specific need for one of the alternative queueing strategies.

PR #204, opened 2022-01-21
remove async code from query service

vthacker:
remove async code from query service 😢 - we use a top level blocking task executor so async is providing no value

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
@@ -13,6 +13,7 @@ import com.slack.kaldb.logstore.search.SearchResultAggregator;
 import com.slack.kaldb.logstore.search.SearchResultAggregatorImpl;
 import com.slack.kaldb.proto.config.KaldbConfigs;
 import com.spotify.futures.CompletableFutures;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.*;
@@ -56,7 +57,7 @@ public abstract class ChunkManager<T> extends AbstractIdleService {
    * 2. histogram over a fixed time range
    * We will not aggregate locally for future use-cases that have complex group by etc
    */
-  public CompletableFuture<SearchResult<T>> query(SearchQuery query) {
+  public SearchResult<T> query(SearchQuery query) throws IOException {
 
     SearchResult<T> errorResult =
         new SearchResult<>(new ArrayList<>(), 0, 0, new ArrayList<>(), 0, 0, 1, 0);
@@ -96,12 +97,16 @@ public abstract class ChunkManager<T> extends AbstractIdleService {
     // CompletableFuture.allOf and converting the CompletableFuture<Void> to
     // CompletableFuture<List<SearchResult>>
     CompletableFuture<List<SearchResult<T>>> searchResults = CompletableFutures.allAsList(queries);
-
-    // Increment the node count right at the end so that we increment it only once
-    //noinspection unchecked
-    return ((SearchResultAggregator<T>) new SearchResultAggregatorImpl<>(query))
-        .aggregate(searchResults)
-        .thenApply(this::incrementNodeCount);
+    try {
+      //noinspection unchecked
+      SearchResult<T> aggregatedResults =
+          ((SearchResultAggregator<T>) new SearchResultAggregatorImpl<>(query))
+              .aggregate(searchResults.get(30, TimeUnit.SECONDS));
+      return incrementNodeCount(aggregatedResults);
+    } catch (Exception e) {
+      LOG.error("Error searching across chunks ", e);
+      throw new IOException(e);
+    }
   }
 
   private SearchResult<T> incrementNodeCount(SearchResult<T> searchResult) {
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
@@ -13,7 +13,6 @@ import com.slack.kaldb.logstore.search.SearchResultAggregator;
 import com.slack.kaldb.logstore.search.SearchResultAggregatorImpl;
 import com.slack.kaldb.proto.config.KaldbConfigs;
 import com.spotify.futures.CompletableFutures;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.*;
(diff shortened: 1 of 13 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -151,21 +152,25 @@ private void updateStubs() {
...
+      List<SearchResult<LogMessage>> searchResults =
+          distributedSearch(request).get(30, TimeUnit.SECONDS);
+      SearchResult<LogMessage> aggregatedResult =
+          ((SearchResultAggregator<LogMessage>)
+                  new SearchResultAggregatorImpl<>(SearchResultUtils.fromSearchRequest(request)))
+              .aggregate(searchResults);
+      return SearchResultUtils.toSearchResultProto(aggregatedResult);
+    } catch (Exception e) {
+      LOG.error("Distributed search failed", e);
+      throw new IOException(e);
```
mansu, line 173: Can we throw a RuntimeException instead of a checked exception here?

mansu: Also, can we add some logging here so we know if doSearch is really recursive as they traces indicate it to be?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultAggregator.java
```diff
@@ -6,4 +6,6 @@
 public interface SearchResultAggregator<T> {
   CompletableFuture<SearchResult<T>> aggregate(
       CompletableFuture<List<SearchResult<T>>> searchResults);
+
+  SearchResult<T> aggregate(List<SearchResult<T>> searchResults);
```
mansu, line 10: Please add a comment on when we need to use one over the other.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -101,12 +102,16 @@ private static ExecutorService queryThreadPool() {
     // CompletableFuture.allOf and converting the CompletableFuture<Void> to
     // CompletableFuture<List<SearchResult>>
     CompletableFuture<List<SearchResult<T>>> searchResults = CompletableFutures.allAsList(queries);
```
mansu, line 104: Is CompletableFutures abstraction still needed after this change?

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -101,12 +102,16 @@ private static ExecutorService queryThreadPool() {
...
-        .thenApply(this::incrementNodeCount);
+    try {
+      //noinspection unchecked
+      SearchResult<T> aggregatedResults =
+          ((SearchResultAggregator<T>) new SearchResultAggregatorImpl<>(query))
+              .aggregate(searchResults.get(30, TimeUnit.SECONDS));
+      return incrementNodeCount(aggregatedResults);
+    } catch (Exception e) {
+      LOG.error("Error searching across chunks ", e);
+      throw new IOException(e);
```
mansu, line 113: Do we need to throw a checked exception here?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
vthacker: If we throw Runtime exception how would you suggest catching it in `KaldbQueryServiceBase` ?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: We can work on logging and other tracing additions in another issue - I think this PR we just want to remove async code

vthacker: RE - runtime exception , I left a comment on the previous conversation as it's the same

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
vthacker: frixed

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultAggregator.java
vthacker: fixed - removed it so we don't have two method :)

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -105,7 +108,7 @@ private void updateStubs() {
           addedStubs.get(),
           removedStubs.get());
     } catch (Exception e) {
-      LOG.error("Cannot update SearchMetadata cache on the query service" + e);
+      LOG.error("Cannot update SearchMetadata cache on the query service", e);
```
bryanlb, line 111: Nice.

kaldb/src/main/java/com/slack/kaldb/server/KaldbQueryServiceBase.java
```diff
@@ -19,23 +18,13 @@ public void search(
...
-                responseObserver.onCompleted();
-              }
-              return null;
-            });
+    try {
+      responseObserver.onNext(doSearch(request));
+    } catch (Exception e) {
+      responseObserver.onError(e);
+    }
+    responseObserver.onCompleted();
```
bryanlb, line 26: Should we still call `onCompleted` when an error occurs? It appears previously `onCompleted` only executed when there was no error.

vthacker: done

kaldb/src/test/java/com/slack/kaldb/chunk/IndexingChunkManagerTest.java
```diff
@@ -330,12 +329,13 @@ private void testChunkManagerSearch(
       int totalSnapshots,
       int expectedSnapshotsWithReplicas,
       long startTimeEpochMs,
-      long endTimeEpochMs) {
+      long endTimeEpochMs)
+      throws IOException {
```
mansu, line 333: Do we still need these IOExceptions?

vthacker: done

mansu (approved):

PR #209, opened 2022-01-24
Refactor chunk hydration instrumentation

bryanlb:
Replaces existing set of four counters with two timers, and uses tags to separate success from failure. This will allow us to see how long it takes to process an assignment/eviction, as well as compare them across success and failures.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -3,7 +3,6 @@ package com.slack.kaldb.chunk;
 import static com.slack.kaldb.logstore.BlobFsUtils.copyFromS3;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.blobfs.s3.S3BlobFs;
 import com.slack.kaldb.logstore.search.LogIndexSearcher;
@@ -21,14 +20,12 @@ import com.slack.kaldb.metadata.snapshot.SnapshotMetadata;
 import com.slack.kaldb.metadata.snapshot.SnapshotMetadataStore;
 import com.slack.kaldb.metadata.zookeeper.MetadataStore;
 import com.slack.kaldb.proto.metadata.Metadata;
-import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Timer;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Instant;
-import java.util.Collection;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -202,14 +204,13 @@ private synchronized void handleChunkAssignment(CacheSlotMetadata cacheSlotMetad
       }
 
       registerSearchMetadata(snapshotMetadata.name);
-      successfulChunkAssignments.increment();
+      assignmentTimer.stop(chunkAssignmentTimerSuccess);
```
mansu, line 207: timers should always be closed to capture the elapsed time. Otherwise it results in weird metrics data. Further, counting success and failures from timers is an odd pattern in metrics.

I think it's better to leave the counters as is and add a timer to capture the time for chunk assignment and chunk evictions.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: For timer samples, there's no need (nor option) to close them - [per the documentation](https://micrometer.io/docs/concepts#_storing_start_state_in_timer_sample) the expectation is that after stopping a timer sample it will go out of scope.

If we want I can keep the existing counters, but the [micrometer documentation](https://micrometer.io/docs/concepts#_counters) specifically calls this out as an anti-pattern and recommends against using the combination of timers and counters:

> Never count something you can time with a Timer or summarize with a DistributionSummary! Both Timer and DistributionSummary always publish a count of events in addition to other measurements.
> 
> Be sure to read through the timer section before jumping into using counters, as timers record a count of timed events as part of the suite of metrics that go into timing. For those pieces of code you intend to time, you do NOT need to add a separate counter.

Lastly, encoding the status (ie success, fail) is the approach suggested by micrometer for [various](https://micrometer.io/docs/concepts#_multi_gauge) [types](https://micrometer.io/docs/concepts#_storing_start_state_in_timer_sample) of instrumentation.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
```diff
@@ -145,7 +147,8 @@ public void shouldHandleChunkLivecycle() throws Exception {
                 500,
                 0));
     assertThat(logMessageSearchResult.hits.size()).isEqualTo(10);
-    assertThat(readOnlyChunk.successfulChunkAssignments.count()).isEqualTo(1);
+    assertThat(meterRegistry.get(CHUNK_ASSIGNMENT_TIMER).tag("successful", "true").timer().count())
```
mansu, line 150: Please add an assert for the failure timer count and eviction timer count to make sure that both those timers are not reported at the same time.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
mansu: Didn't realize Timer also emitted counts. Thought it was only the distribution summary. Thanks for the doc link.

mansu (changes requested):
Once the asserts are added this code is good to go.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
bryanlb: Done.

mansu (approved):

PR #213, opened 2022-01-26
Add logging and some span tags on the query path.

mansu:
Add logging and some span tags on the query path to make debugging easier.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -119,11 +119,13 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
   }
 
   private List<SearchResult<LogMessage>> distributedSearch(KaldbSearch.SearchRequest request) {
+    LOG.info("Starting distributed search for request: {}", request);
     ScopedSpan span =
         Tracing.currentTracer().startScopedSpan("KaldbDistributedQueryService.distributedSearch");
 
     try {
       List<ListenableFuture<SearchResult<LogMessage>>> queryServers = new ArrayList<>(stubs.size());
+      span.tag("stub_count", String.valueOf(stubs.size()));
 
       for (KaldbServiceGrpc.KaldbServiceFutureStub stub : stubs.values()) {
         ListenableFuture<KaldbSearch.SearchResult> searchRequest =
@@ -149,6 +151,7 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
       span.error(e);
       return List.of(SearchResult.empty());
     } finally {
+      LOG.info("Finished distributed search for request: {}", request);
       span.finish();
     }
   }
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -3,7 +3,6 @@ package com.slack.kaldb.logstore.search;
 import brave.ScopedSpan;
 import brave.Tracing;
 import brave.grpc.GrpcTracing;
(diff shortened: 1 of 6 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -119,11 +119,13 @@ private void updateStubs() {
   }
 
   private List<SearchResult<LogMessage>> distributedSearch(KaldbSearch.SearchRequest request) {
+    LOG.info("Starting distributed search for request: {}", request);
     ScopedSpan span =
         Tracing.currentTracer().startScopedSpan("KaldbDistributedQueryService.distributedSearch");
 
     try {
       List<ListenableFuture<SearchResult<LogMessage>>> queryServers = new ArrayList<>(stubs.size());
+      span.tag("stub_count", String.valueOf(stubs.size()));
```
bryanlb, line 128: Tagging this as a stub count is somewhat awkward, since stubs is just a grpc specific detail. What if we referred to this as `queryServerCount`? We should also change this to camel cased, since all the other tags use that format.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -119,11 +119,13 @@ private void updateStubs() {
   }
 
   private List<SearchResult<LogMessage>> distributedSearch(KaldbSearch.SearchRequest request) {
+    LOG.info("Starting distributed search for request: {}", request);
```
bryanlb, line 122: Is the string representation of the `request` readable, or are we better off pulling out the fields we're interested in directly?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbLocalQueryService.java
```diff
@@ -16,12 +19,17 @@ public KaldbLocalQueryService(ChunkManager<T> chunkManager) {
...
   @Override
   public KaldbSearch.SearchResult doSearch(KaldbSearch.SearchRequest request) {
+    LOG.info("Received search request: {}", request);
     ScopedSpan span = Tracing.currentTracer().startScopedSpan("KaldbLocalQueryService.doSearch");
     SearchQuery query = SearchResultUtils.fromSearchRequest(request);
+    span.tag("query", query.toString());
     SearchResult<T> searchResult;
     searchResult = chunkManager.query(query);
     KaldbSearch.SearchResult result = SearchResultUtils.toSearchResultProto(searchResult);
+    span.tag("total_nodes", String.valueOf(result.getTotalNodes()));
```
bryanlb, line 29: Same as above, both of these should be camel cased

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbLocalQueryService.java
mansu: Good point. Fixed.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
mansu: It's readable. Just logged a sample log line from unit tests. 

```
[INFO ] 2022-01-26 14:06:20.339 [main] KaldbLocalQueryService - Received search request: index_name: "testindex"
query_string: "Message1"
start_time_epoch_ms: 1601547000000
end_time_epoch_ms: 1601547010000
how_many: 10
bucket_count: 2
```

mansu: We can start with this and see how it does.

bryanlb (dismissed):

bryanlb (approved):

PR #216, opened 2022-01-28
Fix snapshot deletion timeout

bryanlb:
Since we introduced rate limiting to slow down these calls, we need to increase the overall allowable time to complete the deletions.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
@@ -2,7 +2,6 @@ package com.slack.kaldb.clusterManager;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.util.concurrent.Futures.addCallback;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -90,6 +89,12 @@ public SnapshotDeletionService(
     this.s3BlobFs = s3BlobFs;
     this.meterRegistry = meterRegistry;
 
+    // This functions as the overall "timeout" for deleteExpiredSnapshotsWithoutReplicas, and should
+    // not exceed that of the schedule period. This ensures that we never enter a situation where
+    // we are queuing faster than we are draining.
+    this.futuresListTimeoutSecs =
```
vthacker, line 95: would it be safer if this is the value -1 or something?

vthacker (approved):
+1

Checked with Bryan the timeout == the scheduler run is okay. i.e we don't need to be slightly lower than the timeout value since it's one thread that runs the execution

PR #218, opened 2022-01-31
Add extra logging to the snapshot deletion

bryanlb:
This adds additional logging before attempting to delete snapshots of what will be attempted to be deleted, from both S3 and ZK.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
@@ -187,6 +187,7 @@ public class SnapshotDeletionService extends AbstractScheduledService {
                               // the
                               // metadata and try again on the next run.
                               URI snapshotUri = URI.create(snapshotMetadata.snapshotPath);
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -187,6 +187,7 @@ protected int deleteExpiredSnapshotsWithoutReplicas() {
                               // the
                               // metadata and try again on the next run.
                               URI snapshotUri = URI.create(snapshotMetadata.snapshotPath);
+                              LOG.info("Starting delete of snapshot {}, snapshot URI of {}", snapshotMetadata.name, snapshotMetadata.snapshotPath);
```
mansu, line 190: I think it may be better to log the entire snapshot metadata here to help with other debugging scenarios.

bryanlb: Done. There should be relatively infrequent amounts of snapshots being deleted (1k / week) that this isn't ever an issue if we decide to permanently keep it.

PR #219, opened 2022-01-31
Improve snapshot deletion for expired LIVE snapshots

bryanlb:
Now will attempt to delete the expired LIVE snapshots instead of throwing a blobfs exception.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
@@ -13,6 +13,7 @@ import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.blobfs.s3.S3BlobFs;
 import com.slack.kaldb.metadata.replica.ReplicaMetadataStore;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/SnapshotDeletionService.java
```diff
@@ -184,11 +185,11 @@ protected int deleteExpiredSnapshotsWithoutReplicas() {
 
                               // First try to delete the object from S3, then delete from metadata
                               // store. If for some reason the object delete fails, it will leave
-                              // the
-                              // metadata and try again on the next run.
+                              // the metadata and try again on the next run.
                               URI snapshotUri = URI.create(snapshotMetadata.snapshotPath);
                               LOG.info("Starting delete of snapshot {}", snapshotMetadata);
-                              if (s3BlobFs.exists(snapshotUri)) {
+                              if (!SnapshotMetadata.isLive(snapshotMetadata)
```
mansu, line 191: I think it's better to contain the stale snapshot deletion logic in the indexer since it would be hard to keep both  of these pieces of code in sync and not cause a race condition. So for now, I think it's best to move the isLive check into the filter above and have this logic only deal with the non-live snapshots.

PR #221, opened 2022-02-01
Fix JSON Logging Issues

vthacker:
- Explicitly add traceId/spanId to the logs. What was there currently wasn't working. Deployed and checked this works
- Don't rely on the explicit truncation symbol ( https://unicode-table.com/en/2026/ ) by explicitly setting it ( this wasn't causing any of the parse failures but still beneficial to set it ) 
- Rename the `error.foo` with underscores . `error.foo` was being interpreted as a nested object by ES and failing to get indexed

### DIFF
--- a/kaldb/src/main/resources/KaldbLayout.json
+++ b/kaldb/src/main/resources/KaldbLayout.json
@@ -22,38 +22,40 @@
     "$resolver": "logger",
     "field": "name"
   },
-  "labels": {
+  "traceId": {
     "$resolver": "mdc",
-    "flatten": true,
+    "key": "traceId",
     "stringified": true
   },
-  "tags": {
-    "$resolver": "ndc"
+  "spanId": {
+    "$resolver": "mdc",
+    "key": "spanId",
+    "stringified": true
   },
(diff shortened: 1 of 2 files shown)


mansu (approved):

kaldb/src/main/resources/KaldbLayout.json
```diff
@@ -22,38 +22,40 @@
...
     "$resolver": "exception",
     "field": "className"
   },
-  "error.message": {
+  "error_message": {
     "$resolver": "exception",
     "field": "message"
   },
-  "error.stack_trace": {
+  "error_stack_trace": {
```
mansu, line 43: Instead of `error_stack_trace` would `error:stack_trace` work?

vthacker: `:` is a lucene delimiter for fieldname:fieldvalue query in lucene.

So I don't think `error:stack_trace` will work

mansu: Sounds good!

PR #224, opened 2022-02-04
Disable index filtering at query layer

bryanlb:
Until support is added at the snapshot/query layer this should be removed so that multiple indexes can be searched at once.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
@@ -1,6 +1,9 @@
 package com.slack.kaldb.logstore.search;
 
-import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.*;
+import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.COMMITS_COUNTER;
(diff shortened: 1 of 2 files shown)


kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -37,8 +41,10 @@ public static void beforeClass() throws Exception {
   private void loadTestData(Instant time) {
     strictLogStore.logStore.addMessage(
         makeMessageWithIndexAndTimestamp(1, "apple", TEST_INDEX_NAME, time));
-    strictLogStore.logStore.addMessage(
-        makeMessageWithIndexAndTimestamp(2, "baby", "new" + TEST_INDEX_NAME, time.plusSeconds(1)));
+
+    // todo - re-enable when multi-tenancy is supported
```
mansu, line 45: Please add the issue number to this class also so we can refer back to it.

mansu (changes requested):

vthacker (approved):

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
bryanlb: Done.

mansu (approved):

PR #227, opened 2022-02-07
Refactor query timeouts

bryanlb:
Refactors timeouts such that we always now attempt to return a successful response. Also introduces future cancellations when possible.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
@@ -1,5 +1,7 @@
 package com.slack.kaldb.chunkManager;
 
+import static com.slack.kaldb.config.KaldbConfig.LOCAL_QUERY_TIMEOUT_DURATION;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
@@ -12,7 +14,11 @@ import com.slack.kaldb.logstore.search.SearchResultAggregatorImpl;
 import com.spotify.futures.CompletableFutures;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -29,10 +35,6 @@ public abstract class ChunkManager<T> extends AbstractIdleService {
   // we use a CopyOnWriteArrayList as we expect to have very few edits to this list compared
   // to the amount of reads, and it must be a threadsafe implementation
   protected final List<Chunk<T>> chunkList = new CopyOnWriteArrayList<>();
-
-  // TODO: Move this config to Kaldb config.
-  // Less than KaldbDistributedQueryService#READ_TIMEOUT_MS
-  public static final int QUERY_TIMEOUT_SECONDS = 10;
 
   private static final ExecutorService queryExecutorService = queryThreadPool();
 
@@ -70,7 +72,7 @@ public abstract class ChunkManager<T> extends AbstractIdleService {
                             RequestContext.makeContextPropagating(queryExecutorService))
                         // TODO: this will not cancel lucene query. Use ExitableDirectoryReader in
                         // the future and pass this timeout
-                        .orTimeout(QUERY_TIMEOUT_SECONDS, TimeUnit.SECONDS))
+                        .orTimeout(LOCAL_QUERY_TIMEOUT_DURATION.toMillis(), TimeUnit.MILLISECONDS))
             .map(
                 chunkFuture ->
                     chunkFuture.exceptionally(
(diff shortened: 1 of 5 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -91,16 +91,21 @@ private static ExecutorService queryThreadPool() {
...
       //noinspection unchecked
       SearchResult<T> aggregatedResults =
           ((SearchResultAggregator<T>) new SearchResultAggregatorImpl<>(query))
-              .aggregate(searchResults.get(30, TimeUnit.SECONDS));
+              .aggregate(searchResults);
       return incrementNodeCount(aggregatedResults);
     } catch (Exception e) {
       LOG.error("Error searching across chunks ", e);
       throw new RuntimeException(e);
+    } finally {
```
mansu, line 107: Please add a comment on why we should cancel this future here so in future, we don't delete the line by mistake.

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
```diff
@@ -25,6 +26,14 @@
   public static Duration DEFAULT_START_STOP_DURATION = Duration.ofSeconds(15);
   public static final int DEFAULT_ZK_TIMEOUT_SECS = 15;
 
+  // Timeouts are structured such that we always attempt to return a successful response, as we
+  // include metadata that should always be present.
+  public static final Duration ARMERIA_TIMEOUT = Duration.of(15, ChronoUnit.SECONDS);
```
mansu, line 31: Renaming `ARMERIA_TIMEOUT` to `ARMERIA_QUERY_TIMEOUT_SECS` is more consistent with other fields. Also, please add time units to other timeout values.

Please expand comment explaining where various timeouts are used and why different timeouts are smaller than others.

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -102,6 +105,10 @@ private void addCompression(ServerBuilder sb) {
     sb.decorator(EncodingService.builder().newDecorator());
   }
 
+  private void setTimeout(ServerBuilder sb) {
```
mansu, line 108: inline?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,25 +127,26 @@ private void updateStubs() {
...
-                .withInterceptors(
-                    GrpcTracing.newBuilder(Tracing.current()).build().newClientInterceptor())
-                .search(request);
-        Function<KaldbSearch.SearchResult, SearchResult<LogMessage>> searchRequestTransform =
-            SearchResultUtils::fromSearchResultProtoOrEmpty;
-        queryServers.add(
-            Futures.transform(
-                searchRequest, searchRequestTransform::apply, MoreExecutors.directExecutor()));
-      }
+    List<ListenableFuture<SearchResult<LogMessage>>> queryServers = new ArrayList<>(stubs.size());
```
mansu, line 130: Do the unit tests test these time outs?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
bryanlb: Yes, in `KaldbDistributedQueryServiceTest.testSearchWithOneShardTimeout`

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: I left this as a separate method because it matches what we did for `addCompression` and I believe we'll add additional timeout / exception handling. I went ahead and added a todo along with the mention about including a `Retry-After` based on our earlier Slack discussion, and referenced the api we should be able to use to achieve this.

kaldb/src/main/java/com/slack/kaldb/config/KaldbConfig.java
bryanlb: Done. Per discussion used `_DURATION` for all these fields and added the additional documentation.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
bryanlb: Done - both here and on the other future cancellation section.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -68,9 +70,8 @@ private static ExecutorService queryThreadPool() {
                     CompletableFuture.supplyAsync(
                             () -> chunk.query(query),
                             RequestContext.makeContextPropagating(queryExecutorService))
-                        // TODO: this will not cancel lucene query. Use ExitableDirectoryReader in
-                        // the future and pass this timeout
-                        .orTimeout(QUERY_TIMEOUT_SECONDS, TimeUnit.SECONDS))
+                        .orTimeout(
+                            LOCAL_QUERY_TIMEOUT_DURATION.toMillis() - 100, TimeUnit.MILLISECONDS))
```
mansu, line 74: We should still keep the TODO since we don't cancel the underlying lucene query yet.

Also, we should also extract `LOCAL_QUERY_TIMEOUT_DURATION.toMillis() - 100` into a local variable and add a comment on why it's needed. Also, please document the magic number 100 that way we remember why we set it to that number later and what it can be updated to.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,25 +128,26 @@ private void updateStubs() {
...
+    span.tag("queryServerCount", String.valueOf(stubs.size()));
+
+    for (KaldbServiceGrpc.KaldbServiceFutureStub stub : stubs.values()) {
+      ListenableFuture<KaldbSearch.SearchResult> searchRequest =
+          stub.withDeadlineAfter(READ_TIMEOUT_MS - 100, TimeUnit.MILLISECONDS)
```
mansu, line 136: Please extract LOCAL_QUERY_TIMEOUT_DURATION.toMillis() - 100 into a local variable and add a comment on why it's needed. Also, please document the magic number 100 that way we remember why we set it to that number later and what it can be updated to.

mansu (changes requested):
A couple of minor nits. Sorry, missed this earlier.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
bryanlb: Ah, these were inadvertently included from a previously staged commit and are not needed. I've removed both.

mansu (approved):

PR #230, opened 2022-02-11
Improve tracing instrumentation

bryanlb:
* Add openzipkin as an example config + docker compose setup
* Removes unused spans
* Removes annotations as support is limited outside of zipkin
* Fixes trace inheritance issue with completable futures
* Adds additional span tags to top-level request/response objects

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
@@ -2,6 +2,9 @@ package com.slack.kaldb.chunkManager;
 
 import static com.slack.kaldb.config.KaldbConfig.LOCAL_QUERY_TIMEOUT_DURATION;
 
+import brave.Span;
+import brave.Tracing;
+import brave.propagation.TraceContext;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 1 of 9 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -59,6 +63,9 @@ private static ExecutorService queryThreadPool() {
     SearchResult<T> errorResult =
         new SearchResult<>(new ArrayList<>(), 0, 0, new ArrayList<>(), 0, 0, 1, 0);
 
+    // https://stackoverflow.com/questions/69647913/how-to-propagate-brave-trace-in-lettuce-client-async-command
```
mansu, line 66: Since, this is not a bug, instead of a stack overflow link can you please add a link to brave documentation here if possible.

vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -68,7 +75,10 @@ private static ExecutorService queryThreadPool() {
             .map(
                 (chunk) ->
                     CompletableFuture.supplyAsync(
-                            () -> chunk.query(query),
+                            () -> {
+                              Tracing.current().currentTraceContext().maybeScope(context);
+                              return chunk.query(query);
+                            },
```
mansu, line 81: We should also clear the context after the query here. Prefer to do it a finally block so it's more robust. Otherwise, if some other function uses the same thread pool we will get weird traces and those issues are hard to debug..

mansu (changes requested):

PR #232, opened 2022-02-15
experiment with rate limiting S3 download

vthacker:


kaldb/src/main/java/com/slack/kaldb/blobfs/s3/RateLimitingDelegatingInputStream.java
```diff
@@ -0,0 +1,41 @@
+package com.slack.kaldb.blobfs.s3;
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.lucene.store.RateLimiter;
```
bryanlb, line 6: We probably should use the Guava rate limiter, as that's what we're using throughout the rest of the code base.

PR #235, opened 2022-02-24
query node should only reach out to relevant index and cache nodes for data

vthacker:
Don't query every indexer and cache node

Only query search nodes that have at-least 1 snapshot that belongs in the time range and a service partition window.


kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
```diff
@@ -17,6 +19,9 @@
  * <p>TODO: Have a read only chunk info for read only chunks so we don't accidentally update it.
  */
 public class ChunkInfo {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ChunkInfo.class);
```
mansu, line 23: Can we drop this since it's unused?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +128,42 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+                        snapshotMetadata.endTimeEpochMs,
+                        startTimeMs,
+                        endTimeMs))
+            .map(snapshotMetadata -> snapshotMetadata.name)
+            .collect(Collectors.toSet());
+
+    return searchMetadataStore
+        .getCached()
+        .stream()
+        .filter(searchMetadata -> snapshotsToSearch.contains(searchMetadata.snapshotName))
```
mansu, line 150: If a snapshot has multiple replicas, would we query both replicas in this case?

How do we handle the case when the snapshot exists but is not present in the cache?

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
```diff
@@ -0,0 +1,132 @@
...
+  public void tearDown() throws Exception {
+    snapshotMetadataStore.close();
+    searchMetadataStore.close();
+    zkMetadataStore.close();
+    metricsRegistry.close();
+    testZKServer.close();
+  }
+
+  @Test
+  public void testSnapshotsToSearch() throws Exception {
```
mansu, line 75: Please add a few more unit tests to this code. 

* A case when data is live in indexer but no where else.
* A case when data is in indexer and cache.
* A cache when the same node has no live nodes, but multiple replicas in cache nodes. 
* A unit test where no nodes exist for a given query.
* A unit test to ensure if multiple services exist, we only pick the snapshots for a single service.
* A unit test to ensure that we are picking nodes randomly across multiple calls to the method.
* A unit test to document what happens to distributed query when this function fails.

kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
```diff
@@ -53,16 +53,17 @@ public class KaldbTest {
   private static final String KALDB_TEST_CLIENT_1 = "kaldb-test-client1";
   private static final String KALDB_TEST_CLIENT_2 = "kaldb-test-client2";
 
-  private static KaldbSearch.SearchResult searchUsingGrpcApi(String queryString, int port) {
+  private static KaldbSearch.SearchResult searchUsingGrpcApi(
```
mansu, line 56: Is there a way we can check here that we are only hitting the nodes we want to hit when performing the distributed search?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +131,67 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+        // ensure we randomize the list so that when there are multiple
+        // search metadata nodes with the same snapshot we don't always pick the same one
+        .collect(
+            Collectors.toMap(
+                SearchMetadata::getSnapshotName,
+                SearchMetadata::getUrl,
+                (firstUrl, secondUrl) -> random.nextBoolean() ? firstUrl : secondUrl))
```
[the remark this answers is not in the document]
vthacker: I had done that as the first approach. But that required a deep copy of the list so I thought this might be better off.

I can revert to shuffling like I had in this commit - https://github.com/slackhq/kaldb/pull/235/commits/eebbc330388b55820100afa34061831f419a176b

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +131,67 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+                snapshotMetadata ->
+                    containsDataInTimeRange(
+                        snapshotMetadata.startTimeEpochMs,
+                        snapshotMetadata.endTimeEpochMs,
+                        startTimeMs,
+                        endTimeMs))
+            .map(snapshotMetadata -> snapshotMetadata.name)
+            .collect(Collectors.toSet());
+
+    // TODO: in the future, we'd have another field in search metadata to indicate if it's hosted by
```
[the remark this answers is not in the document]
vthacker: SearchMetada doesn't have an `isLive` today

We can perhaps infer from `snapshotName` since it will start with `LIVE_something`
```
  public final String snapshotName;
  public final String url;
```

But Bryan and I were thinking SearchMetada in the future could have another field which represents am I a cache node, am I an indexer node. And even associate a query priority score to the node so that it's not "random" but we would prefer one over the other in a more nuanced way

let me know if you think we should do `live` vs non live for now?

kaldb/src/test/java/com/slack/kaldb/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
vthacker: > A case when data is live in indexer but no where else.

Isn't that the very first test here?

> A case when data is in indexer and cache.

That's the second test

vthacker: let me break it down to multiple methods so it's easier for you to review

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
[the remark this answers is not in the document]
mansu: > .filter(...)
> .groupBy(snapshotName)
> .map(list -> list.size() > 1 ? list.get(0) : list.get(random(id)))
> .collect()

kaldb/src/test/java/com/slack/kaldb/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
```diff
@@ -0,0 +1,293 @@
...
+    assertThat(snapshots.iterator().next()).isEqualTo(indexer1SearchContext.toString());
+
+    snapshots =
+        getSnapshotsToSearch(
+            snapshotMetadataStore,
+            searchMetadataStore,
+            chunkEndTime.toEpochMilli() + 1,
+            chunkEndTime.toEpochMilli() + 100,
+            Set.of("1"));
+    assertThat(snapshots.size()).isEqualTo(0);
```
mansu, line 107: In addition to size, we should also check the data in the snapshot to make sure that the data we get is the one we expect for all the unit tests.

kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
mansu: In KaldbDistributedQueryServiceTest, we are just ensuring that function works as expected given the metadata in the store. 

It would be good to also have an end to end test where some nodes are hit and some nodes are not hit. Also, would be good to add a check to make sure that we are making one search request per chunk.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +137,105 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+
+    // step 2 - iterate every snapshot and map it to the search metadata while de-duplicating
+    // search metadata with the same snapshot name
+    return searchMetadataStore
+        .getCached()
+        .stream()
+        .filter(searchMetadata -> snapshotsToSearch.contains(searchMetadata.snapshotName))
+        .collect(Collectors.groupingBy(searchMetadata -> searchMetadata.snapshotName))
+        .values()
+        .stream()
```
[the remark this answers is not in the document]
vthacker: do you have an example of how it would work if we dropped the stream?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +137,105 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+        .filter(queryableSearchMetadataNodes -> queryableSearchMetadataNodes.size() > 0)
```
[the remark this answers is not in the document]
vthacker: just safety.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +137,105 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+        .map(KaldbDistributedQueryService::pickBestSearchMetadataNode)
+        .map(searchMetadata -> searchMetadata.url)
+        .collect(Collectors.toList());
+  }
+
+  @VisibleForTesting
+  public static SearchMetadata pickBestSearchMetadataNode(
+      List<SearchMetadata> queryableSearchMetadataNodes) {
+    if (queryableSearchMetadataNodes.size() == 1) {
```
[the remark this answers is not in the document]
vthacker: what benefit does that give?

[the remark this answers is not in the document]
vthacker: from my perspective, it's just tougher to read if it's in the `map` function in the caller.

kaldb/src/test/java/com/slack/kaldb/com/slack/kaldb/logstore/search/KaldbDistributedQueryServiceTest.java
vthacker: done!

kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
vthacker: we do that in KaldbTest.java

vthacker: it's in KaldbTest.java

bryanlb (dismissed):

kaldb/src/main/java/com/slack/kaldb/metadata/search/SearchMetadata.java
```diff
@@ -3,43 +3,48 @@
...
   public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     if (!super.equals(o)) return false;
-
     SearchMetadata that = (SearchMetadata) o;
-
-    if (!snapshotName.equals(that.snapshotName)) return false;
-    return url.equals(that.url);
+    return isLive == that.isLive && snapshotName.equals(that.snapshotName) && url.equals(that.url);
```
mansu, line 42: Is this code auto-generated?

mansu (changes requested):
The code looks great and is very close. There is a small refactor and some minor nits. After this the code is good to go.

kaldb/src/main/java/com/slack/kaldb/metadata/search/SearchMetadata.java
vthacker: yep! I had intellij do it for me.

kaldb/src/main/proto/metadata.proto
```diff
@@ -71,6 +71,11 @@ message SearchMetadata {
 
   // url
   string url = 3;
+
+  // is the snapshot hosted by a indexer node that is currently indexing data
+  // if yes then live=true else live=false
+  // live=false is always false for replicas hosted by cache nodes
+  bool isLive = 4;
```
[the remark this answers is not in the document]
vthacker: I guess I am using `isLive` as a way to say this is a snapshot being served by the index node.

When a chunk is rolled over, LIVE is gone from the prefix name. Hence we need this `isLive`

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +136,135 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+      ServiceMetadataStore serviceMetadataStore,
+      long startTimeEpochMs,
+      long endTimeEpochMs,
+      String indexName) {
+    return serviceMetadataStore
+        .getCached()
+        .stream()
+        .filter(serviceMetadata -> serviceMetadata.name.equals(indexName))
+        .flatMap(serviceMetadata -> serviceMetadata.partitionConfigs.stream())
+        .filter(
```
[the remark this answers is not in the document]
vthacker: I think the flatMap is deliberate to map serviceMetadata to it's partition configs. 

Do you wanna try it and see if it works? I'm not sure how it would work if I swap it like you suggested.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +136,135 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+  }
+
+  public static boolean isSnapshotInPartition(
+      SnapshotMetadata snapshotMetadata,
+      ServiceMetadataStore serviceMetadataStore,
+      long queryStartTimeEpochMs,
+      long queryEndTimeEpochMs,
+      String indexName) {
+    List<ServicePartitionMetadata> partitions =
+        findPartitionsToQuery(
```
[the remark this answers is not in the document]
vthacker: good catch! fixed

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -126,6 +136,139 @@ private KaldbServiceGrpc.KaldbServiceFutureStub getKaldbServiceGrpcClient(String
...
+        .stream()
+        .filter(searchMetadata -> snapshotsToSearch.contains(searchMetadata.snapshotName))
+        .collect(Collectors.groupingBy(KaldbDistributedQueryService::getRawSnapshotName))
+        .values()
+        .stream()
+        .map(KaldbDistributedQueryService::pickSearchNodeToQuery)
+        .collect(Collectors.toList());
+  }
+
+  public static boolean isSnapshotInPartition(
```
mansu, line 186: Can we please remove the unused params from this method?

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
```diff
@@ -313,33 +352,52 @@ public void testTwoIndexersAndOneQueryService() throws Exception {
...
+        searchUsingGrpcApi("*:*", indexerPort2, 1633083000000L, 1633083099000L);
     assertThat(indexer2SearchResponse.getTotalNodes()).isEqualTo(1);
     assertThat(indexer2SearchResponse.getFailedNodes()).isEqualTo(0);
     assertThat(indexer2SearchResponse.getTotalCount()).isEqualTo(100);
     assertThat(indexer2SearchResponse.getHitsCount()).isEqualTo(100);
 
     // Query from query service.
     KaldbSearch.SearchResult queryServiceSearchResponse =
-        searchUsingGrpcApi("*:*", queryServicePort);
+        searchUsingGrpcApi("*:*", queryServicePort, 0, 1601547099000L);
```
vthacker, line 393: with `0, 1601547099000L` we only query 1 node

kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
```diff
@@ -348,7 +406,7 @@ public void testTwoIndexersAndOneQueryService() throws Exception {
 
     // Query from query service.
     KaldbSearch.SearchResult queryServiceSearchResponse2 =
-        searchUsingGrpcApi("Message100", queryServicePort);
+        searchUsingGrpcApi("Message100", queryServicePort, 0, Long.MAX_VALUE);
```
vthacker, line 409: with `0, Long.MAX_VALUE` we will query 2 nodes

mansu (dismissed):

bryanlb (approved):

PR #236, opened 2022-02-25
Add initial service metadata

bryanlb:
Adds ability to set the initial service metadata. Basic checks against invalid combinations of partitions are provided.


vthacker:
I actually don't know much about the service currently. So I don't think I can +1 this.

However the code looks okay to me in that it's adding a ZK serializer and then a ServiceMetadata ( the name is confusing to me since it sounds abstract to me but like I said I don't understand the pre-processor work )

The `checkArgument` is the only method that has some logic which looks fine to me - From what I understanding it's just asserting for each service the metadata timestamps don't overlap and if it does the service will fail to startup

kaldb/src/main/proto/metadata.proto
```diff
@@ -109,3 +109,25 @@ message RecoveryTaskMetadata {
   // Created timestamp
   int64 created_time_epoch_ms = 5;
 }
+
```
mansu, line 112: Add some documentation for the messages.

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
```diff
@@ -0,0 +1,74 @@
...
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), partitionList);
+  }
+
+  @Override
+  public String toString() {
+    return "ServiceMetadata{" + "name='" + name + '\'' + ", partitionList=" + partitionList + '}';
+  }
+
+  /** Ensures the provided partition list contains a valid configuration */
```
mansu, line 48: Please describe the check in more detail.

mansu (changes requested):

kaldb/src/main/proto/metadata.proto
```diff
@@ -109,3 +109,25 @@ message RecoveryTaskMetadata {
...
+message ServiceMetadata {
+  // Unique ID for service record
+  string name = 1;
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/proto/metadata.proto
```diff
@@ -109,3 +109,25 @@ message RecoveryTaskMetadata {
...
+
+  // List of configured partitions
+  repeated ServicePartitionMetadata partitions = 2;
+}
+
+message ServicePartitionMetadata {
```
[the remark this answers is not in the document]
bryanlb: Updated model as discussed to start, end, and list of strings

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
bryanlb: Done. This check is also significantly more straightforward with the revised data model.

kaldb/src/main/proto/metadata.proto
bryanlb: Revised descriptions for fields to include additional detail.

kaldb/src/main/proto/metadata.proto
```diff
@@ -109,3 +109,28 @@ message RecoveryTaskMetadata {
   // Created timestamp
   int64 created_time_epoch_ms = 5;
 }
+
+message ServiceMetadata {
```
mansu, line 113: While it's obvious now, in the past it was harder to understand what these structs did at a later time. So, add a comment stating that this struct captures service ownership information, the throughput allocated to the service and the owner. 

Also, define what the partition_configs field would contain.

kaldb/src/main/proto/metadata.proto
```diff
@@ -109,3 +109,28 @@ message RecoveryTaskMetadata {
...
+  string owner = 2;
+
+  // Service throughput max
+  int64 throughput_bytes = 3;
+
+  // List of configured partitions
+  repeated ServicePartitionMetadata partition_configs = 4;
+}
+
+message ServicePartitionMetadata {
```
mansu, line 127: Please add a comment saying that this struct contains the partition allocation for a service from  `start_time_epoch_ms` to `end_time_epoch_ms`.

Since this is an internal struct in ServiceMetadata, does this message need to be a metadata? It's logic can be added to ServiceMetadata alone. No?

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+   * Check that the list of partitionConfigs do not overlap start and end times. This sorts the list
+   * by start of partitions by start time, and then ensures that the end of a given item does not
+   * overlap with the start of the next item in the list.
+   */
+  private void checkPartitions(
+      List<ServicePartitionMetadata> partitionConfig, String errorMessage) {
+    List<ServicePartitionMetadata> sortedConfigsByStartTime =
+        partitionConfig
+            .stream()
+            .sorted(Comparator.comparingLong(ServicePartitionMetadata::getStartTimeEpochMs))
```
mansu, line 80: Since we mostly need access to the most recent partition allocation most of the time, sorting this list in reverse order may be better since we will mostly get the latest allocation by checking one element.

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.slack.kaldb.metadata.core.KaldbMetadata;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Configurations for an upstream service for use in the pre-processor - including rate limits, and
```
mansu, line 13: This comment isn't quite accurate since this data is also used in query service. I think it's better to just add the explanations in protobuf file. 

In this specific file, we can say this is a wrapper class that stores the service specific metadata and the constraints on partition list.

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServicePartitionMetadata.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.common.collect.ImmutableList;
+import com.slack.kaldb.proto.metadata.Metadata;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Metadata for a specific partition configuration at a point in time. For partitions that are
+ * currently active we would expect to have an endTime of max long.
+ */
+public class ServicePartitionMetadata {
```
mansu, line 14: Do we need a separate class for this message? Can we include this in service metadata itself? 

Also, since this is an internal struct, may we can just call this a partition map and not a metadata?

If we need it to be a separate class can it be an internal static class in ServiceMetadata?

kaldb/src/test/java/com/slack/kaldb/metadata/service/ServiceMetadataSerializerTest.java
```diff
@@ -0,0 +1,131 @@
...
+    assertThat(deserializedServiceMetadata.name).isEqualTo(name);
+    assertThat(deserializedServiceMetadata.partitionConfigs).isEqualTo(list);
+  }
+
+  @Test
+  public void testServiceMetadataSerializerEmptyList() throws InvalidProtocolBufferException {
+    final String name = "testService";
+    final String owner = "testOwner";
+    final long throughput = 2000;
+    final List<ServicePartitionMetadata> list = Collections.emptyList();
```
mansu, line 79: Why not throw an exception on an empty partition list?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
bryanlb: Updated comment to reflect this includes the query service as well. I'm not a fan of stating this is a wrapper class for service metadata that contains a partition list, as that just restates what is obvious from the class name and fields.

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServiceMetadata.java
bryanlb: I ended up leaving this as-is. We expect that a given topic will only have a handful of configurations, as this is a manual, user-driven rebalancing operation. When applying a new configuration the existing one will be mutated, and a new configuration will be added to the list. The cost of checking the entire list is low, and will help more quickly identify any logic issues as we develop the remainder of the partition allocation code.

kaldb/src/main/java/com/slack/kaldb/metadata/service/ServicePartitionMetadata.java
bryanlb: [Per previous feedback](https://github.com/slackhq/kaldb/pull/144#discussion_r747761893) it was requested that we avoid nesting where possible, due to limited namespace support. If we intend to change guidance for this we should revisit the other protobuf definition files that also largely conform to this pattern (ie `kaldb_configs`).

I'm not a fan of calling this a partition map, as it's not a map. This makes it unnecessarily confusing when trying to consume the object as you end up with code like `partitionMap.stream()` or `partitionMap.sort()`

kaldb/src/main/proto/metadata.proto
bryanlb: Comments added to top-level objects, and improved description of the `parition_configs`.

kaldb/src/main/proto/metadata.proto
bryanlb: Comment added to top-level message. Discussion about using an internal struct being addressed in https://github.com/slackhq/kaldb/pull/236/files#r817925357

kaldb/src/test/java/com/slack/kaldb/metadata/service/ServiceMetadataSerializerTest.java
bryanlb: I initially did have checks for an empty partition list, but removed them for now. I believe that we will have a few scenarios where we will want the ability define a service without immediately defining a partition map.

mansu (approved):

PR #237, opened 2022-02-25
Convert ArmeriaService to builder pattern

bryanlb:
This allows for easily sharing a common set of configs, while still supporting specialized tweaks depending on the use case.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
@@ -21,7 +21,6 @@ import com.linecorp.armeria.server.grpc.GrpcServiceBuilder;
 import com.linecorp.armeria.server.healthcheck.HealthCheckService;
 import com.linecorp.armeria.server.logging.LoggingService;
 import com.linecorp.armeria.server.logging.LoggingServiceBuilder;
-import com.linecorp.armeria.server.management.ManagementService;
 import com.linecorp.armeria.server.metric.MetricCollectingService;
 import com.slack.kaldb.config.KaldbConfig;
 import com.slack.kaldb.elasticsearchApi.ElasticsearchApiService;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -38,130 +36,106 @@
...
+    public Builder(int port, String serviceName, PrometheusMeterRegistry prometheusMeterRegistry) {
+      this.serviceName = serviceName;
+      this.serverBuilder = Server.builder().http(port);
+
+      // initialize event loop
+      EventLoopGroup eventLoopGroup = EventLoopGroups.newEventLoopGroup(WORKER_EVENT_LOOP_THREADS);
+      EventLoopGroups.warmUp(eventLoopGroup);
+      this.serverBuilder.workerGroup(eventLoopGroup, true);
+
+      // timeout
```
mansu, line 64: Refactoring this code is good. But, I think adding these options in the constructor is an anti-pattern. Instead, the following may be a better option:
* Keep the private methods for each of the options as is. 
* Add an additional method like `addDefaultBuildParams` that adds all of these features automatically. 
* Add an additional static methods in ArmeriaService that like build that will take the params you need and build the service want using the builders.

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -38,130 +36,106 @@
...
-        .requestHeadersSanitizer((ctx, headers) -> DefaultHttpHeaders.EMPTY_HEADERS);
+    public ArmeriaService build() {
+      return new ArmeriaService(serverBuilder.build(), serviceName);
+    }
   }
 
   @Override
   protected void startUp() throws Exception {
-    LOG.info("Starting {} on port {}", serviceName, serverPort);
+    LOG.info("Starting {} on ports {}", serviceName, server.config().ports());
```
mansu, line 138: If there are multiple ports, it's better to list each port number separately here.

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -38,130 +36,106 @@
...
+              .newDecorator());
+
+      // management endpoints
+      this.serverBuilder
+          .service("/health", HealthCheckService.builder().build())
+          .service("/metrics", (ctx, req) -> HttpResponse.of(prometheusMeterRegistry.scrape()))
+          .serviceUnder("/docs", new DocService());
+
+      // tracing
+      String endpoint = KaldbConfig.get().getTracingConfig().getZipkinEndpoint();
```
mansu, line 90: I am slowly converting all code to not use KaldbConfig.get directly. So, pass in the Tracing config explicitly and remove the dependency on `KaldbConfig.get()`. This will also make the unit tests easier to set up. 

Also, should we make the tracing config optional in the builder since it's not needed in unit tests?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: This is actually intentional - the `ports()` method returns a list of `ServerPort` which has a custom defined `toString()`, and includes a log more information (local address, port, protocols).

```
Starting kalDbIndex on ports [ServerPort(0.0.0.0/0.0.0.0:8080, [http])]
```

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: For mandatory parameters, as long as there are only a few of them keeping them in the builder constructor is [recommended](https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java). In the event there were a significant amount of mandatory params, a step builder would be a better pattern to use here.

I do agree that this ends up with our constructor doing too much though. I went ahead and added back the private methods for initialization, and automatically call them as part of the constructor.

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: Ah, that is a good call-out. I went ahead and added another builder option that allows you to pass in the tracing endpoint when building ArmeriaService. If you don't provide it, it will automatically default to the non-reporting tracing like we do today.

For now it's still using the `KaldbConfig.get()` in `Kaldb.java` but directly passes the endpoint in to ArmeriaService - once we convert `Kaldb.java` it'll be an easy swap.

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
vthacker: Totally not a big deal - but should the `initialize*` private methods also be composable and not return void?

vthacker (approved):
LGTM

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
bryanlb: As they're private methods that are called from the constructor, I don't think it significantly matters. Our `ServerBuilder` is final, so if we were to use that approach we'd need to create a temporary instance of `ServerBuilder` and pass it to each of the the initializations methods before finally assigning it. That's certainly an option, but I think it ends up being a bit less readable overall.

mansu (approved):

PR #240, opened 2022-03-01
KaldbDistributedQueryServiceTest potential fix - ensure grpc timeout is lower than global timeout

vthacker:
This PR is an attempt to fix the flaky `KaldbDistributedQueryServiceTest` test

`successfulAsList` is tolerant of failed futures - not pending futures. So `successfulAsList().get(TIMEOUT)` will throw a timeout exception if one of the component futures is still pending. This means we can run into a scenario in the test where 2/3 servers that succeeded also end up failing.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -57,6 +57,8 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
   @VisibleForTesting
   public static long READ_TIMEOUT_MS = DISTRIBUTED_QUERY_TIMEOUT_DURATION.toMillis();
 
+  private static final long GRPC_TIMEOUT_BUFFER_MS = 100;
+
   // For now we will use SearchMetadataStore to populate servers
   // But this is wasteful since we add snapshots more often than we add/remove nodes ( hopefully )
   // So this should be replaced cache/index metadata store when that info is present in ZK
@@ -133,8 +135,10 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
     span.tag("queryServerCount", String.valueOf(stubs.size()));
 
     for (KaldbServiceGrpc.KaldbServiceFutureStub stub : stubs.values()) {
+      // make sure all underlying futures finish executing (successful/cancelled/failed/other)
+      // and cannot be pending when the successfulAsList.get(SAME_TIMEOUT_MS) runs
       ListenableFuture<KaldbSearch.SearchResult> searchRequest =
-          stub.withDeadlineAfter(READ_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+          stub.withDeadlineAfter(READ_TIMEOUT_MS - GRPC_TIMEOUT_BUFFER_MS, TimeUnit.MILLISECONDS)
               .withInterceptors(
                   GrpcTracing.newBuilder(Tracing.current()).build().newClientInterceptor())
               .search(request);
@@ -151,6 +155,8 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
     try {
       List<SearchResult<LogMessage>> searchResults =
           searchFuture.get(READ_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      LOG.debug("searchResults.size={} searchResults={}", searchResults.size(), searchResults);
+
       return searchResults
           .stream()
           .map(searchResult -> searchResult == null ? SearchResult.empty() : searchResult)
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -133,8 +134,10 @@ private void updateStubs() {
     span.tag("queryServerCount", String.valueOf(stubs.size()));
 
     for (KaldbServiceGrpc.KaldbServiceFutureStub stub : stubs.values()) {
+      // make sure all underlying futures finish executing (successful/cancelled/failed/other)
+      // and cannot be pending when the successfulAsList.get(SAME_TIMEOUT_MS) runs
       ListenableFuture<KaldbSearch.SearchResult> searchRequest =
-          stub.withDeadlineAfter(READ_TIMEOUT_MS, TimeUnit.MILLISECONDS)
+          stub.withDeadlineAfter(GRPC_QUERY_TIMEOUT_DURATION.toMillis(), TimeUnit.MILLISECONDS)
```
bryanlb, line 140: This may not work as expected - since this is calculated once as a static value, any test that dynamically updates the `READ_TIMEOUT_MS` will not be reflected here.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -151,6 +154,21 @@ private void updateStubs() {
     try {
       List<SearchResult<LogMessage>> searchResults =
           searchFuture.get(READ_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      // we want to be super sure the inner LOG.debug statements are not getting evaluated and then
+      // thrown away because debug is enabled in production as this is a hot code path
+      if (LOG.isDebugEnabled()) {
```
bryanlb, line 159: I would just do this as a normal `LOG.debug` instead of adding a special check - additionally, when constructing the value we should be using `{}` instead of constructing the string with concatenation, as it defers the computation.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -151,6 +154,21 @@ private void updateStubs() {
...
+        LOG.debug("searchResults.size() = " + searchResults.size());
+        for (SearchResult<LogMessage> searchResult : searchResults) {
+          LOG.debug(
```
bryanlb, line 162: Instead of doing a loop can't we just print the list directly? The toString methods have reasonable defaults.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -175,6 +193,12 @@ private void updateStubs() {
           ((SearchResultAggregator<LogMessage>)
                   new SearchResultAggregatorImpl<>(SearchResultUtils.fromSearchRequest(request)))
               .aggregate(searchResults);
+
+      LOG.debug(
```
bryanlb, line 197: As above, this should be using the `{}` replacement instead of string concat.

bryanlb (changes requested):
Couple things that need addressed.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: That is a solid point...

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
vthacker: I'm still thinking of keeping this because we will save on iterating the loop of hundreds of servers potentially

vthacker: I changed the comment to reflect the new reason

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -132,9 +132,13 @@ private void updateStubs() {
     List<ListenableFuture<SearchResult<LogMessage>>> queryServers = new ArrayList<>(stubs.size());
     span.tag("queryServerCount", String.valueOf(stubs.size()));
 
+    long GRPC_QUERY_TIMEOUT_DURATION = READ_TIMEOUT_MS - 100;
```
vthacker, line 135: Hmm! i am not sure how we want this to be? Since you rightly brought up the fact that the test can override the static, we want `GRPC_QUERY_TIMEOUT_DURATION` to be evaluated  on every request? Is there any benefiting than just hardcoding it with `READ_TIMEOUT_MS - 100` in `withDeadlineAfter`?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResult.java
```diff
@@ -90,4 +90,26 @@ public int hashCode() {
   public static SearchResult<LogMessage> empty() {
     return EMPTY;
   }
+
+  @Override
+  public String toString() {
```
vthacker, line 95: I let IntelliJ generate this.

bryanlb (approved):

PR #242, opened 2022-03-01
Remove direct access to static config in code

mansu:
Remove direct access to `KaldbConfig.get()` in the code since it introduces tight coupling between classes and makes it harder to test code. Instead, pass the protobuf object around instead.


vthacker:
I think we need to fix `Kaldb#start` method - this still uses a static reference `KaldbConfig.get()` - this is important to remove since we want to be able to finally reuse `Kaldb.java` in test cases rather than create bespoke servers for testing that have different threading models than production.

vthacker:
What's to stop us from mistakenly calling `KaldbConfig#get` and accessing the value statically ? Shouldn't `_instance` not be a static anymore?

vthacker:
Looks like CLA assistant is unhappy as well

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
```diff
@@ -95,15 +94,17 @@ protected void shutDown() throws Exception {
   public static CachingChunkManager<LogMessage> fromConfig(
       MeterRegistry meterRegistry,
       MetadataStore metadataStore,
-      KaldbConfigs.ServerConfig serverConfig)
+      KaldbConfigs.ServerConfig serverConfig,
+      KaldbConfigs.S3Config s3Config,
+      KaldbConfigs.CacheConfig cacheConfig)
```
bryanlb, line 99: Why are we passing both `serverConfig` and `cacheConfig`, as `serverConfig` is a property of `cacheConfig`?

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
```diff
@@ -50,8 +50,9 @@
 
   private final IndexingChunkManager<LogMessage> chunkManager;
 
-  public static LogMessageTransformer getLogMessageTransformer() {
-    String dataTransformerConfig = KaldbConfig.get().getIndexerConfig().getDataTransformer();
+  public static LogMessageTransformer getLogMessageTransformer(
+      KaldbConfigs.IndexerConfig indexerConfig) {
+    String dataTransformerConfig = indexerConfig.getDataTransformer();
```
bryanlb, line 55: Why are we passing the `indexerConfig` if we only use the `dataTransformer` string? This should only require the string config to be passed.

kaldb/src/test/java/com/slack/kaldb/testlib/KaldbConfigUtil.java
```diff
@@ -88,4 +70,26 @@ public static void initEmptyIndexerConfig() throws InvalidProtocolBufferExceptio
...
         .build();
   }
+
+  public static int TEST_INDEXER_PORT = 10000;
+
+  public static KaldbConfigs.IndexerConfig makeIndexerConfig() {
+    return makeIndexerConfig(TEST_INDEXER_PORT);
+  }
+
+  public static KaldbConfigs.IndexerConfig makeIndexerConfig(int indexerPort) {
```
bryanlb, line 80: It would be better if we called these `get...Config`, as that more clearly indicates they do not mutate state. `Make` is significantly more ambiguous, and requires reviewing the source to determine if it does indeed deal with state or is static.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
```diff
@@ -95,15 +94,17 @@ protected void shutDown() throws Exception {
...
     return new CachingChunkManager<>(
         meterRegistry,
         metadataStore,
-        getS3BlobFsClient(KaldbConfig.get().getS3Config()),
+        getS3BlobFsClient(s3Config),
         SearchContext.fromConfig(serverConfig),
-        KaldbConfig.get().getS3Config().getS3Bucket(),
-        KaldbConfig.get().getCacheConfig().getDataDirectory(),
-        KaldbConfig.get().getCacheConfig().getSlotsPerInstance());
+        s3Config.getS3Bucket(),
```
bryanlb, line 106: If we only need the `s3Config`, the `dataDirectory`, and `slotsPerInstance` can we just pass those instead?

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -59,12 +60,13 @@
   private final Counter commitsCounter;
   private final Counter refreshesCounter;
 
-  public static LuceneIndexStoreImpl makeLogStore(File dataDirectory, MeterRegistry metricsRegistry)
+  public static LuceneIndexStoreImpl makeLogStore(
+      File dataDirectory, KaldbConfigs.IndexerConfig indexerConfig, MeterRegistry metricsRegistry)
```
bryanlb, line 64: This should just take the commit duration and refresh duration if those are the only arguments we need.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
mansu: Good point. Removed serverConfig param.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
mansu: Yeah, I considered it and then decided to pass most general object, so it's more future proof when we add more params to this object.

mansu:
> I think there are 4 references to `KaldbConfig.initFromConfigObject` in the tests that we need to remove - Those calls were made to workaround the problem that we had one static config. I'd be curious to know if the tests still pass once this is removed.

Removed the calls to `initFromConfigObject` from everywhere and made it a private method.

>I think we need to fix Kaldb#start method - this still uses a static reference KaldbConfig.get() - this is important to remove since we want to be able to finally reuse Kaldb.java in test cases rather than create bespoke servers for testing that have different threading models than production.

Good point. Removed. 


>What's to stop us from mistakenly calling KaldbConfig#get and accessing the value statically ? Shouldn't _instance not be a static anymore?
Moved KaldbConfig into server config and made this method package private. So, no one outside the package should be able to call KaldbConfig.get from the rest of the code.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
mansu: Passing in most general type here in case we add more params in future. Ideally, we should refactor these params into it's own Lucene config.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
mansu: Passing in most general type for now since there may be more params in future,

kaldb/src/test/java/com/slack/kaldb/testlib/KaldbConfigUtil.java
mansu: In general, get is used for a field access and using make for static methods. Also, the name is more consistent with existing convention, so left it as is.

bryanlb: This is somewhat of an anti-pattern, as `get` is expected to be used for methods that return a "non-boolean function or attribute of the object on which they're invoked." Using `make` is fine, but `get` would be a more appropriate use.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
bryanlb: Let's pass the most specific object possible for all of these methods. We do not expect to be changing the behavior of these classes significantly as they are functionally complete at this time. 

It is not a significant amount of work to pass in an additional argument if needed in the future, and it will help when writing and troubleshooting if only the required arguments are passed.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
bryanlb: As above, let's pass in the most specific argument. This case especially since this code will be moving out of the indexer entirely.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -1,6 +1,6 @@
 package com.slack.kaldb.chunkManager;
 
-import static com.slack.kaldb.config.KaldbConfig.LOCAL_QUERY_TIMEOUT_DURATION;
+import static com.slack.kaldb.server.KaldbConfig.LOCAL_QUERY_TIMEOUT_DURATION;
```
vthacker, line 3: What's the reason here to change the package of `KaldbConfigs` class?  It seems better suited for the `config` package right?

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -75,7 +75,8 @@ public static void main(String[] args) throws Exception {
     }
     Path configFilePath = Path.of(args[0]);
 
-    Kaldb kalDb = new Kaldb(configFilePath);
+    KaldbConfig.initFromFile(configFilePath);
+    Kaldb kalDb = new Kaldb(KaldbConfig.get());
     kalDb.start();
```
vthacker, line 80: I think we want `start` to take `KaldbConfig` as an argument so that we can potentially reuse the `start` method in the test cases

mansu: Sure, we can make that change when we change the tests. For now the tests work well.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
bryanlb: This one we agreed to leave for now, and revisit later with potentially a better scoped local config argument.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
mansu: done.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
mansu: Done.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
vthacker: but we literally are using it outside the package :) I don't get the reasoning . Could you please explain more about the change motivation

mansu: These are static fields. The other methods are package private.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
bryanlb: It looks like this wasn't entirely updated, we didn't update the method to take the newly created `LuceneConfig`?

bryanlb (approved):
🐑

PR #244, opened 2022-03-22
Add initial GRPC manager API

bryanlb:
Allows for simple service manipulation, and is fully described in the DocService.

Screen Shot 2022-03-24 at 12 31 01 PM
Screen Shot 2022-03-24 at 12 30 52 PM

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
@@ -1,7 +1,6 @@
 package com.slack.kaldb.server;
 
 import com.google.common.collect.ImmutableList;
-import com.linecorp.armeria.server.ServiceRequestContext;
 import com.slack.kaldb.metadata.service.ServiceMetadata;
 import com.slack.kaldb.metadata.service.ServiceMetadataStore;
 import com.slack.kaldb.metadata.service.ServicePartitionMetadata;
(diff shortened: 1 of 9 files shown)


kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
```diff
@@ -0,0 +1,206 @@
...
+              .build());
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      LOG.error("Error getting services", e);
+      responseObserver.onError(Status.UNKNOWN.withDescription(e.getMessage()).asException());
+    }
+  }
+
+  @Override
+  public void updatePartitionAssignment(
```
mansu, line 116: Please add a high level description of what this method does for various vales of the inputs.

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
```diff
@@ -0,0 +1,206 @@
...
+      ServicePartitionMetadata updatedPreviousActivePartition =
+          new ServicePartitionMetadata(
+              previousActiveServicePartition.get().getStartTimeEpochMs(),
+              partitionCutoverTime,
+              previousActiveServicePartition.get().getPartitions());
+      builder.add(updatedPreviousActivePartition);
+    }
+
+    ServicePartitionMetadata newPartitionMetadata =
+        new ServicePartitionMetadata(partitionCutoverTime + 1, Long.MAX_VALUE, newPartitionIdsList);
```
mansu, line 203: I think it's best to pass in the padding as a parameter to this method. That way the implementation would be oblivious of the padding value. The caller can decide what value to set it as.

Also, please use the static variable instead of directly calling Long.MAX_VALUE.

kaldb/src/test/java/com/slack/kaldb/server/ManagerApiGrpcTest.java
```diff
@@ -0,0 +1,409 @@
...
+            .setOwner(serviceOwner)
+            .build());
+
+    Metadata.ServiceMetadata getServiceMetadataResponse =
+        managerApiStub.getServiceMetadata(
+            ManagerApi.GetServiceMetadataRequest.newBuilder().setName(serviceName).build());
+    assertThat(getServiceMetadataResponse.getName()).isEqualTo(serviceName);
+    assertThat(getServiceMetadataResponse.getOwner()).isEqualTo(serviceOwner);
+    assertThat(getServiceMetadataResponse.getThroughputBytes()).isEqualTo(0);
+    assertThat(getServiceMetadataResponse.getPartitionConfigsList().size()).isEqualTo(0);
```
mansu, line 104: Please add an assert reading this data from the servicemetadata store also. All the CRUD tests can benefit from these additional checks.

kaldb/src/test/java/com/slack/kaldb/server/ManagerApiGrpcTest.java
```diff
@@ -0,0 +1,409 @@
...
+        (StatusRuntimeException)
+            catchThrowable(
+                () ->
+                    managerApiStub.updatePartitionAssignment(
+                        ManagerApi.UpdatePartitionAssignmentRequest.newBuilder()
+                            .setName(serviceName)
+                            .setThroughputBytes(-1)
+                            .addAllPartitionIds(partitionList)
+                            .build()));
+    assertThat(throwable1.getStatus().getCode()).isEqualTo(Status.UNKNOWN.getCode());
```
mansu, line 323: assertThatExceptionIsThrownBy() may be a simpler way to assert this condition.

bryanlb: I initially had that approach but wanted to also assert specific GRPC codes and messages. In order to this, I needed access to the throwable so I could access the getters for the methods required.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -0,0 +1,65 @@
...
+  repeated ServiceMetadata service_metadata = 1;
+}
+
+// UpdatePartitionAssignmentRequest represents a request to allocate a new set of partitions
+message UpdatePartitionAssignmentRequest {
+  // Must be an exact string name match
+  string name = 1;
+  // Configured maximum throughput of the service
+  int64 throughput_bytes = 2;
+  // Optional desired partition IDs - if not provided will be automatically assigned
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -0,0 +1,65 @@
...
+option java_package = "com.slack.kaldb.proto.manager_api";
+
+// ManagerApiService is the administration API for operating on service metadata and the associated partition assignments
+service ManagerApiService {
+  // CreateService initializes a new service in the metadata store with no initial allocated capacity
+  rpc CreateServiceMetadata(CreateServiceMetadataRequest) returns (ServiceMetadata) {}
+  // GetService returns a single service metadata by name
+  rpc GetServiceMetadata(GetServiceMetadataRequest) returns (ServiceMetadata) {}
+  // UpdateService updates an existing service with new metadata
+  rpc UpdateServiceMetadata(UpdateServiceMetadataRequest) returns (ServiceMetadata) {}
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
bryanlb: Done.

kaldb/src/test/java/com/slack/kaldb/server/ManagerApiGrpcTest.java
bryanlb: Done. Added service metadata store checks to every method in this test suite.

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
bryanlb: Added static variable for the long max value. Updated comment for when we do support padding that it should be exposed as param.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -0,0 +1,65 @@
...
+  // GetServices returns all available services from the metadata store
+  rpc ListServiceMetadata(ListServiceMetadataRequest) returns (ListServiceMetadataResponse) {}
+  // Update partition
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -0,0 +1,65 @@
...
+// ListServiceMetadataRequest represents a request to list all existing services
+message ListServiceMetadataRequest {}
+
+// ListServiceMetadataResponse represents the response when listing all available services
+message ListServiceMetadataResponse {
+  // List of services, excluding partition assignments
+  repeated ServiceMetadata service_metadata = 1;
+}
+
+// UpdatePartitionAssignmentRequest represents a request to allocate a new set of partitions
```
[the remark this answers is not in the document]
bryanlb: Updated the field descriptions to indicate both throughput and partition IDs are optional, and how to mark them as such in a request. Since this is the doc for the request object and not the service, there's no allocation behavior.

mansu (approved):

PR #253, opened 2022-04-12
Reduce logging level for rate limiter

bryanlb:
This log doesn't generate a significant value, but is generating an excessive amount of logging - so we should reduce the default log level.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
+++ b/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
@@ -72,9 +72,9 @@ public class PreprocessorRateLimiter {
       int bytes = value.toByteArray().length;
       if (serviceName == null || serviceName.isEmpty()) {
         // service name wasn't provided
-        LOG.warn("Message was dropped due to missing service name - '{}'", value);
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
```diff
@@ -72,10 +72,7 @@ public Predicate<String, Trace.Span> createRateLimiter(
       int bytes = value.toByteArray().length;
       if (serviceName == null || serviceName.isEmpty()) {
         // service name wasn't provided
-        LOG.warn("Message was dropped due to missing service name - '{}'", value);
-        // todo - we may consider adding a logging BurstFilter so that a bad actor cannot
```
mansu, line 76: Please leave the TODO in since it's still valuable.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
bryanlb: Done.

mansu (approved):

PR #255, opened 2022-04-16
Add additional logging to Kaldb indexer to ease debugging.

mansu:


kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -329,6 +329,7 @@ void shutdown() {
   }
 
   private void addShutdownHook() {
+    LOG.info("Running shutdown hook.");
```
bryanlb, line 332: This isn't quite accurate - this is adding the shutdown hook, but doesn't invoke it. This will likely be confusing when we're looking at the logs the way it's currently worded.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
mansu: Good catch. Fixed.

bryanlb (approved):

PR #258, opened 2022-04-20
Change Armeria shutdown behavior

bryanlb:
Switch Armeria to take up to the maximum allowable shutdown time.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
@@ -150,7 +150,15 @@ public class ArmeriaService extends AbstractIdleService {
   @Override
   protected void shutDown() throws Exception {
     LOG.info("Shutting down");
-    server.closeAsync().get(15, TimeUnit.SECONDS);
+
+    // On server close there is an option for a graceful shutdown, which is disabled by default.
+    // When it is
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -150,7 +150,15 @@ protected void startUp() throws Exception {
...
-    server.closeAsync().get(15, TimeUnit.SECONDS);
+
+    // On server close there is an option for a graceful shutdown, which is disabled by default.
+    // When it is
+    // disabled it immediately starts rejecting requests and begins the shutdown process, which
+    // includes
+    // running any remaining AsyncClosableSupport closeAsync actions. We want to allow this to take
+    // up to the
+    // maximum permissible shutdown time to successfully close.
+    server.close();
```
vthacker, line 161: does server.close throw exceptions?  If yes what happens if an exception is thrown?

bryanlb: It can, yes. In that scenario that would result in a failed service, and it would get logged like we see today with the `.get` times out.

vthacker: cool! Just wanted to double confirm if anything here can cause a restart loop

vthacker (approved):

PR #259, opened 2022-04-20
Change indexer shutdown process

bryanlb:
Rework the indexer shutdown to prevent timing issue between chunk manager and indexer logic.


kaldb/src/main/java/com/slack/kaldb/server/KaldbConfig.java
```diff
@@ -31,7 +31,6 @@ public class KaldbConfig {
   public static final int DEFAULT_ZK_TIMEOUT_SECS = 15;
 
   public static final String CHUNK_DATA_PREFIX = "log";
-  public static final long DEFAULT_ROLLOVER_FUTURE_TIMEOUT_MS = 30000;
```
mansu, line 34: Are there any side effects of changing the default timeout value to 15 secs from 30 secs?

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
```diff
@@ -79,7 +78,8 @@ public KaldbIndexer(
   protected void startUp() throws Exception {
     LOG.info("Starting Kaldb indexer.");
     long startOffset = indexerPreStart();
-    chunkManager.awaitRunning(DEFAULT_START_STOP_DURATION);
+    // ensure the chunk manager is available to receive messages
+    chunkManager.awaitRunning();
```
mansu, line 82: Just confirming that guava service manager timeout would be applied now in case the service fails to start.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
```diff
@@ -151,17 +151,13 @@ protected void run() throws Exception {
...
-    // Shutdown kafka consumer cleanly and then the chunkmanager so we can be sure, we have indexed
-    // the data we ingested.
+    // Shutdown kafka consumer
     try {
       kafkaConsumer.close();
     } catch (Exception e) {
       LOG.warn("Failed to close kafka consumer cleanly because of an exception.", e);
     }
 
-    chunkManager.stopAsync();
```
mansu, line 162: Please add a comment explaining why we don't wait for chunk manager to be shutdown so we don't add it back later.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
bryanlb: This would be expected to wait indefinitely on the `chunkManager` before the indexer will start indexing (depending on healthcheck configs). Having a timeout here doesn't make a ton of sense, because if the chunk manager isn't ready there's no sense in proceeding reading messages from Kafka.

kaldb/src/main/java/com/slack/kaldb/server/KaldbConfig.java
bryanlb: There is not - the only time we used this value was during shutdown, and this timeout was wrapped in the shutdown that forced a shorter timeout of 15 seconds. This should just prevent all the errors we see when nesting a 30s timeout inside a 15s timeout.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
mansu: Since a timeout would terminate the process isn't that preferable than a hung process?

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
mansu: Missing comment?

mansu (changes requested):

mansu (approved):

PR #268, opened 2022-05-02
Close metadatastores in CachingChunkManager.

mansu:
Close metadata stores in `CachingChunkManager` so we are consistent with the rest of the code base.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
@@ -29,10 +29,10 @@ public class CachingChunkManager<T> extends ChunkManager<T> {
   private final String s3Bucket;
   private final String dataDirectoryPrefix;
   private final int slotCountPerInstance;
-  private final ReplicaMetadataStore replicaMetadataStore;
-  private final SnapshotMetadataStore snapshotMetadataStore;
-  private final SearchMetadataStore searchMetadataStore;
-  private final CacheSlotMetadataStore cacheSlotMetadataStore;
+  private ReplicaMetadataStore replicaMetadataStore;
+  private SnapshotMetadataStore snapshotMetadataStore;
+  private SearchMetadataStore searchMetadataStore;
+  private CacheSlotMetadataStore cacheSlotMetadataStore;
 
   public CachingChunkManager(
       MeterRegistry registry,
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
```diff
@@ -37,26 +41,26 @@ public CachingChunkManager(
...
+    snapshotMetadataStore = new SnapshotMetadataStore(metadataStore, false);
+    searchMetadataStore = new SearchMetadataStore(metadataStore, false);
+    cacheSlotMetadataStore = new CacheSlotMetadataStore(metadataStore, false);
   }
 
   @Override
   protected void startUp() throws Exception {
     LOG.info("Starting caching chunk manager");
 
-    ReplicaMetadataStore replicaMetadataStore = new ReplicaMetadataStore(metadataStore, false);
```
bryanlb, line 54: We should keep these initializations in startup, or move all of the codebase to initializing these in the constructor. All other classes initialize temporary metadata stores in the `startup` (ie RecoveryService, IndexingChunkManager, etc).

The reason these are in the startup is an artifact of when the metadatastore could have been "un-started" when these classes were initialized. This is no-longer the case though, as we've moved the metadata startup prior to service initialization .

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
```diff
@@ -86,6 +90,10 @@ protected void shutDown() throws Exception {
           }
         });
 
+    replicaMetadataStore.close();
```
bryanlb, line 93: These should be closed in reverse order to initialization.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
mansu: Leaving them in startup for now.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
mansu: Done.

bryanlb (approved):

PR #271, opened 2022-05-05
Swap Lucene NIOFS to MMap

bryanlb:
Swap directory implementation for memory-mapped version.

https://stackoverflow.com/questions/12343054/high-cpu-usage-on-search-over-lucene-index


vthacker (approved):

mansu (approved):

dsmiley:
Why was NIO used in the first place?  It is not the default.

vthacker:
I think when Suman first committed the codebase, this was what was added. I don't think it was done intentionally for a specific performance reason right @mansu ?

PR #277, opened 2022-05-09
Improve histogram performance

bryanlb:
Swaps the internal implementation in the histogram builder to use a faster implementation for determining the appropriate bucket. Instead of a binary search we can directly determine which bucket a value belongs in with arithmetic, reducing the complexity from `O(log n)` to `O(1)`

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
@@ -12,6 +12,8 @@ public class FixedIntervalHistogramImpl implements Histogram {
   private final double low;
   private final double high;
   private final int bucketCount;
+
+  private final double bucketSize;
   /**
    * Once the histogram is initialized, the buckets can't be changed since we rely on insertion
    * order of buckets for the findBucket function.
@@ -29,6 +31,7 @@ public class FixedIntervalHistogramImpl implements Histogram {
     this.high = high;
     this.bucketCount = bucketCount;
     this.count = 0;
+    this.bucketSize = (high - low) / bucketCount;
     this.buckets = makeHistogram(low, high, bucketCount);
   }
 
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
```diff
@@ -95,24 +98,6 @@ private Optional<HistogramBucket> findMatchingBucket(HistogramBucket matchingBuc
     return Optional.empty();
   }
 
-  /**
-   * Runs a binary search to find the correct bucket for the value being inserted and return it's
```
mansu, line 99: Can we add a unit test comparing the output from this method and the proposed method. That would be a cheap way to make sure that the outputs are identical. 

Also, can we ensure that this method would work in the aggregation path also?

kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
```diff
@@ -56,8 +59,8 @@ public void add(long value) {
     if (value == high) {
       buckets.get(bucketCount - 1).increment(1);
     } else {
-      int bucketIdx = findMatchingBucket(buckets, 0, bucketCount, (double) value);
-      buckets.get(bucketIdx).increment(1);
+      int index = (int) Math.floor((value - low) / bucketSize);
```
vthacker, line 62: What happens when value is more than `high`?

bryanlb: Good catch! Fixed and added an extra test.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
vthacker: FYI I don't think this value will ever be higher than `high` in how we use it :) the collector I think will filter out matching docs in our time range before this gets invoked..

vthacker:
Nice! LGTM

kaldb/src/main/java/com/slack/kaldb/histogram/FixedIntervalHistogramImpl.java
bryanlb: As discussed via Slack, we verified the new version is generating identical outputs with an uncommitted test prior to removing this old version.

mansu (approved):

PR #278, opened 2022-05-09
Add Jackson Afterburner

bryanlb:
Also switches most of our custom ObjectMappers to the common JsonUtil singleton instance.


mansu:
Please create an issue for incorporating [blackbird](https://github.com/FasterXML/jackson-modules-base/tree/master/blackbird) once we move to Java 11.

PR #280, opened 2022-05-10
Granular recovery tasks

mansu:
Currently, we can create only one chunk per recovery task. However, during recovery task creation the recovery task range can be arbitrarily long which can cause very large chunks and all the recovery chunks will be of different non-uniform sizes.
To work around this, break down large recovery task into smaller uniformly sized tasks.

Currently, `RecoveryTaskMetadata` doesn't allow tasks of size 1. So, updated the code to allow for chunks of size 1. 

Added new unit tests for the new code.

NOTE: We use the `getMaxMessagesPerChunk` setting from indexer config for now. I can add a new config setting if needed.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/recovery/RecoveryTaskMetadata.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/recovery/RecoveryTaskMetadata.java
@@ -23,7 +23,8 @@ public class RecoveryTaskMetadata extends KaldbMetadata {
     checkArgument(
         partitionId != null && !partitionId.isEmpty(), "partitionId can't be null or empty");
     checkArgument(startOffset >= 0, "startOffset must greater than 0");
-    checkArgument(endOffset > startOffset, "endOffset must be greater than the startOffset");
+    checkArgument(
+        endOffset >= startOffset, "endOffset must be greater than or equal to the startOffset");
     checkArgument(createdTimeEpochMs > 0, "createdTimeEpochMs must be greater than 0");
 
     this.partitionId = partitionId;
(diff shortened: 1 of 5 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/recovery/RecoveryTaskMetadata.java
```diff
@@ -23,7 +23,7 @@ public RecoveryTaskMetadata(
     checkArgument(
         partitionId != null && !partitionId.isEmpty(), "partitionId can't be null or empty");
     checkArgument(startOffset >= 0, "startOffset must greater than 0");
-    checkArgument(endOffset > startOffset, "endOffset must be greater than the startOffset");
+    checkArgument(endOffset >= startOffset, "endOffset must be greater than the startOffset");
```
bryanlb, line 26: Probably should update the message here as well.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
```diff
@@ -226,6 +221,43 @@ public long determineStartingOffset(long currentHeadOffsetForPartition) {
...
+    final long creationTimeEpochMs = Instant.now().toEpochMilli();
+    final String recoveryTaskName = getRecoveryTaskName(creationTimeEpochMs, partitionId);
+    recoveryTaskMetadataStore.createSync(
+        new RecoveryTaskMetadata(
+            recoveryTaskName, partitionId, startOffset, endOffset, creationTimeEpochMs));
+    LOG.info(
+        "Created recovery task {} to catchup. Moving the starting offset to head at {}",
+        recoveryTaskName,
+        endOffset);
+    recoveryTasksCreated.increment();
```
bryanlb, line 258: It would be beneficial to also include a tag for the partition ID on the counter as well.

bryanlb (changes requested):
Couple minor things, otherwise looks pretty good.

bryanlb (dismissed):

bryanlb (approved):

bryanlb (approved):

PR #281, opened 2022-05-10
Better metrics for lucene commit/refresh

bryanlb:
Converts the commit and refresh instrumentation to use timers instead of counters, so we have better visibility into how long these calls are taking.


kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -54,8 +54,8 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
   // Stats counters.
   private final Counter messagesReceivedCounter;
   private final Counter messagesFailedCounter;
-  private final Counter commitsCounter;
-  private final Counter refreshesCounter;
+  private final io.micrometer.core.instrument.Timer commitsTimer;
```
mansu, line 57: Please import the dependency.

mansu (changes requested):

mansu (approved):

PR #289, opened 2022-05-20
Add io_uring transport

bryanlb:
This can be enabled with Armeria flags if desired, but we need to have the class loaded.


vthacker:
IO Uring is so cool! I'd seen some performance benchmarks and it looked amaze

vthacker (approved):

PR #291, opened 2022-05-23
Search filter nodes - Take 2

vthacker:
Initial PR https://github.com/slackhq/kaldb/pull/235

This PR made queries slower - So this PR has 1 additional commit to add some trace data where we suspect there can be slowdowns.


mansu:
Closing pull request since this is now merged as 2 diffs.

PR #295, opened 2022-05-24
Use dedicated threadpool for logindexsearcher

bryanlb:
This introduces a dedicated threadpool with a named executor for better visibility. By default parallelStream uses the forkjoin pool, which is sized to available processors - 1. Since we queue so many tasks this introduces possible deadlocks with other tasks attempting to use the forkjoin pool.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
@@ -8,6 +8,7 @@ import brave.ScopedSpan;
 import brave.Tracing;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.histogram.FixedIntervalHistogramImpl;
 import com.slack.kaldb.histogram.Histogram;
 import com.slack.kaldb.histogram.NoOpHistogramImpl;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -51,9 +55,17 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(LogIndexSearcherImpl.class);
 
+  // number of threads to use transforming the results
+  private static final int PARALLEL_THREADS = 12;
+
   private final SearcherManager searcherManager;
   private final StandardAnalyzer analyzer;
 
+  private final ExecutorService executorService =
+      Executors.newFixedThreadPool(
```
mansu, line 65: Why not use a cached thread pool?

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: We could, but there isn't expected to be a meaningful difference. We use this threadpool for every query, so in order for a cached threadpool to reduce the threads it would need to have no queries that had more than 12 results for 60 seconds.

bryanlb:
Attempting a revert of the parallel functionality first - https://github.com/slackhq/kaldb/pull/296

PR #297, opened 2022-05-24
Remove streams in hot code paths.

mansu:
Remove streams usage in hot code paths so the code is more efficient. For example, in this case, there is less garbage since we are able to fix the size of result set and also no wrapper objects are created to iterate on the stream.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -177,6 +177,7 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
     ScopedSpan pickSearchNodeToQuerySpan =
         Tracing.currentTracer()
             .startScopedSpan("KaldbDistributedQueryService.pickSearchNodeToQuery");
+    // TODO: Re-write this code using a for loop.
     var nodes =
         searchMetadataStore
             .getCached()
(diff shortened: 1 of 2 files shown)


vthacker (dismissed):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -158,20 +158,17 @@ public static Collection<String> getSearchNodesToQuery(
...
-                snapshotMetadata ->
-                    containsDataInTimeRange(
-                            snapshotMetadata.startTimeEpochMs,
-                            snapshotMetadata.endTimeEpochMs,
-                            queryStartTimeEpochMs,
-                            queryEndTimeEpochMs)
-                        && isSnapshotInPartition(snapshotMetadata, partitions))
-            .map(snapshotMetadata -> snapshotMetadata.name)
-            .collect(Collectors.toSet());
+    Set<String> snapshotsToSearch = new HashSet<>(256);
```
bryanlb, line 161: We should probably leave this default until we have a better idea of the performance implications. With the default load factor this still isn't high enough to prevent a resize for recent queries.

If we want to keep a magic number here, it should be documented on how we arrived at the value so people adjusting it in the future understand the motivation.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
mansu: Switched to default value. I think tweaking this number to be a small value, would help remove some allocations.

PR #309, opened 2022-06-01
simple fetch result reducer.

mansu:
A simple heuristic to reduce the number of results fetched for a topK query.


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -277,6 +277,13 @@ protected static List<ServicePartitionMetadata> findPartitionsToQuery(
         .collect(Collectors.toList());
   }
 
+  private static int determineDownStreamFetchSize(int shardCount, int howMany) {
+    if (shardCount > 20) {
```
bryanlb, line 281: Maybe add a comment explaining how we got to these numbers?

PR #311, opened 2022-06-02
Lucene query improvements

bryanlb:
* Swaps Collectors to use CollectorManagers, so that we can take advantage of a custom executor (see https://github.com/apache/lucene/pull/240, https://issues.apache.org/jira/browse/LUCENE-10002)
* Adjusts logic so collectors only run when data is requested instead of on every query 
* Simplifies the stats collector to extend the `SimpleCollector`, matching the built-in [collectors](https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/TotalHitCountCollector.html), https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/SimpleCollector.html
* Removes hard-coded ramBuffer config and compound file overrides in favor of default values. This now matches the default behavior of OpenSearch [merge policy](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/MergePolicyConfig.java).

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
@@ -26,6 +26,9 @@ public class LuceneIndexStoreConfig {
   // The name of the lucene log file.
   public final String logFileName;
 
+  // Config to set the IndexWriterConfig.setRAMBufferSizeMB.
+  public final int ramBufferSizeMB;
+
   // A flag that turns on internal logging.
   public final boolean enableTracing;
 
@@ -44,8 +47,18 @@ public class LuceneIndexStoreConfig {
   }
 
   public LuceneIndexStoreConfig(
-      Duration commitDuration, Duration refreshDuration, String indexRoot, boolean enableTracing) {
-    this(commitDuration, refreshDuration, indexRoot, DEFAULT_LOG_FILE_NAME, enableTracing);
+      Duration commitDuration,
+      Duration refreshDuration,
+      String indexRoot,
+      int ramBufferSizeMB,
+      boolean enableTracing) {
+    this(
+        commitDuration,
+        refreshDuration,
+        indexRoot,
+        DEFAULT_LOG_FILE_NAME,
+        ramBufferSizeMB,
+        enableTracing);
   }
 
   public LuceneIndexStoreConfig(
@@ -53,6 +66,7 @@ public class LuceneIndexStoreConfig {
       Duration refreshDuration,
       String indexRoot,
       String logFileName,
+      int ramBufferSizeMB,
       boolean enableTracing) {
     ensureTrue(
         !(commitDuration.isZero() || commitDuration.isNegative()),
@@ -64,6 +78,7 @@ public class LuceneIndexStoreConfig {
     this.refreshDuration = refreshDuration;
     this.indexRoot = indexRoot;
     this.logFileName = logFileName;
+    this.ramBufferSizeMB = ramBufferSizeMB;
     this.enableTracing = enableTracing;
   }
 
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
@@ -26,9 +26,6 @@ public class LuceneIndexStoreConfig {
   // The name of the lucene log file.
   public final String logFileName;
 
-  // Config to set the IndexWriterConfig.setRAMBufferSizeMB.
-  public final int ramBufferSizeMB;
-
   // A flag that turns on internal logging.
   public final boolean enableTracing;
 
@@ -47,18 +44,8 @@ public class LuceneIndexStoreConfig {
   }
 
   public LuceneIndexStoreConfig(
-      Duration commitDuration,
-      Duration refreshDuration,
-      String indexRoot,
-      int ramBufferSizeMB,
-      boolean enableTracing) {
-    this(
-        commitDuration,
-        refreshDuration,
-        indexRoot,
-        DEFAULT_LOG_FILE_NAME,
-        ramBufferSizeMB,
-        enableTracing);
+      Duration commitDuration, Duration refreshDuration, String indexRoot, boolean enableTracing) {
+    this(commitDuration, refreshDuration, indexRoot, DEFAULT_LOG_FILE_NAME, enableTracing);
   }
 
   public LuceneIndexStoreConfig(
@@ -66,7 +53,6 @@ public class LuceneIndexStoreConfig {
       Duration refreshDuration,
       String indexRoot,
       String logFileName,
-      int ramBufferSizeMB,
       boolean enableTracing) {
     ensureTrue(
         !(commitDuration.isZero() || commitDuration.isNegative()),
@@ -78,7 +64,6 @@ public class LuceneIndexStoreConfig {
     this.refreshDuration = refreshDuration;
     this.indexRoot = indexRoot;
     this.logFileName = logFileName;
-    this.ramBufferSizeMB = ramBufferSizeMB;
     this.enableTracing = enableTracing;
   }
 
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
@@ -78,7 +78,7 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
     // TODO: Chunk should create log store?
     LuceneIndexStoreConfig indexStoreCfg =
         new LuceneIndexStoreConfig(
-            commitInterval, refreshInterval, dataDirectory.getAbsolutePath(), false);
+            commitInterval, refreshInterval, dataDirectory.getAbsolutePath(), 8, false);
 
     // TODO: set ignore property exceptions via CLI flag.
     return new LuceneIndexStoreImpl(
@@ -140,6 +140,8 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
     final IndexWriterConfig indexWriterCfg =
         new IndexWriterConfig(analyzer)
             .setOpenMode(IndexWriterConfig.OpenMode.CREATE)
+            .setRAMBufferSizeMB(config.ramBufferSizeMB)
+            .setUseCompoundFile(false)
             .setMergeScheduler(new KalDBMergeScheduler(metricsRegistry))
             .setIndexSort(
                 new Sort(
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
@@ -21,6 +21,8 @@ import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.KeepOnlyLastCommitDeletionPolicy;
 import org.apache.lucene.index.SnapshotDeletionPolicy;
 import org.apache.lucene.search.SearcherManager;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
 import org.apache.lucene.store.FSDirectory;
 import org.apache.lucene.store.MMapDirectory;
 import org.slf4j.Logger;
@@ -139,6 +141,12 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
         new IndexWriterConfig(analyzer)
             .setOpenMode(IndexWriterConfig.OpenMode.CREATE)
             .setMergeScheduler(new KalDBMergeScheduler(metricsRegistry))
+            .setIndexSort(
+                new Sort(
+                    new SortField(
+                        LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName,
+                        SortField.Type.LONG,
+                        true)))
             .setIndexDeletionPolicy(snapshotDeletionPolicy);
 
     if (config.enableTracing) {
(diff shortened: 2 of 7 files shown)


bryanlb:
```
This PR:
	500 results, 60 buckets
		Iteration   1: 13.612 ops/s
		Iteration   2: 13.950 ops/s
		Iteration   3: 14.053 ops/s
		Iteration   4: 13.656 ops/s
		Iteration   5: 14.138 ops/s


	0 results, 60 buckets
		Iteration   1: 38.235 ops/s
		Iteration   2: 37.976 ops/s
		Iteration   3: 36.943 ops/s
		Iteration   4: 38.460 ops/s
		Iteration   5: 38.425 ops/s


	500 results, 0 buckets
		Iteration   1: 23.259 ops/s
		Iteration   2: 23.191 ops/s
		Iteration   3: 23.525 ops/s
		Iteration   4: 21.975 ops/s
		Iteration   5: 21.416 ops/s


Current master:
	500 results, 60 buckets
		Iteration   1: 14.788 ops/s
		Iteration   2: 14.189 ops/s
		Iteration   3: 13.920 ops/s
		Iteration   4: 13.977 ops/s
		Iteration   5: 14.458 ops/s


	0 results, 60 buckets
		Iteration   1: 34.733 ops/s
		Iteration   2: 35.802 ops/s
		Iteration   3: 35.682 ops/s
		Iteration   4: 35.426 ops/s
		Iteration   5: 34.976 ops/s


	500 results, 0 buckets
		Iteration   1: 17.951 ops/s
		Iteration   2: 18.801 ops/s
		Iteration   3: 18.430 ops/s
		Iteration   4: 18.738 ops/s
		Iteration   5: 18.512 ops/s
```

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -56,7 +63,18 @@ public class LogIndexSearcherImpl implements LogIndexSearcher<LogMessage> {
...
     MMapDirectory directory = new MMapDirectory(path);
-    return new SearcherManager(directory, null);
+    class CachedThreadPoolSearchFactory extends SearcherFactory {
+      private final ExecutorService executorService =
+          Executors.newCachedThreadPool(
+              new ThreadFactoryBuilder().setNameFormat("log-searcher-impl-%d").build());
+
+      @Override
+      public IndexSearcher newSearcher(IndexReader reader, IndexReader previousReader) {
+        return new IndexSearcher(reader, executorService);
```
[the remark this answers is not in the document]
bryanlb: Ok, we can certainly split this out.

The motivation behind this is when we have a query hit a node that only has a single chunk we only search it with a single thread. We're regularly matching fewer chunks than we have cores, so there's unused CPU if it's not serving multiple concurrent requests at once.

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -111,30 +129,39 @@ public SearchResult<LogMessage> search(
...
         }
 
-        Histogram histogram = statsCollector.histogram;
-
         elapsedTime.stop();
         return new SearchResult<>(
             results,
             elapsedTime.elapsed(TimeUnit.MICROSECONDS),
-            histogram.count(),
+            bucketCount > 0 ? histogram.count() : results.size(),
```
vthacker, line 164: niceee!! This is likely what will cause the perf improvements since lucene can short circuit

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
```diff
@@ -47,26 +44,15 @@ public static Duration getRefreshDuration(final long refreshDurationSecs) {
...
+      Duration commitDuration, Duration refreshDuration, String indexRoot, boolean enableTracing) {
+    this(commitDuration, refreshDuration, indexRoot, DEFAULT_LOG_FILE_NAME, enableTracing);
   }
 
   public LuceneIndexStoreConfig(
       Duration commitDuration,
       Duration refreshDuration,
       String indexRoot,
       String logFileName,
-      int ramBufferSizeMB,
```
mansu, line 69: I think we would need this field exposed still. Setting the RamBufferSize controls how frequently we flush the docs to disk which in turn affects the indexing throughput, cpu usage on indexing and data freshness. At 2MBPS, we flush every ~8 secs with the default value of 16MBPS. I think we should increase the value over default so we flush every 30 seconds or so for larger segments and more efficient compression and less merge overhead. Further, we would need to tune this value depending on how much data is written to each partition by the pre-processor.

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -111,30 +129,39 @@ public SearchResult<LogMessage> search(
       // This is a useful optimization for indexes that are static.
       IndexSearcher searcher = searcherManager.acquire();
       try {
-        TopFieldCollector topFieldCollector = buildTopFieldCollector(howMany);
-        StatsCollector statsCollector =
+        CollectorManager<TopFieldCollector, TopFieldDocs> topFieldCollector =
```
mansu, line 132: I think the code is easier to read if this code is moved into the conditional below.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -139,9 +141,13 @@ private IndexWriterConfig buildIndexWriterConfig(
     final IndexWriterConfig indexWriterCfg =
         new IndexWriterConfig(analyzer)
             .setOpenMode(IndexWriterConfig.OpenMode.CREATE)
-            .setRAMBufferSizeMB(config.ramBufferSizeMB)
-            .setUseCompoundFile(false)
```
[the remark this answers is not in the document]
bryanlb: Using the defaults here is better unless we have an explicit need to override them. The benefit with this is as we upgrade versions we float with the recommended best approaches instead of having old overrides that are forgotten.

I'll split this and rambuffer change into a separate PR though.

mansu:
A few questions about the benchmarks:

* Should the benchmarks be updated with the index sorting results?

* From the benchmarks calculating histogram takes roughly 2x the time than getting the topK results.
0 results, 60 buckets
		Iteration   1: 38.235 ops/s
500 results, 0 buckets
		Iteration   1: 17.951 ops/s

But calculating both histogram and topK takes less time than either:
500 results, 60 buckets
		Iteration   1: 14.788 ops/s

I am curious how we can explain the speed up? Could the speed up be due to a bug in our code? Can you ensure that we have unit tests(existing or new) that test all of these outputs to make sure the outputs match the expected result?

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
bryanlb: Let's see how the defaults work before attempting to override these values, as overriding these has downstream effects on how segments are flushed. OpenSearch does override this to a percentage of the overall total ([1](https://github.com/opensearch-project/OpenSearch/blob/a0030dfb47b4312eb4087f17cc3cee0139c74b6a/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L2470), [2](https://github.com/opensearch-project/OpenSearch/blob/a0030dfb47b4312eb4087f17cc3cee0139c74b6a/server/src/main/java/org/opensearch/index/engine/EngineConfig.java#L303)), but they're intentionally doing to take advantage of a custom `IndexingMemoryController`.

If we do intend to expose this as a config this needs reworked, as this setting isn't actually settable - it's just exposed as a param in this class and then just called with hardcoded values. I'll put the ramBuffer and compound file changes into a separate PR.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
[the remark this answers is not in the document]
bryanlb: https://github.com/slackhq/kaldb/pull/313

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
bryanlb: https://github.com/slackhq/kaldb/pull/313

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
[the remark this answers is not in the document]
bryanlb: Removed. We can add this again in a followup PR once this is merged.

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -139,9 +141,13 @@ private IndexWriterConfig buildIndexWriterConfig(
...
             .setMergeScheduler(new KalDBMergeScheduler(metricsRegistry))
+            .setIndexSort(
```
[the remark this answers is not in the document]
bryanlb: Done. Indexing didn't show any significant deviance in performance on benchmarks with a 4.5G file, where each run was 26k ops/s +/- 3k - with or without the sorting. There would be expected to have a small negative impact to indexing performance, though with local benchmarking it's difficult to detect.

bryanlb:
Updated benchmarks:

```
New Code:
  * 500,60 - 15.263 ops/s
  * 500,0 - 68.357 ops/s
  * 0,60 - 32.471 ops/s

Master:
  * 500,60 - 14.137 op/s
  * 500,0 - 18.546 op/s
  * 0,60 - 39.498 op/s

```

vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
```diff
@@ -26,6 +26,9 @@ public class LuceneIndexStoreConfig {
   // The name of the lucene log file.
   public final String logFileName;
 
+  // Config to set the IndexWriterConfig.setRAMBufferSizeMB.
+  public final int ramBufferSizeMB;
```
mansu, line 30: Delete ramBufferSizeMB?

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -111,30 +113,39 @@ public SearchResult<LogMessage> search(
...
             buildStatsCollector(bucketCount, startTimeMsEpoch, endTimeMsEpoch);
-        Collector collectorChain = MultiCollector.wrap(topFieldCollector, statsCollector);
 
-        searcher.search(query, collectorChain);
-        List<LogMessage> results;
         if (howMany > 0) {
-          ScoreDoc[] hits = topFieldCollector.topDocs().scoreDocs;
+          CollectorManager<TopFieldCollector, TopFieldDocs> topFieldCollector =
+              buildTopFieldCollector(howMany, bucketCount > 0);
+          MultiCollectorManager collectorManager = new MultiCollectorManager(topFieldCollector);
```
mansu, line 125: Can we initialize `collectorManager` only once?

kaldb/src/test/java/com/slack/kaldb/chunk/IndexingChunkImplTest.java
```diff
@@ -531,7 +531,7 @@ public void testSnapshotToS3UsingChunkApi() throws Exception {
       assertThat(chunk.snapshotToS3(bucket, "", s3BlobFs)).isTrue();
       assertThat(chunk.info().getSnapshotPath()).isNotEmpty();
 
-      assertThat(getCount(INDEX_FILES_UPLOAD, registry)).isEqualTo(4);
+      assertThat(getCount(INDEX_FILES_UPLOAD, registry)).isEqualTo(15);
```
mansu, line 534: I think this change should also be undone since we use compound files now.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -167,22 +178,45 @@ private LogMessage buildLogMessage(IndexSearcher searcher, ScoreDoc hit) {
     }
   }
 
-  private TopFieldCollector buildTopFieldCollector(int howMany) {
+  private CollectorManager<TopFieldCollector, TopFieldDocs> buildTopFieldCollector(
+      int howMany, boolean useMaxHitsThreshold) {
     if (howMany > 0) {
       SortField sortField = new SortField(SystemField.TIME_SINCE_EPOCH.fieldName, Type.LONG, true);
-      return TopFieldCollector.create(new Sort(sortField), howMany, howMany);
+      // if we need an accurate totalResults count then useMaxHitsThreshold should be true,
```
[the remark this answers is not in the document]
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: Done.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreConfig.java
bryanlb: Done.

kaldb/src/test/java/com/slack/kaldb/chunk/IndexingChunkImplTest.java
bryanlb: Done. This was a rebase issue.

mansu (approved):

vthacker (approved):

PR #315, opened 2022-06-06
Add a `ChunkDownloader` interface that allows us to download all the chunk data.

mansu:
Currently, the chunk downloading logic is hardcoded to an S3 API call. Add an abstraction around the snapshot downloading logic so we can plugin in different backends and also employ various downloading and rate limiting strategies.

Added a `ChunkDownloader` interface that provides this abstraction. In future, we can wrap this class with a rate limiter so we can easily rate limit the number of parallel rate limit calls.

Existing tests should validate the code, so no additional tests are needed.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/SerialS3ChunkDownloaderImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/SerialS3ChunkDownloaderImpl.java
@@ -5,7 +5,10 @@ import static com.slack.kaldb.logstore.BlobFsUtils.copyFromS3;
 import com.slack.kaldb.blobfs.BlobFs;
 import java.nio.file.Path;
 
-/** This downloader downloads all the chunk related data from S3 in a serial fashion. */
+/**
+ * SerialS3ChunkDownloaderImpl downloads all the chunk related data from S3 to a local store one
+ * file at a time.
+ */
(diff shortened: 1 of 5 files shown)


kaldb/src/main/java/com/slack/kaldb/chunk/SerialS3ChunkDownloaderImpl.java
```diff
@@ -0,0 +1,30 @@
...
+      String s3Bucket, String snapshotId, BlobFs blobFs, Path localDataDirectory) {
+    this.s3Bucket = s3Bucket;
+    this.snapshotId = snapshotId;
+    this.blobFs = blobFs;
+    this.dataDirectory = localDataDirectory;
+  }
+
+  @Override
+  public boolean download() throws Exception {
+    return copyFromS3(s3Bucket, snapshotId, blobFs, dataDirectory).length == 0;
```
vthacker, line 28: how would this API work for rate limiting? Wouldn't it need to be done on a per file basis?

mansu: Currently, the plan is to control the number of chunks downloaded in parallel. For that, I am adding a wrapper around this downloader that will limit the number of chunks downloaded in parallel in a different PR. Since, we only have a serial s3 file downloader, this will work for adding the rate limits.

vthacker (approved):

PR #318, opened 2022-06-09
zipkin API added & set up endpoints "getServices" and "getSpans"

shellywu815-zz:
Zipkin API setup according to https://zipkin.io/zipkin-api/#/default/
Added the helper function JsonUtil.writeJsonValues to parse json objects to json lists., which is the expected output
Screen Shot 2022-06-08 at 5 04 21 PM


kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ZipkinService.java
```diff
@@ -0,0 +1,186 @@
...
+    "OptionalUsedAsFieldOrParameterType") // Per https://armeria.dev/docs/server-annotated-service/
+public class ZipkinService {
+  private final KaldbServiceGrpc.KaldbServiceImplBase searcher;
+
+  public ZipkinService(KaldbServiceGrpc.KaldbServiceImplBase searcher) {
+    this.searcher = searcher;
+  }
+
+  @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+  public static final class GetServicesResult {
```
vthacker, line 30: Can we move this out to a separate file?

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ZipkinService.java
```diff
@@ -0,0 +1,186 @@
...
+      this.services = services;
+    }
+
+    @JsonProperty("services")
+    public List<String> services() {
+      return services;
+    }
+  }
+
+  public static final class Service {
```
vthacker, line 44: same with this and the span class - we can move this out to their individual files.

kaldb/src/main/java/com/slack/kaldb/util/JsonUtil.java
```diff
@@ -35,6 +35,12 @@ public static <T> T read(String s, TypeReference<T> valueTypeRef) throws JsonPro
     return ourInstance.mapper.readValue(s, valueTypeRef);
   }
 
+  public static <T> String writeJsonValues(String jsonString) {
```
vthacker, line 38: Can we write a small comment as to why we need this

We can call this method `writeJsonArray` since the Zipkin output format doesn't use a object but an array right?

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -0,0 +1,591 @@
...
+  }
+
+  /**
+   * Get Span Names API
+   *
+   * @return all the span names recorded by a particular service
+   * @see <a href="https://zipkin.io/zipkin-api/#/default/get_spans">API doc</a>
+   */
+  @Get("/spans")
+  public String getSpans(@Param("serviceName") Optional<String> serviceName) throws IOException {
```
vthacker, line 47: According to the documentation https://zipkin.io/zipkin-api/#/default/get_spans , `serviceName` is a required field. So I think we can remove the `Optional<String>` and make it just a `String` ?

vthacker: I think the next step here is to add start adding some of the additional parameters the GET spans api supports and using it when creating the `searchRequestBuilder`

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -0,0 +1,591 @@
...
+    return JsonUtil.writeJsonArray(JsonUtil.writeAsString(service));
+    */
+
+    // TODO: consider how to qualify the query variables
+    String queryString = "serviceName:" + serviceName;
+    KaldbSearch.SearchRequest.Builder searchRequestBuilder = KaldbSearch.SearchRequest.newBuilder();
+    KaldbSearch.SearchResult searchResult =
+        searcher.doSearch(
+            searchRequestBuilder
+                .setIndexName("testindex") // [Q] as of now we don't need to worry about index?
```
vthacker, line 90: yes - we can just hardcode this for now.

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -0,0 +1,591 @@
...
+      final String messageTimestamp = (String) source.get("@timestamp");
+      Instant instant = Instant.parse(messageTimestamp);
+      final long messageTimestampMicros =
+          TimeUnit.SECONDS.toMicros(instant.getEpochSecond())
+              + TimeUnit.NANOSECONDS.toMicros(instant.getNano());
+      final long messageDurationMicros = ((Number) source.get("duration_ms")).longValue();
+      final String messageServiceName = (String) source.get("service_name");
+      final String messageName = (String) source.get("name");
+      // [Q]what to put for msgtype
+      final String messageMsgType = "test message type";
```
vthacker, line 119: I would just change the `SpanUtil#makeSpanBuilder` class to add a null check around the  msgType and then here pass `null` 

```
    if (msgType != null) {
      tags.add(
              Trace.KeyValue.newBuilder()
                      .setKey(LogMessage.SystemField.TYPE.fieldName)
                      .setVTypeValue(Trace.ValueType.STRING.getNumber())
                      .setVStr(msgType)
                      .build());
    }
```

kaldb/src/main/java/com/slack/kaldb/zipkinApi/Service.java
```diff
@@ -0,0 +1,15 @@
+package com.slack.kaldb.zipkinApi;
+
+import java.util.List;
```
vthacker, line 3: Looks like this class is not used anymore? Can we remove it? Or if you want to keep as reference just comment it out for now?

kaldb/src/main/java/com/slack/kaldb/zipkinApi/Span.java
```diff
@@ -0,0 +1,91 @@
...
+  private final long timestamp;
+  private final long duration;
+  private final String kind;
+  private final HashMap<String, Object> localEndpoint;
+  private final HashMap<String, Object> remoteEndpoint;
+
+  private final List<HashMap<String, Object>> annotations;
+  private final HashMap<String, String> tags;
+
+  public Span(
```
vthacker, line 23: Same with this class. Can we just leave  a comment here for now that this class was for initial testing ( it's currently used by the main method and not any actual code ) and we will remove this class eventually.

vthacker:
We have added getTraceById ( https://github.com/slackhq/kaldb/pull/351 )  , this PR was just a POC that currently doesn't have any additional functionality

PR #319, opened 2022-06-10
update jackson library

vthacker:
update jackson library

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -19,8 +19,7 @@
         <armeria.version>1.16.0</armeria.version>
         <micrometer.version>1.8.5</micrometer.version>
         <kafka.version>2.5.1</kafka.version>
-        <jackson.version>2.10.5</jackson.version>
-        <jackson.databind.version>${jackson.version}.1</jackson.databind.version>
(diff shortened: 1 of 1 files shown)


kaldb/pom.xml
```diff
@@ -19,8 +19,7 @@
         <armeria.version>1.16.0</armeria.version>
         <micrometer.version>1.8.5</micrometer.version>
         <kafka.version>2.5.1</kafka.version>
-        <jackson.version>2.10.5</jackson.version>
-        <jackson.databind.version>${jackson.version}.1</jackson.databind.version>
```
vthacker, line 23: jackson databind exists for 2.13.3

mansu (dismissed):

solidspark (dismissed):

kaldb/pom.xml
bryanlb: We may want to leave this in fwiw - I looked into this a bit more after adding the initial comment, and it looks like the databind version pretty frequently has point releases past the major Jackson version.

mansu:
[Fixed here. Closing this one.](https://github.com/slackhq/kaldb/pull/327)

PR #324, opened 2022-06-14
Adjust chunk query threadpool

bryanlb:
Reviewing the performance data in our test environment it appears this cached threadpool is generating significantly more threads than is desired. This can exceed 100 active threads at a time, and has high thread turnover (~ 1k threads created per hour).

In order to better utilize the CPU this sets the max threads to twice that of the active cores or four - whichever is larger.


vthacker:
This makes sense! I think we should take this approach.

FYI we had it initially set to 200 and then we made it unlimited in this PR -https://github.com/slackhq/kaldb/pull/207

vthacker (approved):

PR #330, opened 2022-06-21
KalDbDistributedQueryService.java: Use loops instead of streams

kx-chen:
Refactor `getSearchNodesToQuery` to use loops instead of streams


mansu (changes requested):

mansu (approved):
Looks great! Thanks for addressing the TODO.

PR #331, opened 2022-06-23
Improve ES api instrumentation

bryanlb:
Collection of small refactors that should help improve the visibility when troubleshooting ES responses including:

* Support for common tracing tags
* Tracing ID in the response payload
* Return took/shard information into the response payload
* Adding additional meters for distributed query performance

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -32,6 +32,8 @@ s3Config:
 
 tracingConfig:
   zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-http://localhost:9411/api/v2/spans}
+  commonTags:
+    clusterName: ${KALDB_CLUSTER_NAME:-kaldb_local}
 
 queryConfig:
   serverConfig:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -34,6 +34,7 @@ tracingConfig:
   zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-http://localhost:9411/api/v2/spans}
   commonTags:
     clusterName: ${KALDB_CLUSTER_NAME:-kaldb_local}
+    env: ${KALDB_CLUSTER_ENV:-local}
 
 queryConfig:
   serverConfig:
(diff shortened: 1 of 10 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -80,6 +96,17 @@ public KaldbDistributedQueryService(
     searchMetadataTotalChangeCounter = meterRegistry.counter(SEARCH_METADATA_TOTAL_CHANGE_COUNTER);
     this.searchMetadataStore.addListener(this::updateStubs);
 
+    this.distributedQueryTotalNodes =
```
mansu, line 99: Also, do we need this level of granularity in the data? I think it may be simpler if we can just count total number of requests, timed out requests and partial requests.

If we need it, I am curious why these values are guages? If we run multiple queries, a guage would only record the results of the last query. I think a counter that is incremented by the number of failed and successful nodes is better.

mansu (changes requested):

config/config.yaml
```diff
@@ -32,6 +32,8 @@ s3Config:
 
 tracingConfig:
   zipkinEndpoint: ${ZIPKIN_TRACING_ENDPOINT:-http://localhost:9411/api/v2/spans}
+  commonTags:
+    clusterName: ${KALDB_CLUSTER_NAME:-kaldb_local}
```
mansu, line 36: add env as a tag also.

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
bryanlb: Swapped these out for counters, and removed the rate gauge entirely since that can now be calculated from the counters.

config/config.yaml
bryanlb: Done.

bryanlb: Added the env, but left off the component for now. An individual node can participate in one or more roles, so mapping a single key/value to a comma separated list isn't tremendously useful.

mansu (approved):

PR #332, opened 2022-07-06
Change preprocessor to start from latest offset on bootstrap

vthacker:
When we start a new service, indexing from latest offset would be a better start experience.

From my understanding this setting defines the behavior of the consumer when there is no committed position (which would be the case when the group is first initialized) or when an offset is out of range.

So once the bootstrap of a new service has taken place this setting isn't going to govern anything.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
@@ -310,6 +310,7 @@ public class PreprocessorService extends AbstractService {
     props.put(StreamsConfig.APPLICATION_ID_CONFIG, kafkaStreamConfig.getApplicationId());
     props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaStreamConfig.getBootstrapServers());
     props.put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, kafkaStreamConfig.getNumStreamThreads());
+    props.put("auto.offset.reset", "latest");
 
     // This will allow parallel processing up to the amount of upstream partitions. You cannot have
     // more threads than you have upstreams due to how the work is partitioned


kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -310,6 +310,7 @@ protected static Properties makeKafkaStreamsProps(
     props.put(StreamsConfig.APPLICATION_ID_CONFIG, kafkaStreamConfig.getApplicationId());
     props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, kafkaStreamConfig.getBootstrapServers());
     props.put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, kafkaStreamConfig.getNumStreamThreads());
+    props.put("auto.offset.reset", "latest");
```
bryanlb, line 313: Couple issues - this should use the variable for the key instead of the string. Additionally, since this is a stream config we need to use prefix approach for the consumer (see https://github.com/slackhq/kaldb/blob/master/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java#L324).

Other than those two issues I agree that the latest offset for a new consumer is desired and expected. For a topic with several hours/days of retention that you're setting up for the first time you don't expect this to start from the beginning of time. This results in a lengthy amount of time before you start seeing recent data, and puts a significant load on your cluster depending on how many preprocessors you deployed.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
mansu: So, this setting is only useful for new consumers. In those cases, we didn't set the offset to latest since we would loose data. If a consumer offset needs to be reset to latest, a user can do it easily via Kafka CLI. Since we are a DB, I don't think we should set this config in our default config since it results in data loss. 

Another way to set this is to expose all the Kafka config as a key value map in the config. In those cases, we can allow the user to set what ever config he choses.

bryanlb: I don't think `latest` is a problem here, since that data was produced prior to KalDB being deployed. As soon as you deploy KalDB with `earliest`, you will start reading from the topic at the maximum speed the preprocessor can achieve.. and then discarding everything over your configured rate limit. Either way this results in "data loss" - but without setting earliest you not only have data loss, but also put a significant load on Kafka and will have multiple hours before the system is caught up to realtime.

That said, I think we need to move these all to the config map for both components (preprocessor stream and index consumer), as we need the ability to turn specific values per deployment.

PR #333, opened 2022-07-06
Bump Armeria version

bryanlb:
Bumps Armeria and related dependencies to latest where possible. Scala version conflicts made this a bit more interesting than it should have been, but this gets us closer to latest across the board.

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -14,17 +14,17 @@
         <uberjar.name>kaldb</uberjar.name>
         <http.client.version>4.5.13</http.client.version>
         <http.core.version>4.4.15</http.core.version>
-        <protobuf.version>3.20.1</protobuf.version>
-        <grpc.version>1.46.0</grpc.version>
-        <armeria.version>1.16.0</armeria.version>
-        <micrometer.version>1.8.5</micrometer.version>
-        <kafka.version>2.5.1</kafka.version>
+        <protobuf.version>3.21.1</protobuf.version>
+        <grpc.version>1.47.0</grpc.version>
+        <armeria.version>1.17.0</armeria.version>
+        <micrometer.version>1.9.1</micrometer.version>
+        <kafka.version>3.2.0</kafka.version>
         <jackson.version>2.13.3</jackson.version>
-        <jackson.databind.version>2.10.5.1</jackson.databind.version>
+        <jackson.databind.version>2.12.7</jackson.databind.version>
         <lucene.version>8.4.1</lucene.version>
         <curator.version>5.2.1</curator.version>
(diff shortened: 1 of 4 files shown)


kaldb/pom.xml
```diff
@@ -14,17 +14,17 @@
...
-        <micrometer.version>1.8.5</micrometer.version>
-        <kafka.version>2.5.1</kafka.version>
+        <protobuf.version>3.21.1</protobuf.version>
+        <grpc.version>1.47.0</grpc.version>
+        <armeria.version>1.17.0</armeria.version>
+        <micrometer.version>1.9.1</micrometer.version>
+        <kafka.version>3.2.0</kafka.version>
         <jackson.version>2.13.3</jackson.version>
-        <jackson.databind.version>2.10.5.1</jackson.databind.version>
+        <jackson.databind.version>2.12.7</jackson.databind.version>
```
bryanlb, line 23: This gets us to a version without any known vulnerabilities - https://snyk.io/vuln/maven:com.fasterxml.jackson.core%3Ajackson-databind

kaldb/pom.xml
```diff
@@ -14,17 +14,17 @@
...
-        <protobuf.version>3.20.1</protobuf.version>
-        <grpc.version>1.46.0</grpc.version>
-        <armeria.version>1.16.0</armeria.version>
-        <micrometer.version>1.8.5</micrometer.version>
-        <kafka.version>2.5.1</kafka.version>
+        <protobuf.version>3.21.1</protobuf.version>
+        <grpc.version>1.47.0</grpc.version>
+        <armeria.version>1.17.0</armeria.version>
+        <micrometer.version>1.9.1</micrometer.version>
+        <kafka.version>3.2.0</kafka.version>
```
mansu, line 21: Let's leave Kafka version as is. Also, unless there is a known advantage let's not bump up the versions of the software that we don't need to use it.

mansu (changes requested):

kaldb/pom.xml
bryanlb: I had to upgrade this due to the fact we need to upgrade `kafka-junit`. We need to upgrade `kafka-junit` because of an issue with `jackson.databind` and Scala version mismatches (see the support matrix here - https://github.com/charithe/kafka-junit).

`jackson.databind` has to be updated due to multiple security vulnerabilities, and 2.12 is the oldest available that doesn't have those vulnerabilities.

The dependencies listed here were the minimum set I was able to find that satisfied upgrading Armeria and its associated dependencies (ie, micrometer, java grpc, brave), and addressed the vulnerable packages.

mansu (approved):

mansu (approved):

PR #338, opened 2022-07-13
Add `ReplicaRestoreService`

kx-chen:
* Add a new service, `ReplicaRestoreService`, responsible for rehydrating Snapshots
* Add tests for said service
* Hook up the new `ReplicaRestoreService` to the `restoreSnapshots` GRPC endpoint in the manager API

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -78,6 +78,10 @@ managerConfig:
   snapshotDeletionServiceConfig:
     schedulePeriodMins: ${KALDB_MANAGER_SNAPSHOT_DELETE_PERIOD_MINS:-15}
     snapshotLifespanMins: ${KALDB_MANAGER_SNAPSHOT_LIFESPAN_MINS:-10080}
+  replicaRestoreServiceConfig:
+    schedulePeriodMins: ${KALDB_MANAGER_REPLICA_RESTORE_PERIOD_MINS:-15}
+    maxReplicasPerRequest: ${KALDB_MANAGER_REPLICA_RESTORE_MAX_REPLICAS_PER_REQUEST:-200}
+    replicaLifespanMins: ${KALDB_MANAGER_REPLICA_RESTORE_LIFESPAN_MINS:-60}
 
 clusterConfig:
   clusterName: ${KALDB_CLUSTER_NAME:-kaldb_local}
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
@@ -1,0 +1,177 @@
+package com.slack.kaldb.clusterManager;
+
+import static com.slack.kaldb.clusterManager.ReplicaCreationService.replicaMetadataFromSnapshotId;
+
+import com.google.common.util.concurrent.AbstractScheduledService;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.slack.kaldb.metadata.replica.ReplicaMetadata;
+import com.slack.kaldb.metadata.replica.ReplicaMetadataStore;
+import com.slack.kaldb.metadata.snapshot.SnapshotMetadata;
+import com.slack.kaldb.proto.config.KaldbConfigs;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import javax.naming.SizeLimitExceededException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Scheduled service responsible for restoring user requested Snapshots that have expired. Users can
+ * call the queueSnapshotsForRestoration to queue Snapshots for restoration in the future. This
+ * service will automatically handle de-duping, ensuring that no Snapshot is restored more than
+ * once. Additionally, the maximum number of Snapshots that can be requested at once is also
+ * configurable to prevent overwhelming the service.
+ */
+public class ReplicaRestoreService extends AbstractScheduledService {
+  private ScheduledFuture<?> pendingTask;
+  private final KaldbConfigs.ManagerConfig managerConfig;
+  private final ScheduledExecutorService executorService =
+      Executors.newSingleThreadScheduledExecutor(
+          new ThreadFactoryBuilder().setNameFormat("replica-restore-service-%d").build());
+  private final BlockingQueue<SnapshotMetadata> queue = new LinkedBlockingQueue<>();
+  private final ReplicaMetadataStore replicaMetadataStore;
+  private final MeterRegistry meterRegistry;
+  private final Counter successfullyCreatedReplicas;
+  private final Counter failedReplicas;
+  private final Counter skippedReplicas;
+  private final Timer restoreTimer;
+  protected static final Logger LOG = LoggerFactory.getLogger(ReplicaCreationService.class);
+  public static String REPLICAS_CREATED = "replicas_created";
+  public static String REPLICAS_FAILED = "replicas_failed";
+  public static String REPLICAS_SKIPPED = "replicas_skipped";
+  public static String REPLICAS_RESTORE_TIMER = "replicas_restore_timer";
+
+  public ReplicaRestoreService(
+      ReplicaMetadataStore replicaMetadataStore,
+      MeterRegistry meterRegistry,
+      KaldbConfigs.ManagerConfig managerConfig) {
+    this.managerConfig = managerConfig;
+    this.replicaMetadataStore = replicaMetadataStore;
+    this.meterRegistry = meterRegistry;
+    this.skippedReplicas = meterRegistry.counter(REPLICAS_SKIPPED);
+    this.successfullyCreatedReplicas = meterRegistry.counter(REPLICAS_CREATED);
+    this.failedReplicas = meterRegistry.counter(REPLICAS_FAILED);
+    this.restoreTimer = meterRegistry.timer(REPLICAS_RESTORE_TIMER);
+  }
+
+  @Override
+  protected void runOneIteration() {
+    if (pendingTask == null || pendingTask.getDelay(TimeUnit.SECONDS) <= 0) {
+      pendingTask =
+          executorService.schedule(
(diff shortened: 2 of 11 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+  public static String REPLICAS_RESTORE_TIMER = "replicas_restore_timer";
+
+  public ReplicaRestoreService(
+      ReplicaMetadataStore replicaMetadataStore,
+      MeterRegistry meterRegistry,
+      KaldbConfigs.ManagerConfig managerConfig) {
+    this.managerConfig = managerConfig;
+    this.replicaRestoreServiceConfig = managerConfig.getReplicaRestoreServiceConfig();
+
+    this.queue = new LinkedBlockingQueue<>();
```
kx-chen, line 51: Just a quick note: I had to switch to using a [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html) as a [ArrayBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ArrayBlockingQueue.html) is bounded

config/config.yaml
```diff
@@ -78,6 +78,10 @@ managerConfig:
   snapshotDeletionServiceConfig:
     schedulePeriodMins: ${KALDB_MANAGER_SNAPSHOT_DELETE_PERIOD_MINS:-15}
     snapshotLifespanMins: ${KALDB_MANAGER_SNAPSHOT_LIFESPAN_MINS:-10080}
+  replicaRestoreServiceConfig:
+    schedulePeriodMins: ${KALDB_MANAGER_REPLICA_RESTORE_PERIOD_MINS:-15}
+    maxReplicasPerRequest: ${KALDB_MANAGER_REPLICA_RESTORE_MAX_REPLICAS_PER_REQUEST:-200}
+    replicaLifetimeMins: ${KALDB_MANAGER_REPLICA_RESTORE_LIFETIME_MINS:-60}
```
bryanlb, line 84: Above we call this config the `replicaLifespanMins` - probably should keep that same name here as it's the same functionality.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.*;
+import javax.naming.SizeLimitExceededException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Scheduled service responsible for queueing and creating replica snapshots */
```
bryanlb, line 24: We should expand this definition slightly - currently this doesn't tell us much about the purpose of this class that we can't get from reading the first line of the class definition.

We should say something explaining what this is used for (ie, user requested snapshots that have expired), a mention of the fact it prevents additional snapshots from being created when they already exist, what protections we have in place to prevent a user from overwhelming this service, etc.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+public class ReplicaRestoreService extends AbstractScheduledService {
+  private ScheduledFuture<?> pendingTask;
+  private final KaldbConfigs.ManagerConfig managerConfig;
+  private final KaldbConfigs.ManagerConfig.ReplicaRestoreServiceConfig replicaRestoreServiceConfig;
+  private final ScheduledExecutorService executorService =
+      Executors.newSingleThreadScheduledExecutor();
```
bryanlb, line 30: This probably should have a thread factory provided to it - I know we're not super consistent with this in the codebase but it makes troubleshooting a bit easier since the thread name is printed into the logs.

Check out ReadOnlyChunkImpl:91 for an example of us passing a named thread factory to this

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+  public static String REPLICAS_FAILED = "replicas_failed";
+  public static String REPLICAS_SKIPPED = "replicas_skipped";
+  public static String REPLICAS_RESTORE_TIMER = "replicas_restore_timer";
+
+  public ReplicaRestoreService(
+      ReplicaMetadataStore replicaMetadataStore,
+      MeterRegistry meterRegistry,
+      KaldbConfigs.ManagerConfig managerConfig) {
+    this.managerConfig = managerConfig;
+    this.replicaRestoreServiceConfig = managerConfig.getReplicaRestoreServiceConfig();
```
bryanlb, line 49: I'm slightly confused as to why we set both the manager config, and then a subset of the manager config to their own variables. We probably only need one or the other here.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
bryanlb, line 51: We can init this all in one go on line 31 above.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @Override
+  protected Scheduler scheduler() {
+    return Scheduler.newFixedRateSchedule(
+        managerConfig.getScheduleInitialDelayMins(),
+        replicaRestoreServiceConfig.getSchedulePeriodMins(),
+        TimeUnit.MINUTES);
+  }
+
+  /** Queues snapshots to be restored in the future */
```
bryanlb, line 97: This should have a slightly expanded doc. This will be the method users of this class will be using, so we want to make sure they know what to expect (ie, throws an exception if you submit too many).

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+
+      try {
+        replicaMetadataStore.createSync(
+            replicaMetadataFromSnapshotId(
+                snapshot.snapshotId,
+                Instant.now()
+                    .plus(
+                        replicaRestoreServiceConfig.getReplicaLifetimeMins(), ChronoUnit.MINUTES)));
+      } catch (Exception e) {
+        LOG.error("Error restoring replica for snapshot {}", snapshot.snapshotId);
```
bryanlb, line 151: We should include the error here - anytime there's an error we should always pass the exception

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
```diff
@@ -171,6 +180,29 @@ public void updatePartitionAssignment(
     }
   }
 
+  @Override
+  public void restoreSnapshot(
+      ManagerApi.RestoreSnapshotRequest request,
+      StreamObserver<ManagerApi.RestoreSnapshotResponse> responseObserver) {
+    List<SnapshotMetadata> snapshotsToRestore =
```
bryanlb, line 187: You may want to add some validation here around the request - ie, what happens if you have a GRPC request with nonsense data? Like an end time before a start time? Or an empty service name?

kaldb/src/test/java/com/slack/kaldb/server/ManagerApiGrpcTest.java
```diff
@@ -539,4 +565,91 @@ public void shouldFetchSnapshotsWithinTimeframeAndPartition() {
...
+            1,
+            List.of(new ServicePartitionMetadata(startTime + 5, startTime + 6, List.of("a", "b"))));
+
+    serviceMetadataStore.createSync(serviceWithDataInPartitionA);
+
+    await().until(() -> serviceMetadataStore.getCached().size() == 1);
+    await().until(() -> snapshotMetadataStore.getCached().size() == 3);
+
+    replicaRestoreService.startAsync();
+    replicaRestoreService.awaitRunning();
```
bryanlb, line 640: I see a start without a stop here - those should always be in pairs.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -0,0 +1,160 @@
...
+        managerConfig.getReplicaRestoreServiceConfig().getSchedulePeriodMins(),
+        TimeUnit.MINUTES);
+  }
+
+  /** Queues snapshots to be restored in the future */
+  public synchronized void queueSnapshotsForRestoration(List<SnapshotMetadata> snapshotsToRestore)
+      throws Exception {
+    if (snapshotsToRestore.size()
+        >= managerConfig.getReplicaRestoreServiceConfig().getMaxReplicasPerRequest()) {
+      throw new SizeLimitExceededException();
```
bryanlb, line 100: You probably want to add a message in the `SizeLimitExceededException` params here that says how many were requested, and what the limit actually is. That'll bubble up via your GRPC service and be returned to the user then given the current code.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -20,6 +20,8 @@ service ManagerApiService {
   //
   // In the future if only a rate is provided with an empty list the allocation will be automatically assigned.
   rpc UpdatePartitionAssignment(UpdatePartitionAssignmentRequest) returns (UpdatePartitionAssignmentResponse) {}
+
+  rpc RestoreSnapshot(RestoreSnapshotRequest) returns (RestoreSnapshotResponse) {}
```
bryanlb, line 24: These are `Restore`__`Snapshot`__ requests, but are passed to a __`Replica`__`RestoreService`. I would use consistent terminology for both the GRPC api and the service (ie, are we restoring snapshots, or replicas? Probably the latter, since the replicas are what are being re-created).

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
```diff
@@ -171,6 +179,36 @@ public void updatePartitionAssignment(
...
+              request.getServiceName());
+
+      replicaRestoreService.queueSnapshotsForRestoration(snapshotsToRestore);
+
+      responseObserver.onNext(
+          ManagerApi.RestoreSnapshotResponse.newBuilder().setStatus("success").build());
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      LOG.error("Error restoring snapshot", e);
+      responseObserver.onError(Status.UNKNOWN.withDescription(e.getMessage()).asException());
```
bryanlb, line 208: If you catch the `SizeLimitExceededException` you can log that out as only an `info` level, and provide a more targeted status code in the GRPC response (perhaps `RESOURCE_EXHAUSTED` or `INVALID_ARGUMENT`? either seems like it could apply).

bryanlb (approved):

PR #340, opened 2022-07-13
make timeouts configurable per service

vthacker:


config/config.yaml
```diff
@@ -53,16 +57,19 @@ cacheConfig:
   slotsPerInstance: ${KALDB_CACHE_SLOTS_PER_INSTANCE:-10}
   dataDirectory: ${KALDB_CACHE_DATA_DIR:-/tmp}
   maxParallelCacheSlotDownloads: ${KALDB_CACHE_MAX_PARALLEL_CACHE_SLOT_DOWNLOADS:-3}
+  defaultQueryTimeoutMs: ${DEFAULT_QUERY_TIMEOUT_MS:-2500}
```
bryanlb, line 60: Probably want to prefix these with the component like the other envs. This way if you want to run KalDB in a single JVM you can still make use of different timeout values per service.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -168,10 +168,15 @@ private static Set<Service> getServices(
...
+      KaldbLocalQueryService<LogMessage> searcher =
+          new KaldbLocalQueryService<>(
+              chunkManager,
+              Duration.ofMillis(kaldbConfig.getIndexerConfig().getDefaultQueryTimeoutMs()));
       final int serverPort = kaldbConfig.getIndexerConfig().getServerConfig().getServerPort();
+      final long requestTimeoutMs =
+          kaldbConfig.getIndexerConfig().getServerConfig().getRequestTimeoutMs();
       ArmeriaService armeriaService =
-          new ArmeriaService.Builder(serverPort, "kalDbIndex", meterRegistry)
+          new ArmeriaService.Builder(serverPort, requestTimeoutMs, "kalDbIndex", meterRegistry)
```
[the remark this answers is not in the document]
vthacker: good call! Fixed

kaldb/src/main/java/com/slack/kaldb/server/ArmeriaService.java
```diff
@@ -57,10 +56,16 @@ private ArmeriaService(Server server, String serviceName) {
   public static class Builder {
     private final String serviceName;
     private final ServerBuilder serverBuilder;
+    private final Duration requestTimeout;
     private final List<SpanHandler> spanHandlers = new ArrayList<>();
 
-    public Builder(int port, String serviceName, PrometheusMeterRegistry prometheusMeterRegistry) {
+    public Builder(
+        int port,
+        Duration requestTimeout,
```
bryanlb, line 64: This probably shouldn't be an argument on the constructor, but instead a builder method.

The only arguments that are expected to go to a builder's constructor are those that are required (when it's not a step-builder). Request timeout isn't a required argument, and if it's not provided we can leave it unset and fallback to the default Armeria behavior.

bryanlb (approved):

PR #341, opened 2022-07-14
Rename `ServiceMetadata` to `DatasetMetadata`

mansu:
Similarly rename `ServicePartitionMetadata` to `DatasetPartitionMetadata`.

Updated all code references. Now we don't have different meanings for the word service in the code. Service is now called dataset.


kaldb/src/main/java/com/slack/kaldb/metadata/dataset/DatasetMetadataStore.java
```diff
@@ -1,22 +1,22 @@
...
+package com.slack.kaldb.metadata.dataset;
 
 import com.slack.kaldb.metadata.core.PersistentMutableMetadataStore;
 import com.slack.kaldb.metadata.zookeeper.MetadataStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ServiceMetadataStore extends PersistentMutableMetadataStore<ServiceMetadata> {
+public class DatasetMetadataStore extends PersistentMutableMetadataStore<DatasetMetadata> {
   public static final String SERVICE_METADATA_STORE_ZK_PATH = "/service";
```
bryanlb, line 9: We should add a todo here explaining why this is still service, and what we expect to happen

bryanlb (changes requested):
Just need a comment explaining why dataset has a service path as we discussed, then this looks good to go after a rebase.

bryanlb (approved):

PR #343, opened 2022-07-22
Rework query performance stats

bryanlb:
Total nodes / failed nodes is not a useful measurement, as long queries skew the results with much larger node counts. This PR switches the instrumentation to an [apdex](https://www.apdex.org/wp-content/uploads/2020/09/ApdexTechnicalSpecificationV11_000.pdf) model.

This will allow us to calculate the percent of queries that have no failed nodes, as well as the near-misses so we can track this performance over time.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -58,15 +58,21 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
   private final Map<String, KaldbServiceGrpc.KaldbServiceFutureStub> stubs =
       new ConcurrentHashMap<>();
 
-  public static final String DISTRIBUTED_QUERY_TOTAL_NODES = "distributed_query_total_nodes";
-  public static final String DISTRIBUTED_QUERY_FAILED_NODES = "distributed_query_failed_nodes";
+  public static final String DISTRIBUTED_QUERY_APDEX_SATISFIED =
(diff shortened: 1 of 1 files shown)


vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -361,8 +371,17 @@ public KaldbSearch.SearchResult doSearch(KaldbSearch.SearchRequest request) {
                   new SearchResultAggregatorImpl<>(SearchResultUtils.fromSearchRequest(request)))
               .aggregate(searchResults);
 
-      distributedQueryTotalNodes.increment(aggregatedResult.totalNodes);
-      distributedQueryFailedNodes.increment(aggregatedResult.failedNodes);
+      // We report a query with more than 0% of requested nodes, but less than 2% as a tolerable
+      // response. Anything over 2% is considered an unacceptable.
+      if (aggregatedResult.totalNodes == 0 || aggregatedResult.failedNodes == 0) {
```
vthacker, line 376: `aggregatedResult.totalNodes == 0` is a good check so that the remaining if-else doesn't get into funky divisions

bryanlb: Indeed. 

Theoretically we should never have non-zero failed nodes and zero total nodes, but it doesn't hurt to check. 😄

mansu (commented):
I think it's best not to do appdex calculations in the code.

PR #344, opened 2022-07-27
Revert "Add a rate limiter for chunk downloader. (#316)"

bryanlb:
This reverts commit 8fa3afd5180ccddf9e0d16412fe0621db053d601 (#316). This approach was unsuccessful for several reasons; one - because regardless of the amount of chunks, they are downloaded at the maximum throughput possible; two - multiple instances can be deployed to a single logical host which still results in host saturation; three - instance types have different throughput maximums and this approach provides no mechanisms to throttle against this.

We will be pursing the approach in #232 (or similar) to set a maximum download in throughput instead of chunk concurrency.


vthacker (approved):

bryanlb:
To followup on this - we'll be pursuing using Kubernetes [traffic shaping](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping) for rate limiting at this point instead of a code solution.

PR #345, opened 2022-07-29
Add endpoint to restore replica from Snapshot ID

kx-chen:
* Add new endpoint on ManagerApiGrpc to restore replicas from specific Snapshot IDs
* Add new test, `shouldRestoreGivenSnapshotIds`, to test the new endpoint
* Refactor ManagerApiGrpc and ReplicaRestoreService code to use snapshot IDs exclusively (previously it was using SnapshotMetadatas)

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
@@ -4,6 +4,7 @@ import static com.slack.kaldb.metadata.dataset.DatasetMetadataSerializer.toDatas
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Sets;
 import com.slack.kaldb.chunk.ChunkInfo;
 import com.slack.kaldb.clusterManager.ReplicaRestoreService;
 import com.slack.kaldb.metadata.dataset.DatasetMetadata;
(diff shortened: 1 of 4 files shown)


kaldb/src/test/java/com/slack/kaldb/server/ManagerApiGrpcTest.java
```diff
@@ -89,7 +88,7 @@ public void setUp() throws Exception {
             .build();
 
     replicaRestoreService =
-        new ReplicaRestoreService(replicaMetadataStore, meterRegistry, managerConfig);
+        spy(new ReplicaRestoreService(replicaMetadataStore, meterRegistry, managerConfig));
```
bryanlb, line 91: This shouldn't need a spy correct?

kaldb/src/main/proto/manager_api.proto
```diff
@@ -80,3 +81,11 @@ message RestoreReplicaRequest {
 message RestoreReplicaResponse {
   string status = 1;
 }
+
+message RestoreReplicaIdsRequest {
+  repeated string ids_to_restore = 2;
```
bryanlb, line 86: This should be a `1`

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/ManagerApiGrpc.java
```diff
@@ -250,6 +276,28 @@ protected static List<SnapshotMetadata> fetchSnapshots(
     return snapshotMetadata;
   }
 
+  /**
+   * Fetches all SnapshotMetadata that match the IDs in snapshotIds
+   *
+   * @return List of SnapshotMetadata that are within specified timeframe and from queried service
+   */
+  protected static List<SnapshotMetadata> fetchSnapshots(
```
bryanlb, line 284: Small nit - this isn't really "fetching" as much as it is just calculating the value from the provided values.

bryanlb (approved):

bryanlb (approved):

PR #346, opened 2022-08-03
Kaldb zipkin api

shellywu815:
Added the skeleton code for functions in ZipkinService.java and added a zipkinService instance into Kaldb.java

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
@@ -22,8 +22,12 @@ import java.util.*;
 public class ZipkinService {
   private final KaldbQueryServiceBase searcher;
 
+
+
   public ZipkinService(KaldbQueryServiceBase searcher) {
     this.searcher = searcher;
+
+    System.out.println("test");
   }
 
   @Get
--- a/kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
@@ -1,0 +1,57 @@
+package com.slack.kaldb.zipkinApi;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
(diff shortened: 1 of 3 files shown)


kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -0,0 +1,57 @@
...
+            @Param("limit") @Default("10") Integer limit)
+            throws IOException {
+        return null;
+    }
+
+    @Get("/api/v2/trace/{traceId}")
+    public HttpResponse getTraceByTraceId(@Param("traceId") String traceId) throws IOException {
+        return null;
+    }
+}
```
vthacker, line 57: maybe add a newline at the end?

vthacker:
Nice! This is a good start. Can you make two changes to this
1. Make any of the 4 functions that you have implemented not return null, but return `httpresponse(OK,string)` where the string can be `[]` 
2. Add a test like ElasticsearchApiServiceTest to verify if you are able to query ZipkinService

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -0,0 +1,52 @@
...
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.server.annotation.*;
+import com.slack.kaldb.server.KaldbQueryServiceBase;
+import java.io.IOException;
+import java.util.*;
+
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+public class ZipkinService {
```
bryanlb, line 12: We should have a comment here that explains what the purpose of the class is, sorta like https://github.com/slackhq/kaldb/blob/master/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java#L37 or https://github.com/slackhq/kaldb/blob/master/kaldb/src/main/java/com/slack/kaldb/clusterManager/RecoveryTaskAssignmentService.java#L36

bryanlb (approved):
Suggest adding a comment, other than that this looks fine 👍

PR #349, opened 2022-08-05
Switch to directory based rollover

bryanlb:
Switches from indexed message size rollover to a size-on-disk rollover


vthacker:
I like this! Very simple and works would work effectively

Two small nits:
1. Maybe `directorySize` variable doesn't convey the meaning? Maybe `directorySizeCalculator` ? 
2. Can we add a one assert test for the gauge that we added?

vthacker (approved):

PR #348, opened 2022-08-05
Kaldb zipkin api

shellywu815:
I have added the functionalities for getTraceByTraceId functionalities in the Zipkin API. Additionally added related helper functions and tests

PR #352, opened 2022-08-09
use standard span field names in the proto definition 

vthacker:
While @shellywu815  was working on https://github.com/slackhq/kaldb/pull/348 , she had to add `ZipkinSpan` to the trace.proto . Ideally we should be able to use the Span proto that exists

So here are two changes to variable names to the Span proto that don't conform to the standard. Using different variable names causes frontend applications ( in our case Grafana ) to not be able to render the timestamp and duration correctly.

### DIFF
--- a/kaldb/src/main/proto/trace.proto
+++ b/kaldb/src/main/proto/trace.proto
@@ -57,12 +57,12 @@ message Span {
   // For example: a value of 1551849569000000 represents March 6, 2019 5:19:29 UTC.
   // We use fixed64 since it is more wire efficient for than int64 for larger numbers.
   // This field is required.
-  fixed64 start_timestamp_micros = 5;
+  fixed64 timestamp = 5;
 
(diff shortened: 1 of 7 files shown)


kaldb/src/main/proto/trace.proto
```diff
@@ -57,12 +57,12 @@ message Span {
   // For example: a value of 1551849569000000 represents March 6, 2019 5:19:29 UTC.
   // We use fixed64 since it is more wire efficient for than int64 for larger numbers.
   // This field is required.
-  fixed64 start_timestamp_micros = 5;
+  fixed64 timestamp = 5;
```
vthacker, line 60: https://github.com/openzipkin/zipkin-api/blob/master/zipkin.proto#L123-L129

this is what the name of the field is on the zipkin proto ( also linked in this proto file )

solidspark (approved):

PR #354, opened 2022-08-11
Add `managerConnectString` to query service config

kx-chen:
Add `managerConnectString` to query service config.
To be used by the query service for connections to the manager.


vthacker (approved):
LGTM! as discussed in the call I think longer term we should have the manager discovered via ZK 

For now this is the quickest way forward and a good solution.

bryanlb:
@alllcontributors add @kx-chen for code

bryanlb:
@allcontributors add @kx-chen for code

PR #355, opened 2022-08-17
Restore replicas upon query

kx-chen:
Automatically call the restore API when any `Snapshots` that need to be searched cannot be found

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -44,7 +44,7 @@ queryConfig:
     serverAddress: ${KALDB_QUERY_SERVER_ADDRESS:-localhost}
     requestTimeoutMs: ${KALDB_QUERY_REQUEST_TIMEOUT_MS:-5000}
   defaultQueryTimeoutMs: ${KALDB_QUERY_DEFAULT_QUERY_TIMEOUT_MS:-3000}
-  managerConnectString: ${KALDB_MANAGER_CONNECTION_STRING:-localhost:8083}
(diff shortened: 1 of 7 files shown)


config/config.yaml
```diff
@@ -44,7 +44,7 @@ queryConfig:
     serverAddress: ${KALDB_QUERY_SERVER_ADDRESS:-localhost}
     requestTimeoutMs: ${KALDB_QUERY_REQUEST_TIMEOUT_MS:-5000}
   defaultQueryTimeoutMs: ${KALDB_QUERY_DEFAULT_QUERY_TIMEOUT_MS:-3000}
-  managerConnectString: ${KALDB_MANAGER_CONNECTION_STRING:-localhost:8083}
+  managerConnectString: ${KALDB_MANAGER_CONNECTION_STRING:-gproto+http://localhost:8083}
```
kx-chen, line 47: This change is required as `GrpcClients.builder` requires the protocol 

https://github.com/slackhq/kaldb/pull/355/files#diff-44afaa56545846a976a196d21248f374ac4d50d4338ce07d5bce1ab9e63a4313R124

bryanlb:
Reviewed this and discussed with @vthacker today. 

Hydrating automatically on query was originally intended to be a short-term solution to demonstrate the functionality. The desired outcome is to be able to notify the user that some data is unavailable and allow them to issue a followup request for the data. Automatic hydration carries a risk in that a user creating a visualization or alert could target too long of a time window, inadvertently causing constant hydration requests.

We'll close this PR but keep the branch, so we can reference the approach for calculating the missing snapshots.

PR #356, opened 2022-08-18
Add `isRestored` to `ReplicaMetadata`

kx-chen:
Add a `isRestored` value to `ReplicaMetadata` in order to determine which Replicas are created from the restore API


vthacker:
I'd like us to create a new unit test in`ReplicaMetadataSerializerTest` where we create a object with isRestore=true , then serialize it to JSON. then when we go to deseriaize it ensure it works? Also a test with isRestore=false

I think it might fail because ReplicaMetadataSerializer needs to account for this new variable?

mansu:
Continuing this work in https://github.com/slackhq/kaldb/pull/357

PR #357, opened 2022-08-23
Add a `isRestored` field to `ReplicaMetadata`

mansu:


kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
```diff
@@ -148,6 +148,7 @@ private void restoreQueuedSnapshots() {
       }
     }
     restoreReplicasTimer.stop(restoreTimer);
+    LOG.info("Restored {} snapshots.", snapshotsToRestore.size());
```
vthacker, line 151: Are we done restoring when we look at this log statement? From my understanding we just create a metadatastore entry to restore , but the actual restoring is done by another service which is async. 

I'm unsure if the log line captures the correct state?

vthacker:
LGTM otherwise. Lemme know if you think we should change the new log entry and then we can commit this

kaldb/src/main/java/com/slack/kaldb/clusterManager/ReplicaRestoreService.java
mansu: Once we create the entry in the metadata store for the replica entry we are done restoring the snapshot. The async process will assign that replica to a cache node, just like a regular node. So, this statement is accurate.

vthacker (approved):

PR #358, opened 2022-08-23
Calculate approximate disk size based on latest chunk

vthacker:
In https://github.com/slackhq/kaldb/pull/349/ , we moved to calculating index size based on the entire directory. What happens is that after 1 chunk is written the approximate size calculation takes that into account since the calcuation is on the directory.

This PR changes the calculation to be for the last active chunk.


kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -343,7 +348,7 @@ protected void startUp() throws Exception {
     stopIngestion = false;
 
     directorySizeExecutorService.scheduleAtFixedRate(
-        () -> approximateDirectoryBytes = FileUtils.sizeOf(dataDirectory),
+        () -> approximateDirectoryBytes = FileUtils.sizeOf(activeChunkDirectory),
```
mansu, line 351: Please add a unit test to test this change. Also, add a unit test to ensure that the current bug is fixed. 

The gauge still needs to be reset on a roll over otherwise I don't think the bug is fixed.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -277,6 +279,9 @@ private ReadWriteChunk<T> getOrCreateActiveChunk(
               LuceneIndexStoreImpl.makeLogStore(
                   dataDirectory, indexerConfig.getLuceneConfig(), meterRegistry);
 
+      // this returns the latest active chunk on which we will calculate disk size
+      activeChunkDirectory = logStore.getDirectory().toFile();
```
mansu, line 283: Also, let's move this logic into the chunk roll over strategy since we don't need that info for indexing logic.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
vthacker: By gauge you mean `long approximateDirectoryBytes` ? It doesn't hurt to reset it, but strictly speaking I don't think it's needed? i.e if you see where we calculate it the reset will happen automatically on the next document add.

I will address it anyways

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkRollOverStrategyImpl.java
```diff
@@ -1,43 +0,0 @@
-package com.slack.kaldb.chunkManager;
```
mansu, line 1: Let's keep the existing strategy as is. Let's make a new roll over strategy and use it in the code. The roll over strategy code is already configurable.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkRollOverStrategy.java
```diff
@@ -1,6 +1,14 @@
...
 
+import java.io.File;
+
 // TODO: Rename ChunkRollOverPolicyPredicate
 public interface ChunkRollOverStrategy {
   boolean shouldRollOver(long currentBytesIndexed, long currentMessagesIndexed);
+
+  public void setActiveChunkDirectory(File activeChunkDirectory);
+
+  public long getApproximateDirectoryBytes();
```
mansu, line 11: Let's remove this method from the interface. I don't think we need to expose this as a guage in the IndexingChunkManager. If needed, we can expose the same info as a metric in the roll over strategy impl class.

kaldb/src/main/java/com/slack/kaldb/chunkManager/SizeOrMessageBasedRolloverStrategy.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.File;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.io.FileUtils;
+
+/**
+ * This class implements a rolls over a chunk when the chunk reaches a specific size or when a chunk
```
mansu, line 16: Class doc should be updated for the new class.

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
```diff
@@ -1084,7 +1084,7 @@ public void testMultipleByteRollOversSuccessfully()
     final long msgsPerChunk = 3L;
     final long maxBytesPerChunk = 100L;
     final ChunkRollOverStrategy chunkRollOverStrategy =
-        new ChunkRollOverStrategyImpl(maxBytesPerChunk, msgsPerChunk);
+        new SizeOrMessageBasedRolloverStrategy(maxBytesPerChunk, msgsPerChunk);
```
mansu, line 1087: Also, can you please add a new unit test for the current failure condition. I think creating a chunk with a lot of messages and not waiting for rollover synchronously  would recreate the failure.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/SizeOrMessageBasedRolloverStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class implements a rolls over a chunk when the chunk reaches a specific size or when a chunk
+ * hits a max message limit.
+ *
+ * <p>TODO: Also consider rolling over a chunk based on the time range for messages in the chunk.
+ */
+public class SizeOrMessageBasedRolloverStrategy implements ChunkRollOverStrategy {
```
vthacker, line 17: no changes from the old impl.

kaldb/src/main/java/com/slack/kaldb/chunkManager/DiskOrMessageCountBasedRolloverStrategy.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a disk based rollover strategy that calculates disk size every 10 seconds The disk size
+ * being calculated in the lucene index at that point in time. In addition, if we hit the max messages limit we also rollover
+ */
+public class DiskOrMessageCountBasedRolloverStrategy implements ChunkRollOverStrategy {
```
vthacker, line 22: would have liked this to be disk based only.

However since this is the default strategy, tests like `KaldbTest` which want to add exactly N message and perform a rollover is tough to fix with just a disk based rollover strategy

In practice, we'll keep the maxMessagesPerChunk high like 2B or something

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkRollOverStrategy.java
```diff
@@ -1,6 +1,12 @@
 package com.slack.kaldb.chunkManager;
 
+import java.io.File;
+
 // TODO: Rename ChunkRollOverPolicyPredicate
```
mansu, line 5: Remove this TODO.  Add a class doc for this interface.

Also, please add a new TODO to state that in future `ChunkRollOverStrategy` should take a chunk as an input and get statistics like message count, size etc.. from the chunk.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkRollOverStrategy.java
```diff
@@ -1,6 +1,12 @@
 package com.slack.kaldb.chunkManager;
```
mansu, line 1: Please move chunk roll over strategies to their own package like chunkrollover in the kaldb folder.

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/writer/LogMessageWriterImplTest.java
```diff
@@ -519,7 +519,7 @@ public void testAvgMessageSizeCalculationOnSpanIngestion() throws Exception {
 
     // even though we may have over 1k bytes, we only update the bytes ingested every 10s
     // as such this will only have 1 chunk at the time of query
-    assertThat(chunkManager.getChunkList().size()).isEqualTo(1);
+    assertThat(chunkManager.getChunkList().size()).isEqualTo(2);
```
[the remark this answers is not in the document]
vthacker: the underlying rollover strategy was different. Updated it to be same as before so that we don't need to change unit test and avoid confusion.

kaldb/src/main/java/com/slack/kaldb/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+        DIRECTORY_SIZE_EXECUTOR_PERIOD_MS,
+        TimeUnit.MILLISECONDS);
+  }
+
+  @Override
+  public boolean shouldRollOver(long currentBytesIndexed, long currentMessagesIndexed) {
+    liveBytesDirGauge.set(approximateDirectoryBytes);
+    boolean shouldRollover =
+        (approximateDirectoryBytes >= maxBytesPerChunk)
+            || (currentMessagesIndexed >= maxMessagesPerChunk);
```
mansu, line 73: One last thing, can you also please add some unit tests for message count based roll over for this strategy. Sorry, missed it earlier. I think we can copy over the test from the `ChunkRollOverStatrategyImplTest` for this use case.

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/chunkrollover/DiskOrMessageCountBasedRolloverStrategyTest.java
```diff
@@ -0,0 +1,219 @@
...
+    assertThat(chunkRollOverStrategy.shouldRollOver(-1, -1)).isFalse();
+    assertThat(chunkRollOverStrategy.shouldRollOver(0, 0)).isFalse();
+    assertThat(chunkRollOverStrategy.shouldRollOver(100, 100)).isFalse();
+    assertThat(chunkRollOverStrategy.shouldRollOver(1000, 1)).isFalse();
+    assertThat(chunkRollOverStrategy.shouldRollOver(1001, 1)).isFalse();
+
+    assertThat(chunkRollOverStrategy.shouldRollOver(100, 2000)).isTrue();
+    assertThat(chunkRollOverStrategy.shouldRollOver(100, 2001)).isTrue();
+    assertThat(chunkRollOverStrategy.shouldRollOver(1001, 2001)).isTrue();
+  }
```
mansu, line 218: This looks great! Please add a complete test using message count rollover like the unit test `testDiskBasedRolloverWithMultipleChunks`.

kaldb/src/test/java/com/slack/kaldb/chunkrollover/MessageSizeOrCountBasedRolloverStrategyTest.java
```diff
@@ -1,37 +1,55 @@
...
     assertThat(indexerCfg.getMaxBytesPerChunk()).isEqualTo(10737418240L);
-    ChunkRollOverStrategyImpl chunkRollOverStrategy =
-        ChunkRollOverStrategyImpl.fromConfig(indexerCfg);
+    MessageSizeOrCountBasedRolloverStrategy chunkRollOverStrategy =
+        MessageSizeOrCountBasedRolloverStrategy.fromConfig(metricsRegistry, indexerCfg);
     assertThat(chunkRollOverStrategy.getMaxBytesPerChunk()).isEqualTo(10737418240L);
     assertThat(chunkRollOverStrategy.getMaxMessagesPerChunk()).isEqualTo(100);
   }
 
   @Test(expected = IllegalArgumentException.class)
```
mansu, line 39: Please also add the illegal value tests to unit tests above.

mansu (changes requested):

mansu (approved):

PR #359, opened 2022-08-26
Pass in a list of chunks instead of a chunk id in SearchQuery.

mansu:
Pass in list of chunks instead of a single chunk id in the `SearchQuery` so we can search multiple chunks in a single call. 

Also, update the search logic to only search the given chunks when chunk ids are specified. Added additional unit test for this code. Also, updated the unit test to query via the grpc query layer so we can validate end to end.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
@@ -1,7 +1,6 @@
 package com.slack.kaldb.logstore.search;
 
 import java.util.List;
-import org.apache.logging.log4j.util.Strings;
 
 /** A class that represents a search query internally to LogStore. */
 public class SearchQuery {
@@ -50,7 +49,7 @@ public class SearchQuery {
         + ", bucketCount="
         + bucketCount
         + ", chunkIds="
-        + Strings.join(chunkIds, ',')
+        + String.join(",", chunkIds)
         + '}';
   }
 }


kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -7,7 +7,7 @@ option java_package = "com.slack.kaldb.proto.service";
 message SearchRequest {
   // Data sets or chunk_ids to be searched
   string dataset = 1;
-  string chunk_id = 2;
+  repeated string chunk_ids = 2;
```
vthacker, line 10: maybe call it `_chunk_ids` ? This is an internal only param I guess

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
```diff
@@ -43,6 +48,8 @@ public String toString() {
         + howMany
         + ", bucketCount="
         + bucketCount
+        + ", chunkIds="
+        + chunkIds
```
vthacker, line 52: does this work? i.e chunkIds is a List and just saying `chunkIds` here will it override correctly?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
```diff
@@ -10,20 +12,23 @@ public class SearchQuery {
...
 
   public SearchQuery(
       String dataset,
       String queryStr,
       long startTimeEpochMs,
       long endTimeEpochMs,
       int howMany,
-      int bucketCount) {
+      int bucketCount,
+      List<String> chunkIds) {
```
vthacker, line 24: I propose we add another ctor which takes the additional `chunkIds` . That will make the PR significantly smaller

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
```diff
@@ -316,24 +334,49 @@ public void testAddMessage() throws Exception {
 
   private void testChunkManagerSearch(
       ChunkManager<LogMessage> chunkManager,
+      List<String> chunkList,
```
vthacker, line 337: I think this should be called `chunkIds` ?

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
```diff
@@ -316,24 +334,49 @@ public void testAddMessage() throws Exception {
...
+    KaldbLocalQueryService<LogMessage> kaldbLocalQueryService =
+        new KaldbLocalQueryService<>(chunkManager, Duration.ofSeconds(3));
+    KaldbSearch.SearchRequest.Builder searchRequestBuilder = KaldbSearch.SearchRequest.newBuilder();
+    KaldbSearch.SearchResult response =
+        kaldbLocalQueryService.doSearch(
+            searchRequestBuilder
+                .setDataset(MessageUtil.TEST_DATASET_NAME)
+                .setQueryString(searchString)
+                .setStartTimeEpochMs(0)
+                .setEndTimeEpochMs(
```
vthacker, line 352: any specific reason to add one extra indirection to just using ` Long.MAX_VALUE` ?

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
```diff
@@ -316,24 +334,49 @@ public void testAddMessage() throws Exception {
...
+                    com.slack.kaldb.testlib.TemporaryLogStoreAndSearcherRule.MAX_TIME)
+                .setHowMany(10)
+                .setBucketCount(1000)
+                .addAllChunkIds(chunkList)
+                .build());
+
+    assertThat(response.getHitsList().size()).isEqualTo(expectedHitCount);
+    assertThat(response.getTotalSnapshots()).isEqualTo(totalSnapshots);
```
vthacker, line 360: should there be a few asserts around the relationship b/w totalSnapshots / expectedSnapshotsWithReplicas / chunkList.size() ?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
mansu: Leaving it as is since it's better to be explicit in the call params.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
mansu: Done.

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
mansu: Done.

kaldb/src/test/java/com/slack/kaldb/chunkManager/IndexingChunkManagerTest.java
mansu: Replaced it with `Long.MAX_VALUE`. I think we used MAX_TIME to better represent what the value was.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
```diff
@@ -49,7 +50,7 @@ public String toString() {
         + ", bucketCount="
         + bucketCount
         + ", chunkIds="
-        + chunkIds
+        + Strings.join(chunkIds, ',')
```
vthacker, line 53: I'd prefer if we don't use a log4j library for this given there exists one in the standard JDK

https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html

mansu: Good catch. Dang auto-complete. That's what I thought I used. Fixed now.

vthacker (approved):

PR #364, opened 2022-09-15
Distributed search to pass chunk ids to individual nodes

vthacker:
followup from https://github.com/slackhq/kaldb/pull/359 , where we make use of the local search to accept chunk ids and only search amongst those.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
@@ -195,7 +195,7 @@ public class KaldbDistributedQueryService extends KaldbQueryServiceBase {
   protected static Map<String, List<SearchMetadata>> getMatchingSearchMetadata(
       SearchMetadataStore searchMetadataStore, Map<String, SnapshotMetadata> snapshotsToSearch) {
     // iterate every search metadata whose snapshot needs to be searched.
-    // if there are multiple search metadata nodes then pck the most on based on
+    // if there are multiple search metadata nodes then pick the most on based on
     // pickSearchNodeToQuery
     ScopedSpan getMatchingSearchMetadataSpan =
         Tracing.currentTracer()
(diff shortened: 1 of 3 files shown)


kaldb/src/test/java/com/slack/kaldb/server/KaldbTest.java
```diff
@@ -352,7 +426,8 @@ public void testTwoIndexersAndOneQueryService() throws Exception {
             KALDB_TEST_CLIENT_2,
             ZK_PATH_PREFIX,
             2,
-            startTime2);
+            startTime2,
+            indexer2MeterRegistry);
     indexer2.serviceManager.awaitHealthy(DEFAULT_START_STOP_DURATION);
 
     KaldbSearch.SearchResult indexerSearchResponse =
```
mansu, line 433: I think it would be good to have 2 more tests when the same chunk is on multiple nodes:
* The same chunk is on indexer and cache node.
* The same chunk is on multiple cache nodes.

vthacker: I think these tests exist on KaldbDistributedQueryServiceTest.java

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -307,28 +323,31 @@ private List<SearchResult<LogMessage>> distributedSearch(KaldbSearch.SearchReque
...
-    for (String searchNodeUrl : nodes) {
-      queryStubs.add(getStub(searchNodeUrl));
-    }
-
-    span.tag("queryServerCount", String.valueOf(queryStubs.size()));
+            distribSearchReq.getStartTimeEpochMs(),
+            distribSearchReq.getEndTimeEpochMs(),
+            distribSearchReq.getDataset());
+
+    Map<String, List<SearchMetadata>> searchMetadataToQuery =
```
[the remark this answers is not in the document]
vthacker: thoughts on what the name should be? `searchMetadataToQuery` sounds reasonable to me

mansu (approved):

PR #366, opened 2022-09-27
Add an indexType field to `SnapshotMetadata` and `ReplicaMetadata`

mansu:
Add an indexType field to `SnapshotMetadata` and `ReplicaMetadata` to encode the type of index used in replica and snapshots.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ChunkInfo.java
@@ -42,7 +42,7 @@ public class ChunkInfo {
         chunkInfo.getDataEndTimeEpochMs(),
         chunkInfo.maxOffset,
         chunkInfo.kafkaPartitionId,
-        Metadata.IndexType.LUCENE_REGULAR);
+        Metadata.IndexType.LOGS_LUCENE9);
   }
 
   /* A unique identifier for a the chunk. */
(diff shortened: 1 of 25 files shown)


vthacker:
What's the motivation for this PR?

mansu:
When we use NRTSearch, we want to store it as a different type of index if needed. This change will also lay the ground work for supporting different chunk types in the future.

mansu:
> What's the motivation for this PR?

When we use NRTSearch, we want to store it as a different type of index if needed. This change will also lay the ground work for supporting different chunk types in the future.

vthacker:
> > What's the motivation for this PR?
> 
> When we use NRTSearch, we want to store it as a different type of index if needed. This change will also lay the ground work for supporting different chunk types in the future.

If we do this PR, I'd like us not to use `LUCENE_REGULAR` and add the version of lucene being used to create it.

Secondly, what's the plan to match the index type version when we download an index to the cache node?

mansu:
> > > What's the motivation for this PR?
> > 
> > 
> > When we use NRTSearch, we want to store it as a different type of index if needed. This change will also lay the ground work for supporting different chunk types in the future.
> 
> If we do this PR, I'd like us not to use `LUCENE_REGULAR` and add the version of lucene being used to create it.
> 
Changed it to `LOGS_LUCENE9`
> Secondly, what's the plan to match the index type version when we download an index to the cache node?
That's the next step. Once the replica metadata has these fields, we can match the cache nodes with the snapshot type during assignment.

PR #368, opened 2022-09-29
Improve unit test stability

vthacker:
- KaldbIndexerTest should wait till the documents have been indexed ( we make changes to the await and more details in the comments )
- Close tests properly i.e do null checks so that all resources close. Otherwise on a test failure , all subsequent tests fail with `java.util.concurrent.ExecutionException: io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use`

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
+++ b/kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
@@ -1,6 +1,5 @@
 package com.slack.kaldb.testlib;
 
-import com.slack.kaldb.server.Kaldb;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.search.MeterNotFoundException;
 import org.slf4j.Logger;
--- a/kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
+++ b/kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
@@ -32,10 +32,10 @@ public class MetricsUtil {
     try {
       return metricsRegistry.get(counterName).counter().count();
     } catch (MeterNotFoundException e) {
-      LOG.error("Metric not found", e);
+      LOG.warn("Metric not found", e);
       return 0;
     } catch (Exception e) {
-      LOG.error("Error while getting metric", e);
+      LOG.warn("Error while getting metric", e);
       throw e;
     }
   }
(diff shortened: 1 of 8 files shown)


kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
```diff
@@ -1,17 +1,50 @@
...
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.search.MeterNotFoundException;
 
 public class MetricsUtil {
+
   public static double getCount(String counterName, MeterRegistry metricsRegistry) {
-    return metricsRegistry.get(counterName).counter().count();
+    try {
+      return metricsRegistry.get(counterName).counter().count();
+    } catch (MeterNotFoundException e) {
```
mansu, line 11: Log the MeterNotFoundException and return 0 for all these functions. 

Also. log other exceptions and re-throw so we can easily see error in log in the future.

kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
```diff
@@ -1,17 +1,50 @@
...
+      // most likely we'll be calling getCount from a await() and waiting for the counter to match
```
mansu, line 12: Since this comment applies to entire class, please move it to class doc.

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
```diff
@@ -1,17 +1,66 @@
...
 public class MetricsUtil {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MetricsUtil.class);
+
   public static double getCount(String counterName, MeterRegistry metricsRegistry) {
-    return metricsRegistry.get(counterName).counter().count();
+    try {
+      return metricsRegistry.get(counterName).counter().count();
+    } catch (MeterNotFoundException e) {
+      LOG.error("Metric not found", e);
```
mansu, line 35: This can be an info/warn saying we the meter is not found so returning 0. This is not an error since we continue the method execution.

Does the exception have the meter name? If not, can we also log it here?

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/testlib/MetricsUtil.java
vthacker: changed to warn. Meter name is printed. additionally the stack trace usually makes it very evident which test line caused this and also infer the metric being referenced.

mansu (approved):

PR #194, opened 2022-01-10
upgrade to lucene 9

vthacker:
Lucene 9 is the next major version that released over a month ago. 

Right now our usage of lucene API is minimal so upgrading is much easier. Before we start working on more search improvements and use more of lucene APIs let's upgrade first 

complete list of changes are here - https://lucene.apache.org/core/9_0_0/changes/Changes.html

PR #372, opened 2022-10-04
Upgrade to use Lucene 9.3

mansu:
Upgrade to use Lucene 9.3 before we make further changes to schemas.

This update is based on @vthacker's PR: https://github.com/slackhq/kaldb/pull/194


kaldb/pom.xml
```diff
@@ -21,7 +21,7 @@
         <kafka.version>3.2.0</kafka.version>
         <jackson.version>2.13.3</jackson.version>
         <jackson.databind.version>2.12.7</jackson.databind.version>
-        <lucene.version>8.4.1</lucene.version>
+        <lucene.version>9.0.0</lucene.version>
```
vthacker, line 24: I would pick https://central.sonatype.dev/artifact/org.apache.lucene/lucene-core/9.4.0 which is now the latest release

vthacker (changes requested):

kaldb/pom.xml
mansu: 9.4 is a bit too new since it's only a few days old. So, upgraded to Lucene 9.3 for now.

vthacker (approved):

PR #376, opened 2022-10-11
Add SchemaAwareLogDocumentBuilder to Kaldb.

mansu:
In KalDB, the logs are ingested by a LogDocumentBuilder. However, the current document builder enforces types on known fields but any fields whose types are not pre-defined are ingested as Strings. This field conflict handling policy is not ideal for 2 reasons: (a) on dynamically ingested numeric fields, we can't do any range or numeric queries and (b) storing numbers as strings is also less efficient.

So, the `SchemaAwareLogDocumentBuilder` is a new LogDocumentBuilder that ingests the documents in a schema aware manner. It can take a fixed schema as before. In addition, for all dynamic fields, it also preserves the type of the field. If there is a field conflict, there is a field conflict policy that is applied to resolve the error. Currently, we have implemented 4 ways to handle field conflicts and also added unit tests to test the same. Added new tests for this functionality.

The overall work for integrating this schema document builder is being done in this PR: https://github.com/slackhq/kaldb/pull/369. I separated the code into a smaller diff for easier review.


kaldb/src/main/java/com/slack/kaldb/logstore/SchemaAwareLogDocumentBuilderImpl.java
```diff
@@ -0,0 +1,553 @@
...
+        }
+      }
+
+      // Long type
+      if (fromType == FieldType.LONG) {
+        if (toType == FieldType.TEXT) {
+          return ((Long) value).toString();
+        }
+        if (toType == FieldType.INTEGER) {
+          return ((Long) value).intValue();
```
danhermann, line 174: I'm still looking over this and not sure I have all of the context, yet, but a couple comments as I read through it. It seems like this would silently truncate data in the case of narrowing conversions such as `long` -> `int` where the value exceeds `Integer.MAX_VALUE`, `double` -> `float`, etc. Also, I didn't see handling of conversions that would result in runtime exceptions such as `NumberFormatException` if there was an attempt to convert an invalid string to a numeric value.

kaldb/src/test/java/com/slack/kaldb/logstore/SchemaAwareLogDocumentBuilderImplTest.java
```diff
@@ -0,0 +1,737 @@
+package com.slack.kaldb.logstore;
+
+import static com.slack.kaldb.logstore.SchemaAwareLogDocumentBuilderImpl.*;
```
danhermann, line 3: Minor: after merging, this wildcard import will trigger the new errorprone rule that disallows wildcard imports.

kaldb/src/main/java/com/slack/kaldb/logstore/SchemaAwareLogDocumentBuilderImpl.java
mansu: Loss of precision is expected. Handling exceptions is a good one. I will add more unit tests for it.

danhermann: It's not just a loss of precision but arithmetic overflow and such where the converted value is meaningless. I expect there will have to be some kind of handling for fields that cannot be converted to the target type when the `FieldConflictPolicy` is `CONVERT_FIELD_VALUE` or `CONVERT_AND_DUPLICATE_FIELD`, but that seems like something that could be worked out later.

danhermann (approved):
LGTM. Left some non-blocking suggestions on error handling and such.

kaldb/src/main/java/com/slack/kaldb/logstore/SchemaAwareLogDocumentBuilderImpl.java
mansu: Good callout on handling exceptions. Fixed up the code to handle exceptions and added more tests. Also, added tests to account for loss in precision and not errors. The converted values are meaningless, but leaving them as is for now. We can add an additional tag like converted fields etc.. to the document in a future iteration.

danhermann (dismissed):
LGTM. I think the format plugin is complaining about an extra LF at the end of `SchemaAwareLogDocumentBuilderImplTest`.

danhermann (approved):
LGTM pending successful CI

PR #378, opened 2022-10-12
Track number of query clauses 

vthacker:
Add span tag for clauses to see how long many boolean clauses does our query contain.

This stat can help track how many single term queries ( much much faster ) vs conjunctive/disjunctive queries there are

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
@@ -28,19 +28,10 @@ import org.apache.lucene.document.LongPoint;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.queryparser.classic.ParseException;
 import org.apache.lucene.queryparser.classic.QueryParser;
+import org.apache.lucene.search.*;
 import org.apache.lucene.search.BooleanClause.Occur;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -243,7 +234,9 @@ private Query buildQuery(
     if (queryStr.length() > 0) {
       queryBuilder.add(buildQueryParser().parse(queryStr), Occur.MUST);
     }
-    return queryBuilder.build();
+    BooleanQuery query = queryBuilder.build();
```
vthacker, line 237: Need to add span tag here and not parent method because we return `Query` , while all info is as a `BooleanQuery`

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -243,7 +234,9 @@ private Query buildQuery(
...
+    span.tag("lucene query clauses", Integer.toString(query.clauses().size()));
```
mansu, line 238: Use a single string name like "lucene_query_clause" instead.

mansu (changes requested):

mansu (approved):

PR #385, opened 2022-10-23
Rename `span.proto` to `zipkin.proto` to clarify intent.

mansu:


danhermann (approved):
LGTM pending successful CI or known flaky test

PR #388, opened 2022-10-26
Disambiguate dataset name vs service name

vthacker:
1. Rate limits apply to datasets
2. Datasets match with upstream topic names
3. If datasets specify service names then match those from the record

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
@@ -1,9 +1,8 @@
 package com.slack.kaldb.preprocessor;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.slack.kaldb.server.KaldbConfig.DEFAULT_START_STOP_DURATION;
-import static org.apache.curator.shaded.com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.AbstractService;
 import com.slack.kaldb.metadata.core.KaldbMetadata;
 import com.slack.kaldb.metadata.dataset.DatasetMetadata;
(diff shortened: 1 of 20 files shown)


kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -268,40 +262,29 @@ protected static List<Integer> getActivePartitionList(DatasetMetadata datasetMet
...
-      if (!datasetToPartitionList.containsKey(datasetName)) {
-        // this shouldn't happen, as we should have filtered all the missing datasets in the value
-        // mapper stage
-        throw new IllegalStateException(
-            String.format("Dataset '%s' was not found in dataset metadata", datasetName));
-      }
-
-      List<Integer> partitions = datasetToPartitionList.getOrDefault(datasetName, List.of());
+    return (topic, key, value, numPartitions) -> {
+      List<Integer> partitions = getActivePartitionList(datasetMetadata);
```
vthacker, line 273: this also has the advantage that we aren't deserializing the record for the 2nd time - perhaps improving efficiency

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/preprocessor/PreprocessorServiceUnitTest.java
```diff
@@ -121,9 +122,15 @@ public void shouldPreventInvalidStreamPropsConfig() {
   public void shouldReturnRandomPartitionFromStreamPartitioner() {
     String datasetName = "datasetName";
     List<Integer> partitionList = List.of(33, 44, 55);
-    Map<String, List<Integer>> datasetNameToPartitions = Map.of(datasetName, partitionList);
+    DatasetMetadata datasetMetadata =
```
mansu, line 125: Please add unit tests for the new use cases.

mansu (changes requested):

mansu (changes requested):
Other than the field name, this LGTM.

kaldb/src/main/proto/manager_api.proto
```diff
@@ -31,6 +31,8 @@ message CreateDatasetMetadataRequest {
   string name = 1;
   // Owner information, maybe be any string
   string owner = 2;
+  // The service name(s) that the dataset is indexing
```
bryanlb, line 34: This isn't a repeated string, so only one service name can be provided here. If this is supposed to be a regex/glob that could match multiple services, this probably should be named service_name_pattern.

kaldb/src/main/proto/metadata.proto
```diff
@@ -134,6 +134,9 @@ message DatasetMetadata {
 
   // List of partitions assigned to this service, and their effective times
   repeated DatasetPartitionMetadata partition_configs = 4;
+
+  // either same as name or a special value _all
+  string service_name = 5;
```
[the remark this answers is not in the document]
bryanlb: I think `service_name_pattern` is a clearer option as mentioned above. This also won't be stored as straight regex, as it will need to be escaped specific to the flavor in use.

bryanlb (changes requested):
I would suggest splitting this PR will make this change a bit easier to follow and merge, as there's currently two distinct features included in this PR.

The first portion is service/dataset change, and allows ingesting all messages on a given _dataset_ at a fixed capacity. This can be separate from the regex/pattern filtering support, which optionally allows dropping messages that make it past the rate-limiter.

bryanlb:
Closing as this was addressed in https://github.com/slackhq/kaldb/pull/405 and has already been merged! :shipit:

PR #390, opened 2022-10-28
PreprocessorService log statement

vthacker:
Add a log statement to confirm if we hit this illegal state exception when ingesting the traces dataset ie we should hit this when reading any data other than flannel-be since it's not part of the dataset metada.


danhermann (approved):
LGTM

mansu (approved):

PR #393, opened 2022-11-04
add exists code functionality

vthacker:
1. All fields detected as texty fields, we use StringField instead of TextField. If the log is splitting data into another field which is not a "message" field, it's computer generated data. So they probably want to search as-is. In addition they might want to do aggregations on those. Currently we default to text fields and neither of those two were possible.
2. Enable doc-values on boolean fields
3. Support term queries - The lucene query parser had no notion of the schema. So a numeric field term query wouldn't work previously. All term queries would get analyzed with the standard query parser so a string field depending on the term could also not match
4. Add exists query support


kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbQueryParserTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  public KaldbQueryParserTest() throws IOException {}
+
+  @BeforeClass
+  public static void initTests() {
+    Tracing.newBuilder().build();
+  }
+
+  @Test
+  public void testExistsQueryWithStrField() {
```
vthacker, line 33: today fails with this
```
java.lang.IllegalStateException: FieldExistsQuery requires that the field indexes doc values, norms or vectors, but field 'service_name' exists and indexes neither of these data structures

	at org.apache.lucene.search.FieldExistsQuery.rewrite(FieldExistsQuery.java:156)
	at org.apache.lucene.search.BooleanQuery.rewrite(BooleanQuery.java:285)
```

kaldb/src/main/java/com/slack/kaldb/logstore/LogDocumentBuilderImpl.java
```diff
@@ -48,7 +50,7 @@ static class PropertyDescription {
     final boolean isStored;
     final boolean isIndexed;
     final boolean isAnalyzed;
-    final boolean storeNumericDocValue;
+    final boolean addDocValues;
```
vthacker, line 53: if y'all have a better name for this I'd happily change it :)

kaldb/src/test/java/com/slack/kaldb/logstore/LuceneIndexStoreImplTest.java
```diff
@@ -223,8 +223,9 @@ public void indexLongUnbreakableField() {
       MessageUtil.addFieldToMessage(msg, "hugefield", hugeField);
       logStore.logStore.addMessage(msg);
       assertThat(getCount(MESSAGES_RECEIVED_COUNTER, logStore.metricsRegistry)).isEqualTo(1);
-      assertThat(getCount(MESSAGES_FAILED_COUNTER, logStore.metricsRegistry)).isEqualTo(0);
-      // Counters not set since no commit.
+      // UTF8 encoding is longer than the max length 32766
```
vthacker, line 226: since we now default to STRING instead of TEXT ( which is in my opinion the right default for texty fields ) the entire high field is encoded as a single token and lucene has a limit. So the document actually errors out

kaldb/src/test/java/com/slack/kaldb/logstore/schema/ConvertFieldValueAndDuplicateTest.java
```diff
@@ -68,7 +68,7 @@ public void testListTypeInDocument() throws IOException {
                     Map.of("leaf2", "value2", "leaf21", 3, "nestedList", List.of(1)))));
 
     Document testDocument = docBuilder.fromMessage(message);
-    assertThat(testDocument.getFields().size()).isEqualTo(18);
+    assertThat(testDocument.getFields().size()).isEqualTo(23);
```
vthacker, line 71: All changes in this class are also related to the fact that now we default to STRING field. For String field we add a doc-value field as well. So all counts in this test need to be updated

kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldType.java
```diff
@@ -35,6 +43,16 @@ public void addField(Document doc, String name, Object value, LuceneFieldDef fie
         // Since a text field is tokenized, we don't need to add doc values to it.
       }
     }
+
+    @Override
+    public Query termQuery(String field, String queryText, Analyzer analyzer) {
+      throw new NotImplementedException("text fields parsing is currently in KaldbQueryParser");
```
bryanlb, line 49: Looks like we missed a word or two for this error message

kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldTypeUtils.java
```diff
@@ -0,0 +1,10 @@
+package com.slack.kaldb.metadata.schema;
+
+import org.apache.lucene.analysis.core.KeywordAnalyzer;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+
+public class FieldTypeUtils {
```
bryanlb, line 6: Not sure if we need this separate class just yet - if we do though, we should have comments explaining the purpose of this "utils" class, since this is a good chance for it to become a catch-all.

kaldb/src/test/java/com/slack/kaldb/logstore/search/queryparser/KaldbQueryParserTest.java
```diff
@@ -15,4 +41,221 @@ public void testInit() {
...
+    withStringField("my_service_name");
+
+    // indexed=true analyzed=true - Use ReservedField
+    withTextField(LogMessage.ReservedField.USERNAME.fieldName);
+    withTextField(LogMessage.ReservedField.MESSAGE.fieldName);
+    // All texty fields that are not reserved will use analyzed=false aka string field which we
+    // already tested
+
+    // indexed=true storeDocValues=true - Use ReservedField
+    withLongField(LogMessage.ReservedField.DURATION_MS.fieldName);
```
bryanlb, line 59: I would consider how we can make these tests use less method calls, and encapsulate as much of the logic within the actual tests - even at the expense of duplicating code. When the asserts are directly in the test method, it makes troubleshooting failing tests much quicker in the future.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldTypeUtils.java
```diff
@@ -0,0 +1,4 @@
+package com.slack.kaldb.metadata.schema;
+
+/** This class has utility methods that work with the FieldType enum */
+public class FieldTypeUtils {}
```
bryanlb, line 4: This can be deleted now?

vthacker: done!

PR #395, opened 2022-11-07
Add support for aggregations to search request/response classes

danhermann:
This PR adds an optional "aggs" field that provides supports for aggregations (["metrics" aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/search-aggregations-metrics.html) in the terminology of Elasticsearch/Opensearch) to the search request and response classes.

The structure of the search request is modeled after the ES/OS query DSL for ~~metrics aggregations~~ (edit: this was updated by https://github.com/slackhq/kaldb/pull/395/commits/96c78068c751219fe5fcbecbcc1dc1bbcf8074c1 to a bucket aggregation with auto-interval date histogram):
```
GET /my-index-000001/_search
{
  "aggs": {
    "my-first-agg-name": {
      "min": {
        "field": "my-field"
      }
    },
    "my-second-agg-name": {
      "avg": {
        "field": "my-other-field"
      }
    }
  }
}
```

For the sake of simplicity, the request structure is designed to handle only the aggregations that are currently planned for KalDB. These include `sum`, `min`, `max`, `avg`, `value_count`, and `percentile_ranks`.

We may want to refactor the top-level request after this work is completed. Some of this refactoring was done in https://github.com/slackhq/kaldb/pull/395/commits/96c78068c751219fe5fcbecbcc1dc1bbcf8074c1.

### DIFF
--- a/kaldb/src/main/proto/kaldb_search.proto
+++ b/kaldb/src/main/proto/kaldb_search.proto
@@ -15,6 +15,9 @@ message SearchRequest {
   int64 end_time_epoch_ms = 5;
   int32 how_many = 6;
   int32 bucket_count = 7;
+
+  // Aggregations
+  optional Aggregations aggs = 8;
 }
 
 message SearchResult {
@@ -33,6 +36,17 @@ message HistogramBucket {
   double low = 1;
   double high = 2;
   double count = 3;
+}
+
+message Aggregations {
+  repeated Aggregation agg = 1;
+}
+
+message Aggregation {
+  string agg_name = 1;
+  string agg_type = 2;
+  string agg_field = 3;
+  repeated double values = 4;
 }
 
 service KaldbService {
--- a/kaldb/src/main/proto/kaldb_search.proto
+++ b/kaldb/src/main/proto/kaldb_search.proto
@@ -17,7 +17,7 @@ message SearchRequest {
   int32 bucket_count = 7;
 
   // Aggregations
-  optional Aggregations aggs = 8;
+  repeated AggregationRequest aggs = 8;
 }
 
 message SearchResult {
@@ -30,6 +30,8 @@ message SearchResult {
   int32 total_nodes = 7;
   int32 total_snapshots = 8;
   int32 snapshots_with_replicas = 9;
+
+  repeated AggregationResult aggregations = 10;
 }
 
 message HistogramBucket {
@@ -38,15 +40,18 @@ message HistogramBucket {
   double count = 3;
 }
 
-message Aggregations {
-  repeated Aggregation agg = 1;
-}
-
-message Aggregation {
+message AggregationRequest {
   string agg_name = 1;
   AggregationType agg_type = 2;
   string agg_field = 3;
   repeated double values = 4;
+}
+
+message AggregationResult {
+  string agg_name = 1;
+  AggregationType agg_type = 2;
+  string agg_field = 3;
+  double value = 4;
 }
 
 enum AggregationType {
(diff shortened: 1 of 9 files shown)


kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -35,6 +38,17 @@ message HistogramBucket {
   double count = 3;
 }
 
+message Aggregations {
+  repeated Aggregation agg = 1;
+}
+
+message Aggregation {
+  string agg_name = 1;
+  string agg_type = 2;
```
danhermann, line 47: Must be one of `sum`, `min`, `max`, `avg`, `value_count`, or `percentile_ranks`.

kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -35,6 +38,17 @@ message HistogramBucket {
...
+  string agg_field = 3;
+  repeated double values = 4;
```
danhermann, line 49: Used only for `agg_type = 'percentile_ranks'`

danhermann:
As of commit https://github.com/slackhq/kaldb/pull/395/commits/96c78068c751219fe5fcbecbcc1dc1bbcf8074c1 above, the request/response format would support the rough equivalent of an ES aggregation with auto-interval date histogram bucketing. As an illustration, consider the following ES query annotated with the values from the KalDB's search request:

```
POST /[SearchRequest.dataset]/_search?size=0
{
  "query": {
    "bool": {
      "filter": [
        //SearchRequest.query_string goes here
      ]
    }
  }, 
  "aggs": {
    "your_aggregation_name_here": {
      "auto_date_histogram": {
        "field": "_timesinceepoch",
        "buckets": SearchRequest.bucket_count
      },
      "aggs": {
        "your_aggregation_name_here2": {
          "avg": {
            "field": "desired_field_here"
          }
        },
        "your_aggregation_name_here3": {
          "min": {
            "field": "another_field_here"
          }
        }
      }
    }
  }
}
```

The query above would apply the Lucene query, divide matching documents in the specified number of buckets and then perform the `avg` and `min` aggregations for each bucket.

One current limitation is that `HistogramBucket` supports only date-based buckets. We would need to generalize it to support term or range-based buckets.

kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -33,6 +36,31 @@ message HistogramBucket {
   double low = 1;
   double high = 2;
```
danhermann, line 37: One possibility to accommodate bucketing that is not date-based could be:
```
message HistogramBucket {
  oneof bucket_definition {
    DateBasedBucketDefinition date_bucket = 1;
    RangeBasedBucketDefinition range_bucket = 2;
    TermBasedBucketDefinition term_bucket = 3;
  }
  double count = 4;
  repeated AggregationResult aggregations = 5;
```

kaldb/src/main/proto/kaldb_search.proto
mansu: What if we use a generic dictionary in request and response instead do the agg_type can customize the values to be sent and received.

kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -33,6 +36,31 @@ message HistogramBucket {
...
   double count = 3;
+
+  repeated AggregationResult aggregations = 4;
```
mansu, line 40: Should this AggregationResult be in `SearchResult`?

mansu (changes requested):
It would be great to see how this diff works with the aggregations code to better understand it's use.

bryanlb:
Work-in-progress rebased and merged into branch [dhermann_aggs_request](https://github.com/slackhq/kaldb/tree/dhermann_aggs_request)

PR #396, opened 2022-11-07
Add parallel ES multisearch functionality

bryanlb:
When multiple queries are now submitted to the multisearch endpoint these will execute in parallel using the ForkJoinPool. Currently these execute sequentially, which can cause unexpected timeout behavior when submitting multiple queries in a single http request.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
@@ -31,6 +31,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


kaldb/src/test/java/com/slack/kaldb/server/ZipkinServiceTest.java
```diff
@@ -152,6 +153,7 @@ public static List<LogWireMessage> generateLogWireMessagesForOneTrace(
   }
 
   @Test
+  @Ignore // Flakey test, occasionally returns an empty result
```
mansu, line 156: Instead of ignoring it's best to run flaky tests under a flag and only run them locally or as a separate build step.

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -115,11 +113,15 @@ private EsSearchResponse doSearch(EsSearchRequest request) throws IOException {
...
-  private HitsMetadata getHits(KaldbSearch.SearchResult searchResult) throws IOException {
+  private HitsMetadata getHits(KaldbSearch.SearchResult searchResult) {
     List<ByteString> hitsByteList = searchResult.getHitsList().asByteStringList();
     List<SearchResponseHit> responseHits = new ArrayList<>(hitsByteList.size());
     for (ByteString bytes : hitsByteList) {
-      responseHits.add(SearchResponseHit.fromByteString(bytes));
+      try {
+        responseHits.add(SearchResponseHit.fromByteString(bytes));
+      } catch (IOException e) {
+        LOG.error("Error processing search hits bytes", e);
```
mansu, line 123: This api can fail all calls and we would still return a 200 error. It would be good catch this case using a metric. Do we have any metrics that provide visibility into these exceptions? If not, let's add a metric for successful and failed requests.

mansu (changes requested):

kaldb/src/test/java/com/slack/kaldb/server/ZipkinServiceTest.java
bryanlb: As discussed, we'll circle back on this as part of a separate effort to fix the test suites. This isn't flaky as part of this PR and has been having issues for a few weeks since its original introduction (see - https://github.com/slackhq/kaldb/actions/runs/3316587545/jobs/5478551341 for an example from a few weeks ago on `master`).

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
bryanlb: I reworked the error handling here to account for the previous errors with deserialization, as well as any other runtime exceptions that aren't currently handled. These now are properly handled, logged, and reported into the span as well as reported with a 500 status on the multisearch request that failed (along with the trace ID).

mansu (approved):

PR #398, opened 2022-11-07
Address forkjoin missing request context

bryanlb:
This removes the `RequestContext.current()` from wrapping from the direct executor. Since the direct executor is a caller-runs policy, this should not be needed.


mansu (dismissed):

bryanlb:
This isn't generating the trace data as I'm expecting, so I'm going to take a different approach on this. I'll reopen this if needed, but will likely use a different PR.

PR #399, opened 2022-11-08
Add cached threadpool for parallel multisearch queries

bryanlb:
This adds parallel query handling for multisearch, and also supports the appropriate trace instrumentation. I also included some additional helpful troubleshooting info around a potential timeout exception that I observed while working on this, and why it would be expected to happen.

This no longer attempts to use forkjoin, but uses a cached executor and futures like we do in the distributed query code as it's easier to instrument with our tracing and is consistent with our other code.

Screenshot 2022-11-08 at 11 13 30 AM

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
@@ -2,7 +2,11 @@ package com.slack.kaldb.elasticsearchApi;
 
 import brave.ScopedSpan;
 import brave.Tracing;
+import brave.propagation.TraceContext;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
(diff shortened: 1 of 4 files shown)


kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -47,6 +53,9 @@
 public class ElasticsearchApiService {
   private static final Logger LOG = LoggerFactory.getLogger(ElasticsearchApiService.class);
   private final KaldbQueryServiceBase searcher;
+  private final ExecutorService executorService =
```
mansu, line 56: Please add a comment on why we create our own pool for better tracing so it lives with the code.

mansu (changes requested):
Please add a comment and this change is good to go.

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
bryanlb: Done!

mansu (approved):

Issue #400, opened 2022-11-09
NPE when creating a recovery task on Indexer


mansu:
We often see NPEs when creating a recovery task which leads to indexer restarts.

> java.lang.NullPointerException 	at com.slack.kaldb.server.RecoveryTaskCreator.lambda$determineStartingOffset$5(RecoveryTaskCreator.java:154) 	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176) 	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655) 	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) 	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) 	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) 	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) 	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) 	at com.slack.kaldb.server.RecoveryTaskCreator.determineStartingOffset(RecoveryTaskCreator.java:155) 	at com.slack.kaldb.server.KaldbIndexer.indexerPreStart(KaldbIndexer.java:114) 	at com.slack.kaldb.server.KaldbIndexer.startUp(KaldbIndexer.java:81) 	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$2.run(AbstractExecutionThreadService.java:61) 	at com.google.common.util.concurrent.Callables.lambda$threadRenaming$3(Callables.java:103) 	at java.base/java.lang.Thread.run(Thread.java:829) |

PR #403, opened 2022-11-11
Skip null snapshots to avoid NPE.

mansu:
This fix will skip over any NullPointerExceptions and help recovery task creation progress further. 

Added logging to better debug the issue so we can fix the root cause: https://github.com/slackhq/kaldb/issues/400

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
+import org.apache.logging.log4j.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
```diff
@@ -147,11 +148,25 @@ public List<SnapshotMetadata> deleteStaleLiveSnapshots(List<SnapshotMetadata> sn
...
     List<SnapshotMetadata> snapshotsForPartition =
         snapshots
             .stream()
-            .filter(snapshotMetadata -> snapshotMetadata.partitionId.equals(partitionId))
+            .filter(
+                snapshotMetadata -> {
+                  if (snapshotMetadata == null || snapshotMetadata.partitionId == null) {
+                    LOG.warn(
+                        "snapshot metadata or partition id can't be null: "
+                            + Strings.join(snapshots, ','));
```
bryanlb, line 164: This should be `String.join` - then we also won't need the import statement.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
mansu: Done.

bryanlb (approved):

PR #405, opened 2022-11-14
Add service name to datasets and use that to apply rate filtering of documents

vthacker:
Introduce a service name concept within a dataset.

When we read a record from Kafka, today we match the dataset name with the records service name. This meant if we needed to ingest logs with multiple services we need to create N datasets which wasn't what we wanted

In this PR, we introduce a concept of a service name which can be `_all` meaning all services within the dataset will be ingested ( respecting the rate limits ) . One could also ingest only a service by mentioning it's name.

In the future we want to allow for glob matching as well.

This PR should also be back-compat meaning existing clusters will not need service name and just read the dataset name as the service name

This PR is a different approach to https://github.com/slackhq/kaldb/pull/388 and we can close that PR out once this is committed.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/dataset/DatasetMetadata.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/dataset/DatasetMetadata.java
@@ -16,6 +16,7 @@ import java.util.stream.Collectors;
 public class DatasetMetadata extends KaldbMetadata {
 
   public final String owner;
+  public final String serviceNamePattern;
   public final long throughputBytes;
   public final ImmutableList<DatasetPartitionMetadata> partitionConfigs;
 
@@ -23,7 +24,8 @@ public class DatasetMetadata extends KaldbMetadata {
       String name,
       String owner,
       long throughputBytes,
-      List<DatasetPartitionMetadata> partitionConfigs) {
+      List<DatasetPartitionMetadata> partitionConfigs,
+      String serviceNamePattern) {
     super(name);
     checkArgument(name.length() <= 256, "name must be no longer than 256 chars");
     checkArgument(name.matches("^[a-zA-Z0-9_-]*$"), "name must contain only [a-zA-Z0-9_-]");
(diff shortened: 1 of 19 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/dataset/DatasetMetadata.java
```diff
@@ -32,11 +34,22 @@ public DatasetMetadata(
     checkArgument(throughputBytes >= 0, "throughputBytes must be greater than or equal to 0");
     checkPartitions(partitionConfigs, "partitionConfigs must not overlap start and end times");
 
+    checkArgument(
+        serviceNamePattern.length() <= 256, "serviceNamePattern must be no longer than 256 chars");
+    checkArgument(
```
bryanlb, line 39: We probably don't need this check here - as long as it's not null / empty that would be a valid pattern; this will especially be the case if we move to glob/regex.

If this is indeed a regex pattern in the future, we probably would want to compile that pattern and store it in this class once, instead of at runtime.

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
```diff
@@ -102,6 +105,86 @@ protected static RateLimiter smoothBurstyRateLimiter(
...
+        meterRegistry
+            .counter(MESSAGES_DROPPED, getMeterTags("", MessageDropReason.MISSING_SERVICE_NAME))
+            .increment();
+        meterRegistry
+            .counter(BYTES_DROPPED, getMeterTags("", MessageDropReason.MISSING_SERVICE_NAME))
+            .increment(bytes);
+        return false;
+      }
+
+      for (RateLimiterPerDatasetHolder rateLimiterPerDataset : rateLimiterPerDatasetList) {
```
bryanlb, line 164: Why are we iterating through this tuple? Can't we just iterate through the list of datasets to find the matching one, and then get the matching rate limiter from a map, keyed against dataset? It seems like the tuple shouldn't be necessary?

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -261,32 +252,50 @@ protected static List<Integer> getActivePartitionList(DatasetMetadata datasetMet
         .collect(Collectors.toUnmodifiableList());
   }
 
+  public static List<DatasetMetadata> sortDatasetsOnThroughput(
```
bryanlb, line 255: Maybe a comment here, or where it's called explaining why we sort on throughput.

kaldb/src/main/proto/metadata.proto
```diff
@@ -134,6 +134,9 @@ message DatasetMetadata {
 
   // List of partitions assigned to this service, and their effective times
   repeated DatasetPartitionMetadata partition_configs = 4;
+
+  // either same as name or a special value _all
```
bryanlb, line 138: Or `*`

bryanlb (approved):
Just a couple of minor nits if you want to address them.

danhermann (approved):
LGTM

PR #406, opened 2022-11-15
Clarify existing logging and add some diagnostic logging

danhermann:


kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
```diff
@@ -244,7 +244,7 @@ boolean handleRecoveryTask(RecoveryTaskMetadata recoveryTaskMetadata) {
       chunkManager.awaitTerminated(DEFAULT_START_STOP_DURATION);
       return success;
     } catch (Exception ex) {
-      LOG.error("Encountered exception in recovery task: {}", recoveryTaskMetadata, ex);
```
danhermann, line 247: There were a couple of these entries in the logs without the underlying exception so this change should provide that info.

bryanlb (approved):

PR #419, opened 2022-11-21
Convert chunk manager to use listenable futures with cancellation

bryanlb:
Converts the chunk manager to use listenable futures, which can be interrupted, instead of completable futures which cannot. This also should address a long standing todo to return an exception when all results return an error and now returns the exception instead of an empty result.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
@@ -4,20 +4,25 @@ import brave.Tracing;
 import brave.propagation.CurrentTraceContext;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.AbstractIdleService;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.chunk.Chunk;
 import com.slack.kaldb.logstore.search.SearchQuery;
 import com.slack.kaldb.logstore.search.SearchResult;
 import com.slack.kaldb.logstore.search.SearchResultAggregator;
 import com.slack.kaldb.logstore.search.SearchResultAggregatorImpl;
-import com.spotify.futures.CompletableFutures;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
+import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
@@ -36,7 +41,15 @@ public abstract class ChunkManagerBase<T> extends AbstractIdleService implements
   // to the amount of reads, and it must be a threadsafe implementation
   protected final List<Chunk<T>> chunkList = new CopyOnWriteArrayList<>();
 
-  private static final ExecutorService queryExecutorService = queryThreadPool();
+  private static final ListeningExecutorService queryExecutorService = queryThreadPool();
+
+  private static final ScheduledExecutorService queryCancellationService =
+      Executors.newSingleThreadScheduledExecutor(
+          new ThreadFactoryBuilder()
+              .setNameFormat("chunk-manager-query-cancellation-%d")
+              .setUncaughtExceptionHandler(
+                  (t, e) -> LOG.error("Exception on thread {}: {}", t.getName(), e))
+              .build());
 
   /*
    * We want to provision the chunk query capacity such that we can almost saturate the CPU. In the event we allow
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -79,42 +92,67 @@ public SearchResult<T> query(SearchQuery query, Duration queryTimeout) {
               .collect(Collectors.toList());
     }
 
-    List<CompletableFuture<SearchResult<T>>> queries =
+    // Shuffle the chunks to query. The chunkList is ordered, meaning if you had multiple concurrent
```
vthacker, line 95: 🧠

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -36,7 +41,11 @@
   // to the amount of reads, and it must be a threadsafe implementation
   protected final List<Chunk<T>> chunkList = new CopyOnWriteArrayList<>();
 
-  private static final ExecutorService queryExecutorService = queryThreadPool();
+  private static final ListeningExecutorService queryExecutorService = queryThreadPool();
+
+  private static final ScheduledExecutorService queryCancellationService =
+      Executors.newSingleThreadScheduledExecutor(
+          new ThreadFactoryBuilder().setNameFormat("chunk-manager-query-cancellation-%d").build());
```
danhermann, line 48: Could you include an uncaught exception handler on this thread factory?

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -45,10 +54,14 @@
...
-        Runtime.getRuntime().availableProcessors(),
-        new ThreadFactoryBuilder().setNameFormat("chunk-manager-query-%d").build());
+  private static ListeningExecutorService queryThreadPool() {
+    // todo - consider making the thread count a config option; this would allow for more
+    //  fine-grained tuning, but we might not need to expose this to the user if we can set sensible
+    //  defaults
+    return MoreExecutors.listeningDecorator(
+        Executors.newFixedThreadPool(
+            Math.max(1, Runtime.getRuntime().availableProcessors() - 2),
+            new ThreadFactoryBuilder().setNameFormat("chunk-manager-query-%d").build()));
```
danhermann, line 64: Similar to the previous comment, I [just added](https://github.com/slackhq/kaldb/commit/8bf26d9e0034565331584340637ac7226949d05f#diff-e1190a6a806d660135b17a398bb14ba38c3554852e456cfa1812bc836e87ae7cR52) an uncaught exception handler here that we should retain.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -79,42 +92,67 @@ public SearchResult<T> query(SearchQuery query, Duration queryTimeout) {
...
-        CompletableFutures.allAsList(queries);
+                    queryExecutorService.submit(
+                        currentTraceContext.wrap(
+                            () -> {
+                              try {
+                                if (Thread.interrupted()) {
+                                  LOG.warn(
+                                      "Chunk query thread timed out without starting work, returning error result.");
+                                  return errorResult;
+                                }
```
mansu, line 115: Let's leave this TODO in since it will still be helpful to do.

                        // TODO: this will not cancel lucene query. Use ExitableDirectoryReader
                        //  in the future and pass this timeout

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -79,42 +92,67 @@ public SearchResult<T> query(SearchQuery query, Duration queryTimeout) {
...
+                                // represents a parse failure ) and instead of returning an empty
+                                // result we throw back an error to the user
+                                if (err instanceof IllegalArgumentException) {
+                                  throw err;
+                                }
+                                return errorResult;
+                              }
+                            })))
+            .peek(
+                (future) ->
```
mansu, line 133: Can we move this cancellation logic into the map? Does this have to be in peek?

mansu (changes requested):
Couple of minor clarifications needed. Otherwise LGTM!

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
bryanlb: This todo is incorrect which is why it was removed. We do indeed now cancel the lucene query by using interrupts - additionally, the `ExitableDirectoryReader` is not appropriate for this use case. 

The `ExitableDirectoryReader` would only be applicable when you want to shutdown the _reader_ after a given timeout https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/index/ExitableDirectoryReader.html. We don't want to shutdown the reader, as we reuse the same reader across multiple queries. We want to timeout the query instead, which is why this comment was removed.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
bryanlb: We need a reference to the returned future in order to schedule a cancellation. You can do this all in a single map, but it ended up being less readable as you need to expand the lambda body and use an intermediate variable. 

You can also do this with multiple map steps (ie, `.map().map().collect()`) but IntelliJ will flag this as unnecessary, as you do not need to mutate the future. You just need the reference in order to schedule the cancellation.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
bryanlb: Done.

mansu (approved):

PR #420, opened 2022-11-21
Refactor caching chunk manager to limit concurrent thread execution

bryanlb:
Changes the current caching chunk manager behavior to use a single shared executor pool with either one or two threads. This is in contrast to the existing code which instantiates a single thread executor for every cached chunk. This addresses performance issues when large amounts of cached chunks are attempted to be instantiated and compete for limited CPU cores.


kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
```diff
@@ -49,6 +53,14 @@ public CachingChunkManager(
...
     this.dataDirectoryPrefix = dataDirectoryPrefix;
     this.slotCountPerInstance = slotCountPerInstance;
+
+    // todo - consider making the thread count a config option; this would allow for more
+    //  fine-grained tuning, but we might not need to expose this to the user if we can set sensible
+    //  defaults
+    this.executorService =
+        Executors.newFixedThreadPool(
+            Runtime.getRuntime().availableProcessors() >= 4 ? 2 : 1,
+            new ThreadFactoryBuilder().setNameFormat("caching-chunk-manager-%d").build());
```
danhermann, line 63: It would be good to include an uncaught exception handler here, too.

mansu (approved):
LGTM.

kaldb/src/main/java/com/slack/kaldb/chunkManager/CachingChunkManager.java
bryanlb: Done.

danhermann (approved):
LGTM

PR #424, opened 2022-11-29
Don't fail recovery tasks that reference unavailable Kafka offsets

danhermann:


mansu (dismissed):
LGTM!

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -275,43 +275,42 @@ public boolean consumeMessagesBetweenOffsetsInParallel(
...
-                if (endOffsetInclusive >= 0 && record.offset() > endOffsetInclusive) {
-                  throw new IllegalArgumentException(
-                      "Record is after end offset range: " + endOffsetInclusive);
-                }
-                try {
-                  if (logMessageWriterImpl.insertRecord(record)) {
-                    recordsReceivedCounter.increment();
-                  } else {
-                    recordsFailedCounter.increment();
+                  LOG.warn("Record is before start offset range: " + startOffsetInclusive);
```
bryanlb, line 286: Couple things on these log messages - the first is they should be using the token replacement instead of string concat. Additionally, since these now skip the record I would make that clearer in the message (ie, "record is before start, will permanently drop this record").

I would also suggest making this a _debug_ instead of a _warn_, or rethink where this log message will sit. As it is currently this has the chance to easily generate millions or billions of log messages in a short amount of time depending on how far a cluster is behind. 

Another way to address this would be to check the first or last record of a batch and only log it once per batch.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -275,43 +275,42 @@ public boolean consumeMessagesBetweenOffsetsInParallel(
...
+                  recordsFailedCounter.increment();
+                } else if (endOffsetInclusive >= 0 && record.offset() > endOffsetInclusive) {
+                  LOG.warn("Record is after end offset range: " + endOffsetInclusive);
```
bryanlb, line 289: Same as above.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
danhermann: Good call

bryanlb (approved):

PR #426, opened 2022-11-29
Add rejected execution handler that synchronously enqueues work item

danhermann:
I believe this is a better approach than https://github.com/slackhq/kaldb/pull/427

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
+++ b/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
@@ -17,6 +17,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -255,7 +244,7 @@ public boolean consumeMessagesBetweenOffsetsInParallel(
     LOG.info("Pool size for queue is: {}", poolSize);
 
     // TODO: Track and log errors and success better.
-    BlockingArrayBlockingQueue<Runnable> queue = new BlockingArrayBlockingQueue<>(100);
+    ArrayBlockingQueue<Runnable> queue = new ArrayBlockingQueue<>(100);
```
mansu, line 247: Can we add a unit test to this use case to make sure the code works as expected?

danhermann: I think there will need to be some restructuring of the code to be able to test that particular change in isolation.

mansu: I think it's worth doing it since I am not quite sure looking at the code if it will work as intended.

danhermann:
Closing this as the root issue was addressed in https://github.com/slackhq/kaldb/pull/430

PR #427, opened 2022-11-29
Make executor work queue blocking

danhermann:
I think the approach in https://github.com/slackhq/kaldb/pull/426 is better.


kaldb/src/test/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumerTest.java
```diff
@@ -235,5 +235,26 @@ public void testThrowingConsumer() throws Exception {
 
       testConsumer.close();
     }
+
+    @Test
```
mansu, line 239: Let's add this unit test to the existing code.

mansu: Added!

danhermann:
Closing this as the root issue was addressed in https://github.com/slackhq/kaldb/pull/430

PR #428, opened 2022-11-30
validate timestamp for messages in the indexer

vthacker:
validate timestamp for messages in the indexer and the pre-processor

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
@@ -204,14 +204,14 @@ public class IndexingChunkManager<T> extends ChunkManagerBase<T> {
             @Override
             public void onSuccess(Boolean success) {
               if (success == null || !success) {
-                LOG.warn("Roll over failed");
+                LOG.error("RollOverChunkTask success=false for chunk={}", currentChunk.info());
                 stopIngestion = true;
               }
             }
 
             @Override
             public void onFailure(Throwable t) {
-              LOG.warn("Roll over failed with an exception", t);
+              LOG.error("Roll over failed with an exception for chunk={}", currentChunk.info(), t);
               stopIngestion = true;
             }
           },
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
@@ -1,6 +1,7 @@
 package com.slack.kaldb.logstore;
 
 import com.slack.kaldb.proto.config.KaldbConfigs;
+import com.slack.kaldb.util.RuntimeHalterImpl;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.io.File;
@@ -201,9 +202,9 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
       LOG.error(String.format("Indexing message %s failed with error:", message), e);
       messagesFailedCounter.increment();
     } catch (IOException e) {
-      // TODO: For now crash the program on IOException since it is likely a serious issue.
-      // In future may need to handle this case more gracefully.
-      e.printStackTrace();
+      // TODO: In future may need to handle this case more gracefully.
+      LOG.error("failed to add document", e);
+      new RuntimeHalterImpl().handleFatal(e);
     }
   }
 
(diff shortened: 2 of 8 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -175,7 +175,7 @@ private LogMessage buildLogMessage(IndexSearcher searcher, ScoreDoc hit) {
       LogWireMessage wireMessage = JsonUtil.read(s, LogWireMessage.class);
       return new LogMessage(
           wireMessage.getIndex(), wireMessage.getType(), wireMessage.id, wireMessage.source);
-    } catch (IOException e) {
+    } catch (Exception e) {
```
vthacker, line 178: We should be catching all exceptions, because some of the work potentially throws Runtime exceptions which won't get caught otherwise.

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
```diff
@@ -150,6 +151,17 @@ public Predicate<String, Trace.Span> createRateLimiter(
         return false;
       }
 
+      if (value.getTimestamp() <=0) {
```
vthacker, line 154: @bryanlb what do you think about addition validation here in the pre-processor in addition to the indexer.

If we add validation here, I think this is the code we need. But I need to go edit all unit tests to add a timestamp field .

Before I do that wanted to check if we want to add validation here :)

kaldb/src/test/java/com/slack/kaldb/preprocessor/KaldbSerdesTest.java
```diff
@@ -36,7 +36,7 @@ public void shouldSerializeAndDeserializeMurronMessage() {
     assertThat(deserializedMessage.getMessage()).isEqualTo(ByteString.copyFromUtf8(message));
     assertThat(deserializedMessage.getHost()).isEqualTo(host);
     assertThat(deserializedMessage.getType()).isEqualTo(type);
-    assertThat(timestamp).isEqualTo(timestamp);
+    assertThat(deserializedMessage.getTimestamp()).isEqualTo(timestamp);
```
vthacker, line 39: typo fix

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -204,14 +204,14 @@ private void doRollover(ReadWriteChunk<T> currentChunk) {
             @Override
             public void onSuccess(Boolean success) {
               if (success == null || !success) {
-                LOG.warn("Roll over failed");
+                LOG.error("RollOverChunkTask success=false");
```
bryanlb, line 207: Why don't we include the `currentChunk.info()` in this message as well?

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -204,14 +204,14 @@ private void doRollover(ReadWriteChunk<T> currentChunk) {
...
                 stopIngestion = true;
               }
             }
 
             @Override
             public void onFailure(Throwable t) {
-              LOG.warn("Roll over failed with an exception", t);
+              LOG.error("Roll over failed with an exception", t);
```
bryanlb, line 214: Same as above, current chunk info may be useful here

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -203,7 +203,7 @@ public void addMessage(LogMessage message) {
     } catch (IOException e) {
       // TODO: For now crash the program on IOException since it is likely a serious issue.
       // In future may need to handle this case more gracefully.
-      e.printStackTrace();
+      LOG.error("failed to add document", e);
```
bryanlb, line 206: Per the comment this should actually crash the program correct? Do we want to do a `new RuntimeHalterImpl().handleFatal(e);` as we do in the KaldbIndexer.java class?

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
```diff
@@ -150,6 +151,17 @@ public Predicate<String, Trace.Span> createRateLimiter(
         return false;
       }
 
+      if (value.getTimestamp() <= 0) {
```
bryanlb, line 154: I'm not a huge fan of this sitting here - it's not really part of rate limiting; If we intend to drop this due to a missing timestamp within the preprocessor, in the value mapper would be a better fit for it.

vthacker: actually you are right! The value mapper is before the rate limiter so lemme move it there

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -204,14 +204,14 @@ private void doRollover(ReadWriteChunk<T> currentChunk) {
             @Override
             public void onSuccess(Boolean success) {
               if (success == null || !success) {
-                LOG.warn("Roll over failed");
+                LOG.error("RollOverChunkTask success=false for chunk=" + currentChunk.info());
```
bryanlb, line 207: Use `{}` instead of the string concat for these.

bryanlb (changes requested):

bryanlb (approved):

PR #430, opened 2022-12-01
Check kafka offsets before handling recovery tasks

danhermann:


kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
```diff
@@ -217,41 +231,49 @@ private void handleRecoveryTaskAssignment(RecoveryNodeMetadata recoveryNodeMetad
...
-      kafkaConsumer.close();
-      chunkManager.stopAsync();
-      chunkManager.awaitTerminated(DEFAULT_START_STOP_DURATION);
-      LOG.info("Finished handling the recovery task: {}", recoveryTaskMetadata);
-      return success;
-    } catch (Exception ex) {
-      LOG.error("Exception in recovery task [{}]: {}", recoveryTaskMetadata, ex);
-      return false;
+
+    var validatedRecoveryTask = validateKafkaOffsets(recoveryTaskMetadata);
```
bryanlb, line 235: I would avoid using `var` across this PR and use the actual types. We don't use `var` anywhere else in the project today, and it makes it a little tougher to read the code.

kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
```diff
@@ -278,4 +300,83 @@ private void setRecoveryNodeMetadataState(
...
+      // this should never happen, but if it somehow did, the requested recovery range should
+      // be adjusted down to the latest available offset in Kafka
+      LOG.warn(
+          "Partial loss of messages in recovery task. End offset {}, latest available offset {}",
+          recoveryTask.endOffset,
+          latestKafkaOffset);
+      newEndOffset = latestKafkaOffset;
+    }
+
+    return new RecoveryTaskMetadata(
```
bryanlb, line 357: This seems a bit awkward - we technically just need the pair of updated offsets (or null) as we never directly use the returned `RecoveryTaskMetadata` object. This would be better if we had a dedicated object for this (ie, `KafkaOffsetPair`) instead of reusing the recovery task object.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -77,6 +77,7 @@ private static Properties makeKafkaConsumerProps(
         "org.apache.kafka.common.serialization.ByteArrayDeserializer");
     // TODO: Does the session timeout matter in assign?
     props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, kafkaSessionTimeout);
+    props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "none");
```
bryanlb, line 80: A comment here explaining why this is "none" would be good - just to prevent someone from attempting to change it in the future without realizing the motivation.

bryanlb (changes requested):

bryanlb (approved):

PR #431, opened 2022-12-01
Add `_all` field for all the indexed lucene documents.

mansu:
Add an `_all` field to all indexed lucene documents so we can perform full text search on the documents.


bryanlb (changes requested):
Let's make this a config option, but otherwise it looks great!

mansu:
> Let's make this a config option, but otherwise it looks great!

Done!

mansu:
> Changes look good except we still need to set the master `config.yaml` to include this, defaulted to FALSE, and allow an override by ENV similar to the other configs.
> 
> https://github.com/slackhq/kaldb/blob/master/config/config.yaml

Good call out! Updated config.yaml.

config/config.yaml
```diff
@@ -6,6 +6,7 @@ indexerConfig:
   luceneConfig:
     commitDurationSecs: ${INDEXER_COMMIT_DURATION_SECS:-10}
     refreshDurationSecs: ${INDEXER_REFRESH_DURATION_SECS:-11}
+    enableFullTextSearch: ${INDEXER_ENABLE_FULL_TEXT_SEARCH:false}
```
bryanlb, line 9: This still isn't quite right - this needs to be `:-false`

bryanlb (changes requested):

config/config.yaml
mansu: oops. Done.

bryanlb (approved):

PR #432, opened 2022-12-01
fix getTraceById timestamp

vthacker:
1. if we search by traceId query back upto 24 hours instead of the default 2. We keep the lookahead 2 hours ( happy to reduce it, but I believe Suman wanted some lookahead buffer to account for time skew )
2. Re-enable the unit test - I ran to CI test 5 times already and no flakiness ( unsure what changed TBH )


bryanlb:
Closing per our discussion last week - we'll leave this at 2 hours, and to increase this we'll need a new datasource plugin for grafana.

PR #436, opened 2022-12-06
Make `_all` field the default search field.

mansu:
Make `_all` field the default search field so we don't have to enter it every time we have to do a full text search on logs.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
@@ -10,6 +10,7 @@ import static com.slack.kaldb.testlib.MetricsUtil.getCount;
 import static com.slack.kaldb.testlib.MetricsUtil.getTimerCount;
 import static com.slack.kaldb.testlib.TemporaryLogStoreAndSearcherRule.MAX_TIME;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
 
 import brave.Tracing;
 import com.slack.kaldb.logstore.LogMessage;
@@ -498,8 +499,14 @@ public class LogIndexSearcherImplTest {
                 .size())
         .isEqualTo(3);
 
-    // Currently, returns empty since we don't parse wild card queries.
-    // TODO: One we start parsing wild card queries this should return 3.
+    assertThat(
+            strictLogStore
+                .logSearcher
+                .search(TEST_DATASET_NAME, "app*", 0, MAX_TIME, 1000, 1)
+                .hits
+                .size())
+        .isEqualTo(2);
+
     assertThat(
             strictLogStore
                 .logSearcher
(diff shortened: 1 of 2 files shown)


vthacker:
I don't think this is quite right. What if all field is disabled? This PR doesn't take into account of that

Secondly, what if all field was enabled for some chunks and then disabled for some chunks?

vthacker:
> I don't think this is quite right. What if all field is disabled? This PR doesn't take into account of that
> 
> Secondly, what if all field was enabled for some chunks and then disabled for some chunks?

The same questions would apply to the default Message field that we use today. So for now let's merge

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -382,26 +382,112 @@ public void testFullTextSearch() {
         makeMessageWithIndexAndTimestamp(1, "apple", TEST_DATASET_NAME, time.plusSeconds(4));
     msg1.addProperty("field1", "1234");
     strictLogStore.logStore.addMessage(msg1);
+    strictLogStore.logStore.commit();
+    strictLogStore.logStore.refresh();
+    // Search using _all field.
+    assertThat(
```
vthacker, line 388: can we add a test with multiple words with spaces and some numbers etc ?

bryanlb: Let's also go ahead and add tests for empty string and wildcard as well.

mansu: Done.

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -382,26 +382,157 @@ public void testFullTextSearch() {
...
+                .search(TEST_DATASET_NAME, "", 0, MAX_TIME, 1000, 1)
+                .hits
+                .size())
+        .isEqualTo(3);
 
-    SearchResult<LogMessage> babyInAll =
-        strictLogStore.logSearcher.search(TEST_DATASET_NAME, "_all:baby", 0, MAX_TIME, 1000, 1);
-    assertThat(babyInAll.hits.size()).isEqualTo(2);
+    // Currently, returns empty since we don't parse wild card queries.
+    // TODO: One we start parsing wild card queries this should return 3.
```
bryanlb, line 502: This isn't accurate - for fields that are analyzed wildcard queries do work today. You can see this if you query for "app*"  in this test you should have results for "apple". Additionally if this is just "*" it returns the entire dataset when the field is analyzed.

bryanlb (changes requested):

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
mansu: Good catch. Updated the tests for wildcard queries. * throws an exception. If we want * to not throw an exception, we should update the query parser to convert that to empty query.

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -424,15 +580,38 @@ public void testDisabledFullTextSearch() {
...
+                .size())
+        .isEqualTo(0);
+
+    assertThat(
+            strictLogStoreWithoutFts
+                .logSearcher
+                .search(TEST_DATASET_NAME, "1234", 0, MAX_TIME, 1000, 1)
+                .hits
+                .size())
+        .isEqualTo(0);
```
bryanlb, line 614: Probably would be good to include the test asserts in the full text search disabled case as well, and not just the enabled one.

mansu: Most of the additional asserts are related to how query parser handles the queries and not the all field. So, those tests are not as relevant here. However, added them here for completeness. 

We also have simpler higher level test at the chunk level, so we are covered there.

bryanlb (approved):

PR #438, opened 2022-12-07
Bump dependency versions

bryanlb:
Bumps several vulnerable dependencies, bumps Armeria to most recent version.

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -9,18 +9,22 @@
     <packaging>jar</packaging>
 
     <properties>
+        <!--
+            When possible, attempt to keep our direct dependencies
+            aligned with the Armeria included versions.
+        -->
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <javac.target>11</javac.target>
         <uberjar.name>kaldb</uberjar.name>
         <http.client.version>4.5.13</http.client.version>
         <http.core.version>4.4.15</http.core.version>
         <protobuf.version>3.21.7</protobuf.version>
-        <grpc.version>1.47.0</grpc.version>
-        <armeria.version>1.17.0</armeria.version>
-        <micrometer.version>1.9.1</micrometer.version>
-        <kafka.version>3.2.0</kafka.version>
-        <jackson.version>2.13.3</jackson.version>
-        <jackson.databind.version>2.12.7</jackson.databind.version>
+        <grpc.version>1.49.1</grpc.version>
+        <micrometer.version>1.9.4</micrometer.version>
+        <armeria.version>1.20.3</armeria.version>
+        <kafka.version>3.2.3</kafka.version>
+        <jackson.version>2.14.1</jackson.version>
+        <jackson.databind.version>2.14.1</jackson.databind.version>
         <lucene.version>9.3.0</lucene.version>
         <curator.version>5.2.1</curator.version>
         <log4j.version>2.17.2</log4j.version>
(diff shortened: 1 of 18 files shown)


kaldb/pom.xml
```diff
@@ -99,12 +99,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
```
bryanlb, line 102: We added this to the classpath to test `io_uring`, but are blocked on further unrelated infrastructure changes. Pulling out for now as this is an incubator dep, and we'll revisit once we're ready to try `io_uring` again.

kaldb/pom.xml
```diff
@@ -15,12 +15,12 @@
         <http.client.version>4.5.13</http.client.version>
         <http.core.version>4.4.15</http.core.version>
         <protobuf.version>3.21.7</protobuf.version>
-        <grpc.version>1.47.0</grpc.version>
-        <armeria.version>1.17.0</armeria.version>
-        <micrometer.version>1.9.1</micrometer.version>
-        <kafka.version>3.2.0</kafka.version>
-        <jackson.version>2.13.3</jackson.version>
-        <jackson.databind.version>2.12.7</jackson.databind.version>
+        <grpc.version>1.49.1</grpc.version>
```
bryanlb, line 18: Keeps grpc/micrometer consistent with what ships in Armeria.

kaldb/pom.xml
```diff
@@ -437,6 +431,10 @@
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-log4j12</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.module</groupId>
+                    <artifactId>jackson-module-scala_2.13</artifactId>
```
bryanlb, line 436: This was causing us problems attempting to target a newer databind release, and doesn't appear necessary for our use case.

kaldb/src/test/java/com/slack/kaldb/chunk/ReadOnlyChunkImplTest.java
```diff
@@ -55,17 +55,17 @@
...
-      String.format("%sBucket", this.getClass().getSimpleName()).toLowerCase();
+  private static final String TEST_S3_BUCKET = "read-only-chunk-impl-test";
 
   private TestingServer testingServer;
   private MeterRegistry meterRegistry;
   private S3BlobFs s3BlobFs;
 
-  @ClassRule public static final S3MockRule S3_MOCK_RULE = S3MockRule.builder().silent().build();
+  @ClassRule
+  public static final S3MockRule S3_MOCK_RULE =
```
bryanlb, line 67: This is due to a change in Adobe S3Mock 2.5 that started validating against duplicate bucket names - https://github.com/adobe/S3Mock/blob/main/CHANGELOG.md#250. Since we used a class rule but tried to init the same bucket in `@Each`, this would cause subsequent tests to fail.

kaldb/pom.xml
mansu: Please make this a comment in the code so we can update it when we update armeria next.

kaldb/pom.xml
mansu: Please add this as a comment also, so we don't delete it by accident in future.

mansu (changes requested):
LGTM, minor comments.

kaldb/pom.xml
bryanlb: Done.

kaldb/src/test/java/com/slack/kaldb/clusterManager/RecoveryTaskAssignmentServiceTest.java
```diff
@@ -832,6 +833,7 @@ public void shouldHandleNodesAvailableFirstLifecycle() throws Exception {
   }
 
   @Test
+  @Ignore // Flakey, occasionally throws InternalMetadataStore on the recoveryNodeMetadataStore
```
bryanlb, line 836: https://github.com/slackhq/kaldb/actions/runs/3447143804/jobs/5752847656
https://github.com/slackhq/kaldb/actions/runs/3474602284/jobs/5807950536

mansu (approved):

PR #439, opened 2022-12-08
Add more options to provide KafkaConfig parameters

vthacker:
- Remove the TODOs that wanted to move KafkaConfig to IndexerConfig and RecoveryConfig. We implement that in this PR. When committing we need to change our bedrock.yaml file
- Add additional props to provide params that aren't limited to the named parameters we already support within KafkaConfig. This allows us to experiment with new configs to tune the cluster without code changes. This also folds in the overrideProps

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
+++ b/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
@@ -49,13 +49,6 @@ public class KaldbKafkaConsumer {
           ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
           ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
 
-  public static KaldbKafkaConsumer fromConfig(
-      KaldbConfigs.KafkaConfig kafkaCfg,
-      LogMessageWriterImpl logMessageWriter,
-      MeterRegistry meterRegistry) {
-    return new KaldbKafkaConsumer(kafkaCfg, logMessageWriter, meterRegistry);
-  }
-
   @VisibleForTesting
   public static Properties makeKafkaConsumerProps(KaldbConfigs.KafkaConfig kafkaConfig) {
 
--- a/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
+++ b/kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
@@ -43,16 +43,7 @@ public class KaldbKafkaConsumer {
       KaldbConfigs.KafkaConfig kafkaCfg,
       LogMessageWriterImpl logMessageWriter,
       MeterRegistry meterRegistry) {
-    return new KaldbKafkaConsumer(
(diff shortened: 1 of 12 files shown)


kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -38,57 +41,81 @@ public class KaldbKafkaConsumer {
...
   private final LogMessageWriterImpl logMessageWriterImpl;
 
+  private static final String[] REQUIRED_CONFIGS = {ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG};
+  private static final Set<String> OVERRIDABLE_CONFIGS =
+      Set.of(
+          ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG,
+          ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
+          ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+
   public static KaldbKafkaConsumer fromConfig(
```
danhermann, line 51: Is this method still useful now that it just passes its three parameters to a public constructor with the same three parameters?

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -38,57 +41,81 @@ public class KaldbKafkaConsumer {
...
+    String userValue = props.getProperty(key);
+    if (userValue != null) {
+      if (override) {
+        LOG.warn(
+            String.format(
+                "Property %s is provided but will be overridden from %s to %s",
+                key, userValue, value));
+        props.setProperty(key, value);
+        overridden = true;
+      }
```
danhermann, line 112: Maybe also warn in the case that the user specifies a Kafka property in the `addtionalPropsMap` that cannot be overridden?

kaldb/src/main/proto/kaldb_configs.proto
```diff
@@ -192,6 +192,7 @@ message ManagerConfig {
 // Config for the recovery node.
 message RecoveryConfig {
   ServerConfig server_config = 1;
+  KafkaConfig kafka_config = 10;
```
danhermann, line 195: I don't love that we allow for divergence of some Kafka settings (e.g., bootstrap servers) that should never diverge, but may not be worth fragmenting the config structure to disallow that.

danhermann (approved):
LGTM. Left a couple comments below but none that are blocking.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
vthacker: good catch! removed it

kaldb/src/main/proto/kaldb_configs.proto
vthacker: yeah! But having it top level means it's leaky and potentially the manager can try to access these settings and fail in weird ways?

danhermann (approved):

danhermann (dismissed):

danhermann (approved):

PR #441, opened 2022-12-08
Log duration of recovery tasks

danhermann:
Record durations of some subtasks in the processing of recovery tasks to assist with profiling. These are optional and could be removed later (or now) if they don't have long-term usefulness.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
@@ -3,8 +3,6 @@ package com.slack.kaldb.recovery;
 import static com.slack.kaldb.server.KaldbConfig.DEFAULT_START_STOP_DURATION;
 import static com.slack.kaldb.server.ValidateKaldbConfig.INDEXER_DATA_TRANSFORMER_MAP;
 
-import brave.ScopedSpan;
-import brave.Tracing;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.AbstractIdleService;
@@ -35,6 +33,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.ListOffsetsResult;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -287,31 +288,41 @@ public boolean consumeMessagesBetweenOffsetsInParallel(
...
-                  messagesOutsideOffsetRange.incrementAndGet();
-                  recordsFailedCounter.increment();
-                } else {
-                  try {
-                    if (logMessageWriterImpl.insertRecord(record)) {
-                      recordsReceivedCounter.increment();
-                    } else {
-                      recordsFailedCounter.increment();
+              long startTime = System.nanoTime();
+              UUID batchUuid = UUID.randomUUID();
```
danhermann, line 292: Adds a batch UUID to uniquely identify log entries for this batch. Previously, the 3-tuple of [topic, partition, batch record count] was used and that 3-tuple is very much not unique.

kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
```diff
@@ -281,9 +289,25 @@ boolean handleRecoveryTask(RecoveryTaskMetadata recoveryTaskMetadata) {
       } catch (Exception ex) {
         LOG.error("Exception in recovery task [{}]: {}", validatedRecoveryTask, ex);
         return false;
+      } finally {
+        long endTime = System.nanoTime();
+        LOG.info(
+            "Recovery task {} took {}µs, (subtask times {}, {}, {}, {})",
```
mansu, line 295: I think it would be great to prefix the sub-task time so we can understand the time without looking at the code. 

Also, may be time we look at using a structured logging library.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
mansu: I think dropping the uuid would be ideal since beyond the scope of this function, the uuid has no utility. Further, it would be good to any extra work on ingestion.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
danhermann: https://github.com/slackhq/kaldb/pull/441/commits/d19a91277361f5fe82ddfeaa44b48afff1e892ad

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
danhermann: It's easy to drop the uuid, but its value comes from the fact that right now we have a bunch of interleaved log messages that look like this:

```
Ingesting batch: [kaldb_preprocessor-1/500]
Ingesting batch: [kaldb_preprocessor-1/500]
Finished ingesting batch: [kaldb_preprocessor-1/500]
Ingesting batch: [kaldb_preprocessor-1/500]
Finished ingesting batch: [kaldb_preprocessor-1/500]
Finished ingesting batch: [kaldb_preprocessor-1/500]
```

We can't identify the "Finished" log message that corresponds to an "Ingesting batch" message because `topic-partionId/recordCount` doesn't uniquely identify a batch. Is that useful or do you still want the UUID dropped?

mansu: Can we use the thread-id instead to know the start stop instead of a new uuid?

mansu (dismissed):

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
danhermann: :+1:

mansu (approved):

PR #443, opened 2022-12-09
Reduce cardinality of chunk assignment timers

bryanlb:
Reduces the cardinality of the chunk assignment timers by removing the slot information. For sufficiently large kaldb clusters this can result in poor query performance when using these metrics.


mansu (approved):
Good call!

PR #446, opened 2022-12-14
Time conversion utility methods for succinctness

danhermann:


kaldb/src/main/java/com/slack/kaldb/recovery/RecoveryService.java
```diff
@@ -314,20 +314,17 @@ boolean handleRecoveryTask(RecoveryTaskMetadata recoveryTaskMetadata) {
...
             recoveryTaskMetadata,
-            TimeUnit.MICROSECONDS.convert(endTime - startTime, TimeUnit.NANOSECONDS),
-            TimeUnit.MICROSECONDS.convert(offsetsValidatedTime - startTime, TimeUnit.NANOSECONDS),
-            TimeUnit.MICROSECONDS.convert(
-                consumerPreparedTime - offsetsValidatedTime, TimeUnit.NANOSECONDS),
-            TimeUnit.MICROSECONDS.convert(
-                messagesConsumedTime - consumerPreparedTime, TimeUnit.NANOSECONDS),
-            TimeUnit.MICROSECONDS.convert(
-                rolloversCompletedTime - messagesConsumedTime, TimeUnit.NANOSECONDS));
+            nanosToMicros(endTime - startTime),
```
mansu, line 317: I think it would be confusing to have different units for different log messages. Can we just use millis here also? Also, please update the units in the log message above for sub-task times.

PR #449, opened 2022-12-16
Add full text search to `SchemaAwareLogDocumentBuilder`. 

mansu:
Add full text search to `SchemaAwareLogDocumentBuilder`.  Also, fixed a bug in lucene field definition. Added new tests.

Pre-fixed internal fields with '_' following ES conventions. Moved Type into reserved field. Updated tests.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LogMessage.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LogMessage.java
@@ -29,10 +29,10 @@ public class LogMessage extends LogWireMessage {
   public enum SystemField {
     // The source field contains the input document.
     SOURCE("_source"),
-    ID("id"),
-    INDEX("index"),
+    ID("_id"),
+    INDEX("_index"),
     TIME_SINCE_EPOCH("_timesinceepoch"),
-    TYPE("type"),
+    TYPE("_type"),
     ALL("_all");
 
     public final String fieldName;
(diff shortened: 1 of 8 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/LogMessage.java
```diff
@@ -117,9 +115,7 @@ && getType() != null
 
   private BadMessageFormatException raiseException(Throwable t) {
     throw new BadMessageFormatException(
-        String.format(
-            "Index:%s, Type: %s, Id: %s, Source: %s".format(getIndex(), getType(), id, source)),
-        t);
+        "Index:%s, Type: %s, Id: %s, Source: %s".format(getIndex(), getType(), id, source), t);
```
bryanlb, line 118: This looks like a refactor went wrong, as this is invalid syntax (static method accessed via an instance reference). IntelliJ should also have highlighted this as incorrect.

```java
System.out.println("foo %s".format("bar"));
```
will output `bar`, not `foo bar`.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/LogMessage.java
mansu: Yeah, looks like the refactor is wrong. Fixed.

bryanlb (approved):

PR #453, opened 2023-01-01
Add a starter grafana dashboard for KalDB.

mansu:


bryanlb (changes requested):
As discussed on Slack, we should revisit how this defined. This config file as it exists here is too specific and cannot easily be adopted by another team. Additionally there it not an easy way to ensure this is kept up to date moving forward.

Something like this is still needed, but using a dashboards-as-code or similar approach would be much less prone to going stale.

mansu:
For the time being, not merging this dashboard into the repo but keep this PR as a reference.

PR #454, opened 2023-01-04
allow additional kafka properties for the preprocessor

vthacker:
followup from https://github.com/slackhq/kaldb/pull/439/files , where we also allow for additional properties for the preprocessor.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
@@ -4,6 +4,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.slack.kaldb.metadata.dataset.DatasetMetadata.MATCH_ALL_SERVICE;
 import static com.slack.kaldb.metadata.dataset.DatasetMetadata.MATCH_STAR_SERVICE;
 import static com.slack.kaldb.server.KaldbConfig.DEFAULT_START_STOP_DURATION;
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -337,6 +337,12 @@ protected static Properties makeKafkaStreamsProps(
     props.put(
         StreamsConfig.PROCESSING_GUARANTEE_CONFIG, kafkaStreamConfig.getProcessingGuarantee());
 
+    // don't override any property we already set
+    for (Map.Entry<String, String> additionalProp :
+        kafkaStreamConfig.getAdditionalPropsMap().entrySet()) {
+      maybeOverride(props, additionalProp.getKey(), additionalProp.getValue(), false);
```
bryanlb, line 343: We don't necessarily need the `maybeOverride` method - we can just set the user values first and then set our values after. If you still wanted a log message you can just rely on the fact that putting a value returns the previous value if it existed, else null.

bryanlb (approved):

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
vthacker: I didn't want the user to override thing like acks, replication etc. So I thought this approach would be nicer.

Eventually we can have a list of params that are overridable and for those pass true

PR #455, opened 2023-01-05
expose consumer.fetch.min.bytes as a sysprop

vthacker:
The way we deploy currently, we need to allow for configurable params to be exposed as sysprops.

This is kinda lame, but for now this is the quickest way. To experiment with more kafka tunables, we might want to copy over the default config.yaml to our kube deployment and add more settings there?


config/config.yaml
```diff
@@ -114,6 +114,9 @@ preprocessorConfig:
     applicationId: ${KAFKA_APPLICATION_ID:-kaldb_preprocessor}
     numStreamThreads: ${KAFKA_STREAM_THREADS:-2}
     processingGuarantee: ${KAFKA_STREAM_PROCESSING_GUARANTEE:-at_least_once}
+    additionalProps:
```
mansu, line 117: We also need to add these props to json file also. Also, please update the KalDB unit tests that read this property.

mansu (changes requested):

mansu (approved):

PR #457, opened 2023-01-10
Do not throw error if we cannot convert a field

vthacker:
When we fail to convert a field ( during a field conflict situation ) we currently throw an exception which causees Kaldb to crash and in a kube environment the container gets restarted when this happens.

With this change we log a warning with an improved error message and also keep a counter for such failures.

With the improved log warn , my hope if we can address to cause as well in a followup PR


kaldb/src/main/java/com/slack/kaldb/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
```diff
@@ -264,12 +264,17 @@ static String makeNewFieldOfType(String key, FieldType valueType) {
...
 
-  private static void convertValueAndIndexField(
+  private void convertValueAndIndexField(
       Object value, FieldType valueType, LuceneFieldDef registeredField, Document doc, String key) {
     Object convertedValue =
         FieldType.convertFieldValue(value, valueType, registeredField.fieldType);
     if (convertedValue == null) {
-      throw new RuntimeException("No mapping found to convert value");
+      LOG.warn(
+          String.format(
```
bryanlb, line 273: A better approach here would be to use the logging token replacements, with the `{}` instead of a string format.

kaldb/src/main/java/com/slack/kaldb/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
```diff
@@ -264,12 +264,17 @@ static String makeNewFieldOfType(String key, FieldType valueType) {
...
+              "No mapping found to convert value from %s to %s",
+              registeredField.fieldType.name, registeredField.fieldType.name));
+      convertErrorCounter.increment();
+      ;
```
bryanlb, line 277: Extra `;`?

bryanlb: Should this return/exit here instead of proceeding to the `indexTypedField` method below?

bryanlb (changes requested):
We should have a small test for this to ensure that continuing with a warn doesn't cause any further downstream issues.

kaldb/src/main/java/com/slack/kaldb/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
vthacker: > Should this return/exit here instead of proceeding to the indexTypedField method below?
yep! gonna add a test now and then ask for a re-review

bryanlb (approved):

kaldb/src/test/java/com/slack/kaldb/logstore/schema/ConvertFieldValueAndDuplicateTest.java
```diff
@@ -0,0 +1,768 @@
...
+                .stream()
+                .filter(f -> f.name().equals(LogMessage.SystemField.ALL.fieldName))
+                .count())
+        .isEqualTo(1);
+    assertThat(convertFieldBuilder.getSchema().keySet())
+        .contains(LogMessage.SystemField.ALL.fieldName);
+  }
+
+  @Test
+  public void testConvertingAndDuplicatingConflictingBooleanField() throws JsonProcessingException {
```
vthacker, line 264: this is the only new test I've added

PR #461, opened 2023-01-12
add query parser

vthacker:


kaldb/src/main/java/com/slack/kaldb/logstore/search/queryparser/KaldbQueryParser.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public KaldbQueryParser(
+      String defaultField,
+      Analyzer analyzer,
+      ConcurrentHashMap<String, LuceneFieldDef> chunkSchema) {
+    super(defaultField, analyzer);
+    if (chunkSchema == null || chunkSchema.isEmpty()) {
+      // When we instantiate SchemaAwareLogDocumentBuilderImpl we load a bunch of default fields. So
+      // this should never be empty
+      throw new RuntimeException("This should never be empty");
```
bryanlb, line 20: This should have a slightly clearer message, as when we see the stack trace in the logs it would just say "This should never be empty" without clearly stating what, or what may be the issue.

bryanlb (changes requested):

bryanlb (approved):

PR #462, opened 2023-01-18
use Suppliers.memoize

vthacker:


kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -195,8 +194,9 @@ protected static Topology buildTopology(
     ValueMapper<byte[], Iterable<Trace.Span>> valueMapper =
         PreprocessorValueMapper.byteArrayToTraceSpans(dataTransformer);
 
+    int lingerMs = Integer.parseInt(kafkaProperties.getProperty("producer.linger.ms", "100"));
     StreamPartitioner<String, Trace.Span> streamPartitioner =
-        streamPartitioner(datasetMetadataList);
+        new PreprocessorPartitioner<>(datasetMetadataList, lingerMs);
```
mansu, line 199: Let's pass in all the kafka properties here so we can use other props in the future.

mansu (changes requested):

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
```diff
@@ -195,8 +194,9 @@ protected static Topology buildTopology(
     ValueMapper<byte[], Iterable<Trace.Span>> valueMapper =
         PreprocessorValueMapper.byteArrayToTraceSpans(dataTransformer);
 
+    int lingerMs = Integer.parseInt(kafkaProperties.getProperty("producer.linger.ms", "100"));
```
bryanlb, line 197: I wouldn't use lingerMs as the config for the sticky partitioner - I would have this as an entirely separate kaldb config option if we decide to use this approach.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorService.java
vthacker: addressed in the latest change

bryanlb (approved):

PR #465, opened 2023-02-03
Opensearch AutoDateHistogramAggregation

bryanlb:
Initial pass implementing the Opensearch AutoDateHistogram aggregations. This first PR swaps the internal implementation of our histogram logic to that of OpenSearch, but keeps as much of the existing request/response path as similar as possible.

Further PRs will expand on this to support different aggregation types.


bryanlb:
⚠️ Do not rebase this on master, due to #475

kaldb/pom.xml
```diff
@@ -153,6 +153,11 @@
             <artifactId>lucene-queryparser</artifactId>
             <version>${lucene.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.opensearch</groupId>
+            <artifactId>opensearch</artifactId>
+            <version>2.3.0</version>
```
vthacker, line 159: uses lucene 9.3.0 so the deps should all add up

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -80,8 +80,7 @@ private static ListeningExecutorService queryThreadPool() {
    * We will not aggregate locally for future use-cases that have complex group by etc
    */
   public SearchResult<T> query(SearchQuery query, Duration queryTimeout) {
-    SearchResult<T> errorResult =
-        new SearchResult<>(new ArrayList<>(), 0, 0, new ArrayList<>(), 0, 0, 1, 0);
+    SearchResult<T> errorResult = new SearchResult<>(new ArrayList<>(), 0, 0, 0, 0, 1, 0, null);
```
vthacker, line 83: can this also be s static final ?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultUtils.java
```diff
@@ -71,7 +67,11 @@ public static <T> KaldbSearch.SearchResult toSearchResultProto(SearchResult<T> s
     span.tag("totalSnapshots", String.valueOf(searchResult.totalSnapshots));
     span.tag("snapshotsWithReplicas", String.valueOf(searchResult.snapshotsWithReplicas));
     span.tag("hits", String.valueOf(searchResult.hits.size()));
-    span.tag("buckets", String.valueOf(searchResult.buckets.size()));
+    span.tag(
+        "internalAggregation",
+        searchResult.internalAggregation != null
+            ? searchResult.internalAggregation.toString()
```
vthacker, line 73: can this be massive? if I request bucketCount as 1000 and have some terms aggregations the payload could get big?

kaldb/src/test/java/com/slack/kaldb/testlib/MessageUtil.java
```diff
@@ -76,6 +76,10 @@ public static LogMessage makeMessageWithIndexAndTimestamp(
     Map<String, Object> fieldMap = new HashMap<>();
     fieldMap.put(LogMessage.ReservedField.TIMESTAMP.fieldName, timeStamp.toString());
     fieldMap.put(LogMessage.ReservedField.MESSAGE.fieldName, msgStr);
+    fieldMap.put(TEST_SOURCE_INT_PROPERTY, i);
```
vthacker, line 79: this did not fail any random tests?

kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbLocalQueryServiceTest.java
```diff
@@ -136,14 +138,18 @@ public void testKalDbSearch() throws IOException {
...
     // Test histogram buckets
-    assertThat(response.getBucketsList().size()).isEqualTo(2);
-    KaldbSearch.HistogramBucket bucket1 = response.getBuckets(0);
-    assertThat(bucket1.getCount()).isEqualTo(0);
-    assertThat(bucket1.getLow()).isEqualTo(chunk1StartTimeMs);
-    assertThat(bucket1.getHigh()).isEqualTo((chunk1StartTimeMs + chunk1EndTimeMs) / 2.0);
-    KaldbSearch.HistogramBucket bucket2 = response.getBuckets(1);
-    assertThat(bucket2.getCount()).isEqualTo(1);
-    assertThat(bucket2.getHigh()).isEqualTo(chunk1EndTimeMs);
+    InternalAutoDateHistogram dateHistogram =
```
vthacker, line 141: all these tests end up with the use-case where there is only 1 chunk. Do we have a test with multiple chunks so that the `reduce` also gets tested? [SearchResultAggregatorImplTest.java](https://github.com/slackhq/kaldb/pull/465/files#diff-6070f57fcffa18fd3e9116c27274b1b84c1f90443868f33337cb0fc74c500417) maybe?

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAggregationAdapter.java
```diff
@@ -0,0 +1,351 @@
...
+   * aggregations implementation.
+   */
+  public static Aggregator buildAutoDateHistogramAggregator(int numBuckets) throws IOException {
+    IndexSettings indexSettings = buildIndexSettings();
+    SimilarityService similarityService = new SimilarityService(indexSettings, null, emptyMap());
+    MapperService mapperService = buildMapperService(indexSettings, similarityService);
+
+    registerField(
+        mapperService,
+        LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName,
```
vthacker, line 328: How does "long" translate to from our schema?

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
bryanlb: It could be final, but it can't be static or moved out of this method due to it using generics.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultUtils.java
bryanlb: Yeah, this can be relatively massive. I went ahead and removed it.

kaldb/src/test/java/com/slack/kaldb/testlib/MessageUtil.java
bryanlb: It did not - the other test helper methods already added these additional fields, but this one specifically did not. I went ahead and added these so I could more easily test the other aggregators (ie, avg).

kaldb/src/test/java/com/slack/kaldb/logstore/search/KaldbLocalQueryServiceTest.java
bryanlb: In the class you identified we attempt to run an `aggregate` on multiple search results, which results in the internal `reduce`  logic being invoked across multiple search result sets.

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAggregationAdapter.java
bryanlb: This would map to the same ones you can put into OpenSearch using the mapping API, which the entirety is listed here - https://opensearch.org/docs/2.1/opensearch/supported-field-types/index/. For example, a date format would look something like the following:

```
b -> b.field("type", "date")
      .field("format", "E, d MMM yyyy HH:mm:ss Z")
      .field("locale", "de")
```

In the case of date, it's actually indexed as a long (https://opensearch.org/docs/2.1/opensearch/supported-field-types/date/), and then the mapping informs how to resolve this into a string when outputting the final format.

vthacker (approved):

PR #466, opened 2023-02-06
Add JSON as an input format

vthacker:
- Add a "json" log transformer to read incoming data in the preprocessor
- Add docs around Kaldb's getting started guide with JSON document as the example

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,9 @@ Index Data
 2. The input data transformer "json" (preprocessorConfig/dataTransformer in config.yaml) is how the preprocessor will parse the data.
 3. Each document must contain 2 mandatory fields - "service_name" and "timestamp" (DateTimeFormatter.ISO_INSTANT)
 4. There needs to be a dataset entry for the incoming data that maps the incoming service name
-   1. To create a dataset entry, go to the manager-...
-5. The preprocessor writes data into the following kafka topic "test-topic"(preprocessorConfig/downstreamTopic in config.yaml). We apply rate-limits etc.
-6. The indexer service is configured to read from "test-topic" (indexerConfig/kafkaConfig/kafkaTopic in config.yaml) and create lucene indexes locally
-7. 
-```
-
-```
+5. To create a dataset entry, go to the manager node (default http://localhost:8083/docs) and call CreateDatasetMetadata with name/owner as test and serviceNamePattern = "_all"
+6. The preprocessor writes data into the following kafka topic "test-topic"(preprocessorConfig/downstreamTopic in config.yaml). We apply rate-limits etc.
+7. The indexer service is configured to read from "test-topic" (indexerConfig/kafkaConfig/kafkaTopic in config.yaml) and creates lucene indexes locally
 
 Query via Grafana
 ```
--- a/README.md
+++ b/README.md
@@ -70,9 +70,11 @@ Index Data
 2. The input data transformer "json" (preprocessorConfig/dataTransformer in config.yaml) is how the preprocessor will parse the data.
 3. Each document must contain 2 mandatory fields - "service_name" and "timestamp" (DateTimeFormatter.ISO_INSTANT)
 4. There needs to be a dataset entry for the incoming data that maps the incoming service name
-5. To create a dataset entry, go to the manager node (default http://localhost:8083/docs) and call CreateDatasetMetadata with name/owner as test and serviceNamePattern = "_all"
-6. The preprocessor writes data into the following kafka topic "test-topic"(preprocessorConfig/downstreamTopic in config.yaml). We apply rate-limits etc.
-7. The indexer service is configured to read from "test-topic" (indexerConfig/kafkaConfig/kafkaTopic in config.yaml) and creates lucene indexes locally
+5. To create a dataset entry, go to the manager node (default http://localhost:8083/docs) and call CreateDatasetMetadata with name/owner as "test" and serviceNamePattern = "_all"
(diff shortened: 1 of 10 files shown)


mansu (changes requested):

mansu:
Looks I missed the draft tag on this one. Please feel free to ignore the review.

README.md
```diff
@@ -60,3 +61,32 @@ To start the application dependencies (Zookeeper, Kafka, S3) you can use the pro
 ```bash
 docker-compose up
 ```
+
+Compile KalDB
+```
+ mvn clean compile
+```
+
+Start KalDB
```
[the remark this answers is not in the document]
vthacker: We can - There is already an issue for it https://github.com/slackhq/kaldb/issues/469 . We can keep the scope separate from this PR

README.md
```diff
@@ -60,3 +61,32 @@ To start the application dependencies (Zookeeper, Kafka, S3) you can use the pro
...
+mvn -DskipTests=true clean package
+java -jar kaldb/target/kaldb.jar config/config.yaml
+```
+
+Index Data
+1. Data from the  "test-topic-in" (preprocessorConfig/kafkaStreamConfig/upstreamTopics in config.yaml) Kafka topic is read as input by the preprocessor.
+2. The input data transformer "json" (preprocessorConfig/dataTransformer in config.yaml) is how the preprocessor will parse the data.
+3. Each document must contain 3 mandatory fields - "service_name", "timestamp" ( what format? )
+4. There needs to be a dataset entry for the incoming data that maps the incoming service name
+   1. To create a dataset entry, go to the manager-...
```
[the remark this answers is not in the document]
vthacker: PR is in DRAFT state , so I'll add it before making it public

README.md
```diff
@@ -60,3 +61,32 @@ To start the application dependencies (Zookeeper, Kafka, S3) you can use the pro
...
+Start KalDB
+```
+mvn -DskipTests=true clean package
+java -jar kaldb/target/kaldb.jar config/config.yaml
+```
+
+Index Data
+1. Data from the  "test-topic-in" (preprocessorConfig/kafkaStreamConfig/upstreamTopics in config.yaml) Kafka topic is read as input by the preprocessor.
+2. The input data transformer "json" (preprocessorConfig/dataTransformer in config.yaml) is how the preprocessor will parse the data.
+3. Each document must contain 3 mandatory fields - "service_name", "timestamp" ( what format? )
```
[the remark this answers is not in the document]
vthacker: fixed

kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
```diff
@@ -13,16 +13,106 @@
...
+        tags.stream()
+            .anyMatch(
+                keyValue ->
+                    keyValue.getKey().equals(LogMessage.ReservedField.SERVICE_NAME.fieldName));
+    if (!containsServiceName) {
+      tags.add(
+          Trace.KeyValue.newBuilder()
+              .setKey(LogMessage.ReservedField.SERVICE_NAME.fieldName)
+              .setVType(Trace.ValueType.STRING)
+              .setVStr(serviceName)
```
vthacker, line 109:
```
// Replace hyphens with underscore since lucene doesn't like it.
```
`.replace("-", "_"))` was an artifact of using a text analyzer which we no longer use. 

So I'm removing it here as it will cause confusion down the road. We probably need to update some mappings internally

kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
```diff
@@ -13,16 +13,106 @@
...
+        tagBuilder.setVFloat64((double) entry.getValue());
+      } else if (entry.getValue() != null) {
+        tagBuilder.setVType(Trace.ValueType.BINARY);
+        tagBuilder.setVBinary(ByteString.copyFrom(entry.getValue().toString().getBytes()));
+      }
+      tags.add(tagBuilder.build());
+    }
+
+    // Add missing fields from murron message.
+    boolean containsHostName =
```
vthacker, line 82: `Trace.KeyValue` is lame! it should really be a map to allow quick checks to see if a key is present or not.

kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
```diff
@@ -13,16 +13,106 @@
...
+                  Trace.KeyValue.newBuilder()
+                      .setKey(LogMessage.ReservedField.HOSTNAME.fieldName)
+                      .setVType(Trace.ValueType.STRING)
+                      .setVStr(s)
+                      .build()));
+    }
+
+    spanBuilder.setId(ByteString.copyFrom(id.getBytes()));
+
+    boolean containsServiceName =
```
vthacker, line 99: previously without this check, if a `service_name` key was present, there would be two tags that end up with the same key

kaldb/src/main/java/com/slack/kaldb/writer/JsonLogFormatter.java
```diff
@@ -0,0 +1,55 @@
...
+import com.slack.kaldb.logstore.LogMessage;
+import com.slack.kaldb.util.JsonUtil;
+import com.slack.service.murron.trace.Trace;
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+public class JsonLogFormatter {
```
bryanlb, line 13: Add a comment here explaining what this class is for, and who is expected to use it. Specifically I can see how calling out the random UUIDs would be important, and how the default values are handled when they do not exist.

bryanlb (changes requested):
Little bit of class documentation and this looks good to go.

bryanlb (approved):

PR #470, opened 2023-02-10
Support for running KalDB in a docker container

### LINKED ISSUE #469: Allow KalDB to run in docker-compose
Currently docker-compose only runs KalDB's dependencies, such as ZooKeeper, Kafka, and Grafana. KalDB itself has a Dockerfile, but it's out of date and doesn't work anymore. 

We should fix the Dockerfile, and add the image to docker-compose, in a new docker profile so that "docker-compose up" still just runs the dependencies.

gjacoby126:
Resolves #469

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - ALLOW_ANONYMOUS_LOGIN=yes
 
   kafka:
-    image: 'bitnami/kafka:2.8.0'
+    image: 'bitnami/kafka:3.2.3'
     container_name: kaldb_kafka
     ports:
       - '9092:9092'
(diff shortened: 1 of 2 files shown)


mansu:
LGTM! It would be better to check with @vthacker before merging to make sure this change doesn't break the existing deploy pipeline.

mansu (approved):
LGTM! It would be better to check with @vthacker before merging to make sure this change doesn't break the existing deploy pipeline.

docker-compose.yml
```diff
@@ -57,3 +57,29 @@ services:
...
+      - 8080:8080
+      - 8081:8081
+      - 8082:8082
+      - 8083:8083
+      - 8085:8085
+      - 8086:8086
+    environment:
+      - KALDB_ZK_CONNECTION_STRING=zookeeper:2181
+      - KAFKA_BOOTSTRAP_SERVERS=kafka:29092
+    depends_on:
```
vthacker, line 77: I had to change the depends_on to this for it to run locally
```
    depends_on:
      - zookeeper
      - kafka
      - s3
      - openzipkin
```

I think the `depends_on` doesn't exist in docker 3 ? https://stackoverflow.com/questions/47710767/what-is-the-alternative-to-condition-form-of-depends-on-in-docker-compose-versio but then re-introduced in 3.9 https://stackoverflow.com/questions/71060072/docker-compose-depends-on-with-condition-invalid-type-should-be-an-array ?

gjacoby126: @vthacker - I removed the condition clauses to match what you have above and it still starts up properly locally for me. I've pushed the change up.

vthacker (approved):

PR #472, opened 2023-02-10
Fix timestamp indexing type

bryanlb:
Timetamp should be indexed as a long, and not a string so that we can perform aggregations on this field field.


vthacker (approved):

bryanlb:
Superseded by #512

PR #477, opened 2023-02-15
Add histogram aggregation request support

bryanlb:
This PR introduce the ability to make complex aggregation requests and obtain their results in our existing query pipeline. As this does not yet introduce the new query pipeline (see [wiki](https://github.com/slackhq/kaldb/wiki/Aggregators#requestresponse)), a few concessions were made around the return types using a bytearray. 

With complex aggregation support all of the additional fields on the date histogram are now supported (moving from an `AutoDateHistogram` to `DateHistogram` aggregation). Additionally, nested aggregations and the first metric aggregator (`avg`) are supported in this PR. This was done to ensure the design of the methods and objects could support this use case, and more aggregators will be provided in followup PRs.

A few additional items of note:
* Removes unnecessary layer of abstraction from the http API and the GRPC query node API (`EsSearchRequest`/`SearchRequestSort`/etc)
* Removes `totalCount` from `SearchResult` as this was both unused in the response, and didn't make sense once introducing other aggregation types.
* Fixes an issue with aggregations needing to perform a "final aggregation" to ensure some destructive steps are performed before returning the result set. This was needed for minimum bucket count support on the date histogram, among others.


kaldb/src/main/java/com/slack/kaldb/logstore/search/aggregations/AggBuilderBase.java
```diff
@@ -0,0 +1,42 @@
+package com.slack.kaldb.logstore.search.aggregations;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provides common functionality for all aggregations. All aggregations are expected to extend this
+ * class, or another base class that extends this one (ie, ValueSourceAggBuilder).
+ */
+public abstract class AggBuilderBase implements AggBuilder {
```
vthacker, line 10: I think we can call this class `AggBuilder` and not have the AggBuilder class that we have today

And then this class can declare `getType` as an abstract method?
```
abstract String getType();
```

I tried it locally and it seems to work

kaldb/src/main/java/com/slack/kaldb/logstore/search/aggregations/AggBuilderBase.java
```diff
@@ -0,0 +1,42 @@
...
+  // This is the name of the aggregation, as provided by the user. In most cases this will be a
+  // unique string ID to correlate the output values - Grafana generally uses incrementing
+  // integers as a string (ie "1", "2").
+  protected final String name;
+  protected final Map<String, Object> metadata;
```
vthacker, line 15: is the `metadata` used anywhere?

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -69,36 +71,36 @@ public LogIndexSearcherImpl(
...
+      AggBuilder aggBuilder) {
 
     ensureNonEmptyString(dataset, "dataset should be a non-empty string");
     ensureNonNullString(queryStr, "query should be a non-empty string");
     ensureTrue(startTimeMsEpoch >= 0, "start time should be non-negative value");
     ensureTrue(startTimeMsEpoch < endTimeMsEpoch, "end time should be greater than start time");
     ensureTrue(howMany >= 0, "hits requested should not be negative.");
-    ensureTrue(bucketCount >= 0, "bucket count should not be negative.");
-    ensureTrue(howMany > 0 || bucketCount > 0, "Hits or histogram should be requested.");
+    ensureTrue(howMany > 0 || aggBuilder != null, "Hits or aggregation should be requested.");
```
vthacker, line 96: I think we could add another check if both are passed? I think the current behaviour is if both are present we will follow the `howMany` block?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
```diff
@@ -46,10 +47,10 @@ public String toString() {
         + endTimeEpochMs
         + ", howMany="
         + howMany
-        + ", bucketCount="
-        + bucketCount
         + ", chunkIds="
-        + String.join(",", chunkIds)
+        + chunkIds
```
vthacker, line 51: is the default behaviour of List<String> == `String.join(",", chunkIds)` ?

kaldb/src/main/java/com/slack/kaldb/logstore/search/aggregations/AggBuilderBase.java
bryanlb: This is correct, we can simplify this as you mentioned and it will work. The reason for this approach was that it mirrors that of OpenSearch abstract class `AggregationBuilder` and the `BaseAggregationBuilder` [interface](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/BaseAggregationBuilder.java). The `AggBuilder` interface exists so that we have an easy option for adding pipeline aggregation support in the future when we decide to support it.

kaldb/src/main/java/com/slack/kaldb/logstore/search/aggregations/AggBuilderBase.java
bryanlb: Not yet - but this is implemented in OpenSearch `AbstractAggregationBuilder` and can be passed to the aggregator builders, so I went ahead and stubbed it out. I'm not sure if we'll eventually need it, so it's currently just implemented as empty maps for now.

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: If both are passed we build both result sets - the actual `N` log results (hits), and the aggregations. Requesting both hits and aggs is a common pattern, and is used by the custom Grafana plugin as well.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchQuery.java
bryanlb: Good catch - I just used IntelliJ to generate this method and didn't catch the difference. It is indeed similar to what you listed, but with extra brackets:

```java
List<String> stringList = List.of("1", "2", "3");
System.out.println(stringList);
```

```bash
[1, 2, 3]
```

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
vthacker: offcourse. I should have looked more closely at the `if-else` block

kaldb/src/main/java/com/slack/kaldb/logstore/search/aggregations/AggBuilderBase.java
vthacker: sgtm!

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -124,16 +121,12 @@ private EsSearchResponse doSearch(EsSearchRequest request) {
...
-              request.getAggregations(), searchResult.getInternalAggregations());
-
       return new EsSearchResponse.Builder()
           .hits(hits)
-          .aggregations(aggregations)
+          .aggregations(parseAggregations(searchResult.getInternalAggregations()))
           .took(Duration.of(searchResult.getTookMicros(), ChronoUnit.MICROS).toMillis())
           .shardsMetadata(searchResult.getTotalNodes(), searchResult.getFailedNodes())
-          .debugMetadata(getDebugAggregations(searchResult.getInternalAggregations()))
+          .debugMetadata(Map.of())
```
vthacker, line 129: for now, we don't need this `debugMetadata` ?

kaldb/src/test/java/com/slack/kaldb/testlib/TemporaryLogStoreAndSearcherRule.java
```diff
@@ -39,7 +40,14 @@ public static List<LogMessage> addMessages(
   public static List<LogMessage> findAllMessages(
       LogIndexSearcherImpl searcher, String dataset, String query, int howMany, int bucketCount) {
```
vthacker, line 41: bucketCount is unused. Maybe we can just remove it?

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
bryanlb: Correct, the previous use of `getDebugAggregations` was for rendering the raw response from the LogSearcherImpl. Since we directly return this now, there's not currently a need for returning debug metadata.

I left it initialized as an empty map so we remember we can use this, but this entire service will be revisited as part of the query layer rewrite.

vthacker (approved):

PR #484, opened 2023-02-22
Add terms aggregation support

bryanlb:


bryanlb:
Depends on #477 being merged first.

kaldb/src/test/java/com/slack/kaldb/testlib/TemporaryLogStoreAndSearcherRule.java
```diff
@@ -80,7 +85,15 @@ public TemporaryLogStoreAndSearcherRule(
             SchemaAwareLogDocumentBuilderImpl.build(
                 fieldConflictPolicy, enableFullTextSearch, metricsRegistry),
             metricsRegistry);
-    logSearcher = new LogIndexSearcherImpl(logStore.getSearcherManager(), logStore.getSchema());
+
+    ConcurrentHashMap<String, LuceneFieldDef> schema = logStore.getSchema();
+
+    // add schema definition for our string property
+    schema.put(
```
vthacker, line 92: is adding it here got to do with getting a "known" string field in the schema before we index a document? Because we init the open search adapter with the schema right at the start?

bryanlb: This is actually to inform the query of the schema to use, and is unrelated to indexing here. The aggregations use the schema information to correctly build out appropriate aggregation logic (ie, you might do a terms aggregation on either a numeric or text field).

vthacker (approved):
+1 . I think the meat and potatoes for this is `getTermsAggregationBuilder` and sub tricks we use to figure out the order and also to allow sub aggregations

I'd love some tests in the future if we are able to serialize/deserialize and reduce aggregations with more than 1 sub aggregation.

PR #492, opened 2023-02-25
indexer should expose transformer to output a single message

vthacker:
Kaldb Indexer doesn't have any transformer that can multiple messages at once. The preprocessor always writes out once `Trace.Span` proto 

This PR cleans up the interface for the indexer to only read one message at a time - Also incorporates https://github.com/slackhq/kaldb/pull/491 but we can commit them in separate PRs

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/writer/LogMessageTransformer.java
+++ b/kaldb/src/main/java/com/slack/kaldb/writer/LogMessageTransformer.java
@@ -1,11 +1,10 @@
 package com.slack.kaldb.writer;
 
 import com.slack.kaldb.logstore.LogMessage;
-import java.util.List;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 /** An interface a ConsumerRecord message from Kafka into a LogMessage. */
 @FunctionalInterface
 public interface LogMessageTransformer {
-  List<LogMessage> toLogMessage(ConsumerRecord<String, byte[]> record) throws Exception;
+  LogMessage toLogMessage(ConsumerRecord<String, byte[]> record) throws Exception;
 }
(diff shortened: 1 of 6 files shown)


kaldb/src/main/java/com/slack/kaldb/writer/LogMessageWriterImpl.java
```diff
@@ -65,24 +61,14 @@ public class LogMessageWriterImpl implements MessageWriter {
         final Murron.MurronMessage murronMsg =
             murronMessageDeserializer.deserialize("", record.value());
         Trace.Span apiSpan = MurronLogFormatter.fromApiLog(murronMsg);
-        return SpanFormatter.toLogMessage(Trace.ListOfSpans.newBuilder().addSpans(apiSpan).build());
```
vthacker, line 68: I believe we have left this transformer because so many tests depend on it. So we can keep it. However this also always emits one span.

kaldb/src/main/java/com/slack/kaldb/writer/LogMessageTransformer.java
```diff
@@ -1,11 +1,10 @@
 package com.slack.kaldb.writer;
 
 import com.slack.kaldb.logstore.LogMessage;
-import java.util.List;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 
 /** An interface a ConsumerRecord message from Kafka into a LogMessage. */
 @FunctionalInterface
 public interface LogMessageTransformer {
-  List<LogMessage> toLogMessage(ConsumerRecord<String, byte[]> record) throws Exception;
```
mansu, line 10: Keeping a list interface is a deliberate design so we can transform a single message or a list of messages(batch). While it may add some overhead, we should keep this interface since it's much more harder to add batch interface later.

mansu:
Keeping a list interface is a deliberate design so we can transform a single message or a list of messages(batch). While it may add some overhead, we should keep this interface since it's much more harder to add batch interface later.

vthacker:
> Keeping a list interface is a deliberate design so we can transform a single message or a list of messages(batch). While it may add some overhead, we should keep this interface since it's much more harder to add batch interface later.

sounds good to me!

PR #499, opened 2023-03-07
Add percentiles aggregation, pipeline aggregator support, moving average aggregation

bryanlb:
Adds ability to perform percentile aggregations. Also adds general pipeline aggregator support, and introduces the first pipeline aggregator (moving average).

Depends on #498.

- [x] Rework equals/hashcode for all  agg builders to use abstract base classes
- [x] Add additional testing for all equals/hashcodes on agg builders (https://github.com/slackhq/kaldb/pull/498#discussion_r1129873165)


kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAggregationAdapter.java
```diff
@@ -542,7 +710,11 @@ protected static TermsAggregationBuilder getTermsAggregationBuilder(TermsAggBuil
     }
 
     for (AggBuilder subAggregation : builder.getSubAggregations()) {
-      termsAggregationBuilder.subAggregation(getAggregationBuilder(subAggregation));
+      if (isPipelineAggregation(subAggregation)) {
```
vthacker, line 713: NBD - but if the pipeline aggregation hasn't been implemented yet, we'll throw an IllegalArgumentException from the else block saying `Aggregation type %s not yet supported"`

I think that's okay for now? Just wanted to highlight it though

vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAggregationAdapter.java
bryanlb: This is correct - we have the same logic in place for the non-pipeline aggregators. This is to prevent issues attempting to query aggregations that aren't supported, and getting unexpected results. 

When querying from Grafana via envoy this manifests as an http 500.

PR #513, opened 2023-03-27
Fix boolean field type

bryanlb:


bryanlb:
This will require the new query parser to support the `BooleanClause` query type before boolean values are returned as expected.

vthacker (approved):
nice!

PR #515, opened 2023-03-28
Add schema support for OpenSearchAdapter, use OpenSearch query parser

bryanlb:
Adds dynamic schema support for indexers, so newly added fields are queryable with the Opensearch query parser. This is done using a listener hook on the SearcherManager within the LogIndexSearcherImpl class. In addition, a few extras things of note are included:

* Adds support for remaining defined field types (`INTEGER`, `DOUBLE`, `FLOAT`, `BOOLEAN`)
* Adds check for trying to map metadata fields to the OpenSearch schema (previously causing exception on `_index`)
* Renamed `OpenSearchAggregationAdapter` to `OpenSearchAdapter`, as query support has been added to this class reusing the context objects
* Extracted the OpenSearch internal aggregation code to a new class, and marked it as deprecated along with a description why we don't want to use that long-term


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -68,15 +60,24 @@ public static SearcherManager searcherManagerFromPath(Path path) throws IOExcept
...
-    this.chunkSchema = chunkSchema;
-    this.openSearchAggregationAdapter = new OpenSearchAggregationAdapter(chunkSchema);
-  }
+    this.searcherManager.addListener(refreshListener);
 
-  // Lucene's query parsers are not thread safe. So, create a new one for every request.
-  private QueryParser buildQueryParser() {
-    return new KaldbQueryParser(SystemField.ALL.fieldName, analyzer, chunkSchema);
+    // initialize the adapter with whatever the default schema is
+    openSearchAdapter.reloadSchema();
```
vthacker, line 80: can this just be the default before in the open search adapter constructor?

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -103,194 +83,118 @@
...
+      String dataset, String queryStr, long startTimeMsEpoch, long endTimeMsEpoch)
+      throws IOException {
+    LOG.trace("Query raw input string: '{}'", queryStr);
+    try {
+      Query query;
+      if (queryStr != null
+          && !queryStr.isEmpty()
+          && !queryStr.equals("*:*")
+          && !queryStr.equals("*")) {
+        QueryStringQueryBuilder queryStringQueryBuilder =
```
vthacker, line 129: Is there a query parser we can leverage to generate the `queryStr`?

kaldb/src/test/java/com/slack/kaldb/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -250,6 +249,142 @@ public void testSearchMultipleItemsAndIndices() {
...
+            time.plusSeconds(2).toEpochMilli(),
+            10,
+            new DateHistogramAggBuilder(
+                "1", LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName, "1s"));
+    assertThat(notexists.hits.size()).isEqualTo(0);
+  }
+
+  @Test
+  public void testRangeQuery() throws InterruptedException {
+    Instant time = Instant.ofEpochSecond(1593365471);
```
vthacker, line 286: let's keep this dynamic so that it works with the new validation we are adding?

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: I had this approach initially, but I didn't like mixing the patterns where the consumer of the `OpenSearchAdapter ` class was expected to handle reloading themselves, but the first time was "automatic". I think it's much cleaner when the consumer is entirely responsible for this. It also allows for scenarios where you initialize the class, load the data, then initialize the schema without invoking an extra schema reload.

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
bryanlb: Good call - I had wanted to revisit this approach and forgot. 

I've refactored this so it doesn't use the query string in the same manner now. Instead of constructing a query string with all of the clauses and providing this to the string builder, I construct a bool query that combines a range query and the result of the query string builder with just the user's input string.

vthacker (approved):

PR #516, opened 2023-03-31
Add support for Opensearch mapping api

bryanlb:
Adds initial support for getting field schema information from the KalDB API. Also maps this response to something Openseach/Grafana can use, which is used for autocompleting fieldnames in the UI. Field conflict behavior is intentionally undefined at this point (relies on `putAll`, which will result in last-in wins).


kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -183,6 +187,29 @@ private HitsMetadata getHits(KaldbSearch.SearchResult searchResult) throws IOExc
...
   public HttpResponse mapping(@Param("indexName") Optional<String> indexName) throws IOException {
+    // todo - support query params including dataset and time window - this will require
+    //  rework of the Grafana plugin and will deviate from the Opensearch api spec
+
+    // Use a tree map so the results are naturally sorted
+    Map<String, Map<String, String>> propertiesMap = new TreeMap<>();
+    KaldbSearch.SchemaResult schemaResult =
+        searcher.getSchema(
+            KaldbSearch.SchemaRequest.newBuilder()
+                .setDataset("*")
```
gjacoby126, line 198: Wouldn't this only work if data from the past hour has been ingested? If so, that would mean that clusters with only historical data (or in the middle of being bootstrapped) couldn't find schema information.

Also, are we assuming that schema is immutable over time?

bryanlb: We do expect that the schema is mutable over time, and indeed there exists a possibility that two chunks may have different schema for a newly emerging field if it were being published in multiple formats (ie, `int` and `string`). The expectation is that you can use the KalDB API to query for a given time window, and get a schema that applies specifically to that time.

As of this PR, the GRPC api does allow you to query for a given time range and it return schema specific to that window. The Opensearch compatible API (and our fork of the Grafana Opensearch plugin) do not support passing date params, as they do not expect the schema to not change over time. As the Trino connector proof-of-concept is currently using this Opensearch API, I can certainly expose these as optional url params on this endpoint similar to the index name. I would expect we would want to extend our fork of the Grafana plugin to start passing these date params anyhow.

As to your other question regarding bootstrapping - I'm not certain this is necessarily a problem for KalDB. My expectation would be that the request for schema would follow what data is available for query - that is, if the data isn't loaded then the schema of the system would only reflect that which was.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -191,4 +194,11 @@ private SearchResult<T> incrementNodeCount(SearchResult<T> searchResult) {
   public List<Chunk<T>> getChunkList() {
     return chunkList;
   }
+
+  @Override
+  public Map<String, FieldType> getSchema() {
+    Map<String, FieldType> schema = new HashMap<>();
+    chunkList.forEach(chunk -> schema.putAll(chunk.getSchema()));
+    return Collections.unmodifiableMap(schema);
+  }
```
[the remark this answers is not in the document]
bryanlb: This is a good observation, and will indeed overwrite the values provided by other chunks. In this case the last-in "wins". In a properly running cluster we would not expect chunks to have different schemas for the same field, as that would mean someone is producing mixed content for the same field.

We haven't fully resolved what to do in this case - the idea is that for each indexer it would index it according to the predefined schema or for a newly emergent field the first type it sees. Conflicts from then on out would not be indexed to that field would would index to another field (ie, `_field_conflicts`). Since this is not a distributed decision however, it's possible that different indexer would have different types for this field depending on which version they saw first.

When we query for the schema of the cluster we now have an issue - what do we do when different nodes have different opinions about schema? This PR leaves this as "undefined" behavior as we haven't fully finished the schema work.

kaldb/src/main/java/com/slack/kaldb/chunk/Chunk.java
```diff
@@ -24,4 +26,6 @@
 
   /** Return true if the chunk contains data within that time range (epoch ms). */
   boolean containsDataInTimeRange(long startTs, long endTs);
+
+  Map<String, FieldType> getSchema();
```
vthacker, line 30: Hmm! I'm not a big fan of `Map<String, FieldType>`

In the current codebase, we expose a ChunkSchema as `Map<String, LuceneFieldDef>`

Thoughts on keeping it conistent with that and then in the mapping API extracting the fieldtype from it?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -433,4 +433,86 @@ public KaldbSearch.SearchResult doSearch(final KaldbSearch.SearchRequest request
...
     }
   }
+
+  @Override
+  public KaldbSearch.SchemaResult getSchema(KaldbSearch.SchemaRequest distribSchemaReq) {
+    // todo - this shares a significant amount of code with the distributed search request
+    //  and would benefit from refactoring the current "distributedSearch" abstraction to support
+    //  different types of requests
+
+    LOG.info("Starting distributed search for schema request: {}", distribSchemaReq);
```
vthacker, line 443: will INFO make it too noisy?

bryanlb: We have the queries logged today at an INFO level, and this should be significantly lower throughput than the query endpoint for our use case.

kaldb/src/main/java/com/slack/kaldb/chunk/Chunk.java
bryanlb: I didn't like using `LuceneFieldDef`, as the object returned via this API needs to be datasource agnostic. `LuceneFieldDef` today is mostly a wrapper on the `FieldType` that includes Lucene specific constructs (ie `storeDocValue`).

I think we may need another object type here that we could return the `FieldType` with instance-specific options (ie, date fields with different formatting). This abstraction currently does not exist however, so the `FieldType` is the best option we have today.

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -182,19 +186,38 @@ private HitsMetadata getHits(KaldbSearch.SearchResult searchResult) throws IOExc
    */
   @Get
   @Path("/:indexName/_mapping")
-  public HttpResponse mapping(@Param("indexName") Optional<String> indexName) throws IOException {
+  public HttpResponse mapping(
+      @Param("indexName") Optional<String> indexName,
+      @Param("startTimeEpochMs") Optional<Long> startTimeEpochMs,
```
vthacker, line 191: is the start/end a grafana thing? The docs link ( https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-mapping.html ) don't mention of a timestamp

bryanlb: The start/end time is a Kaldb option for the Trino POC - https://github.com/slackhq/kaldb/pull/516#discussion_r1156199036

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
```diff
@@ -433,4 +433,86 @@ public KaldbSearch.SearchResult doSearch(final KaldbSearch.SearchRequest request
...
+    ScopedSpan span =
+        Tracing.currentTracer().startScopedSpan("KaldbDistributedQueryService.distributedSchema");
+
+    Map<String, SnapshotMetadata> snapshotsMatchingQuery =
```
vthacker, line 447: Does this logic need to be same as the query path? Can't it be simpler like - "Query all live nodes" and don't try to find snapshots/search metadata etc?

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultUtils.java
```diff
@@ -390,6 +391,72 @@ public static SearchResult<LogMessage> fromSearchResultProto(
             protoSearchResult.getInternalAggregations().toByteArray()));
   }
 
+  public static FieldType fromSchemaDefinitionProto(
+      KaldbSearch.SchemaDefinition protoSchemaDefinition) {
+    if (protoSchemaDefinition.getType().equals(KaldbSearch.FieldType.TEXT)) {
+      return FieldType.TEXT;
+    } else if (protoSchemaDefinition.getType().equals(KaldbSearch.FieldType.STRING)) {
+      return FieldType.STRING;
```
vthacker, line 399: does this need to account for string being defined as `keyword` by Opensearch/ES?

kaldb/src/main/java/com/slack/kaldb/logstore/search/KaldbDistributedQueryService.java
bryanlb: This approach allows us to query historical data and get an appropriate schema set for that time period. We want to take advantage of the same logic where a single snapshot may be on multiple nodes, and be able to include schema that was available an hour ago but may not be appearing now.

kaldb/src/main/java/com/slack/kaldb/logstore/search/SearchResultUtils.java
bryanlb: This just handles serialization/deserialization of our defined `FieldType` values, which does not include a keyword type. I'm not sure we even need string/text differentiation as far as what we return out to the user, but for completeness of serialization support they're both include here.

As mentioned above, I think we probably need another field abstraction as what we return out but that will require additional schema work.

kaldb/src/main/java/com/slack/kaldb/chunk/Chunk.java
vthacker: yeah agreed! We should not expose LuceneFieldDef but FieldType needs could be extended in the future to contain more attributes.

kaldb/src/test/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiServiceTest.java
```diff
@@ -277,7 +286,32 @@ public void testEmptySearchGrafana8() throws Exception {
...
   @Test
   public void testIndexMapping() throws IOException {
-    HttpResponse response = elasticsearchApiService.mapping(Optional.of("foo"));
+    KaldbQueryServiceBase searcher = mock(KaldbQueryServiceBase.class);
+    ElasticsearchApiService serviceUnderTest = new ElasticsearchApiService(searcher);
+
+    Instant start = Instant.now();
+    Instant end = start.minusSeconds(60);
+
+    when(searcher.getSchema(
```
vthacker, line 295: super!

vthacker (approved):

PR #517, opened 2023-03-31
Do not validate older timestamp

vthacker:
Do not validate older timestamp's in documents. This gets tricky when we bootstrap a cluster and start reading messages from the start. Also if a recovery task gets scheduled after 12 hours ( in the scenario where there are lots of recovery task) this can be a problem as well

For now, let's remove this check


bryanlb (changes requested):
Let's leave the asserts in for the back-dated log messages, but just assert that they don't throw now.

bryanlb (approved):

PR #520, opened 2023-04-03
bootstrap indexers can cause no recovery tasks to get created

vthacker:
Imagine we are running the kaldb preprocessor for a few hours and then bootstrap the indexer service while setting up a new cluster.

We won't have any durable offsets (because there are no snapshots or recovery tasks). The right behaviour in this situation should be we start consuming the latest data and create recovery tasks for the data that is already in the preprocessor topic till now.

The current behaviour is it starts consuming data from the beginning and if the kafka retention is old enough and the preprocessor has been say running for several days will index data that is multiple days old.

This is just some dummy code meant to discuss if we need to fix this and then I'll go ahead and fix existing tests and add a couple of new ones if needed.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
@@ -81,6 +81,7 @@ public class RecoveryTaskCreator {
   // Get the highest offset for which data is durable for a partition.
   @VisibleForTesting
   public static long getHighestDurableOffsetForPartition(
+      long beginningOffSetForPartition,
       List<SnapshotMetadata> snapshots,
       List<RecoveryTaskMetadata> recoveryTasks,
       String partitionId) {
@@ -91,7 +92,7 @@ public class RecoveryTaskCreator {
             .filter(snapshot -> snapshot.partitionId.equals(partitionId))
             .mapToLong(snapshot -> snapshot.maxOffset)
             .max()
-            .orElse(-1);
+            .orElse(beginningOffSetForPartition);
 
     long maxRecoveryOffset =
         recoveryTasks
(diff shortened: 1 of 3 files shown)


kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
```diff
@@ -180,7 +182,7 @@ public long determineStartingOffset(long currentHeadOffsetForPartition) {
     List<RecoveryTaskMetadata> recoveryTasks = recoveryTaskMetadataStore.listSync();
     long highestDurableOffsetForPartition =
         getHighestDurableOffsetForPartition(
-            nonLiveSnapshotsForPartition, recoveryTasks, partitionId);
+            beginningOffSetForPartition, nonLiveSnapshotsForPartition, recoveryTasks, partitionId);
```
vthacker, line 185: the idea here is with this change `highestDurableOffsetForPartition` will return `beginningOffSetForPartition` when there is no metadata.

mansu:
Since Kaldb is a database, we should take data integrity seriously and index all the data. We should preserve the property in bootstrap case also. When indexers are behind we can create recovery tasks, so the indexers can index fresh data so we don't have to fall behind either. Also, kaldb doesn't have a way to re-index skipped data. So, we should avoid skipping data in all scenarios including bootstrap.

Have we considered any of the following options:
* Since Kafka is the source of durability of un-indexed data, we should solve this problem in Kafka, by deleting the old data from kafka (by temporarily setting the data retention to a very short duration to drop the data). 
* We can also set the kafka offset for consumer to head so we can only consume the data from that offset.
* Can we pre-create consumer offsets to point to latest offsets in Kafka before starting the indexers?
* Currently, we don't create recovery tasks on bootstrap. Can we force creation of recovery tasks on bootstrap?
* If we should drop the data, we should only do so upon the direction of the user. In this case, can we have a bootstrap flag to start from head instead?

PR #522, opened 2023-04-04
query parser should be aware off `_all` field was enabled or disabled

vthacker:
query parser should be aware if the `_all` field is present or not

For queries without a field name we have two options
1. Query the `_all` field
2. Make a OR query by expanding against all fields indexed

We base this decision based on the precense of the `_all` field and pass that to the open search adapter

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
@@ -99,9 +99,8 @@ public class OpenSearchAdapter {
   private final MapperService mapperService;
 
   private final Map<String, LuceneFieldDef> chunkSchema;
-  private final boolean enableFullTextSearch;
 
-  public OpenSearchAdapter(Map<String, LuceneFieldDef> chunkSchema, boolean enableFullTextSearch) {
+  public OpenSearchAdapter(Map<String, LuceneFieldDef> chunkSchema) {
     IndexSettings indexSettings = buildIndexSettings();
     SimilarityService similarityService = new SimilarityService(indexSettings, null, emptyMap());
     this.mapperService = buildMapperService(indexSettings, similarityService);
@@ -109,7 +108,6 @@ public class OpenSearchAdapter {
         buildQueryShardContext(
             KaldbBigArrays.getInstance(), indexSettings, similarityService, mapperService);
     this.chunkSchema = chunkSchema;
-    this.enableFullTextSearch = enableFullTextSearch;
   }
 
   /**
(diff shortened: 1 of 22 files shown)


kaldb/src/test/java/com/slack/kaldb/logstore/schema/ConvertFieldValueAndDuplicateTest.java
```diff
@@ -102,8 +102,9 @@ public void testListTypeInDocumentWithoutFullTextSearch() throws IOException {
         build(CONVERT_VALUE_AND_DUPLICATE_FIELD, false, meterRegistry);
     assertThat(docBuilder.getIndexFieldConflictPolicy())
         .isEqualTo(CONVERT_VALUE_AND_DUPLICATE_FIELD);
-    assertThat(docBuilder.getSchema().size()).isEqualTo(17);
-    assertThat(docBuilder.getSchema().keySet()).contains(LogMessage.SystemField.ALL.fieldName);
+    assertThat(docBuilder.getSchema().size()).isEqualTo(16);
+    assertThat(docBuilder.getSchema().keySet())
```
vthacker, line 106: two lines above where we create `SchemaAwareLogDocumentBuilderImpl` we pass false for enableFullTextSearch

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -142,13 +144,16 @@ public Query buildQuery(
           && !queryStr.equals("*")) {
         QueryStringQueryBuilder queryStringQueryBuilder = new QueryStringQueryBuilder(queryStr);
 
-        // the field to use when none are provided, otherwise will OR all fields in the registry (up
-        // to a configurable field limit)
-        queryStringQueryBuilder.defaultField("_all");
+        if (queryShardContext.getMapperService().fieldType("_all") != null) {
+          queryStringQueryBuilder.defaultField("_all");
+          // setting lenient=false will not throw error when the query fails to parse against
+          // numeric
```
bryanlb, line 150: This comment looks like it got messed up by the maven formatter.

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -142,13 +144,16 @@ public Query buildQuery(
           && !queryStr.equals("*")) {
         QueryStringQueryBuilder queryStringQueryBuilder = new QueryStringQueryBuilder(queryStr);
 
-        // the field to use when none are provided, otherwise will OR all fields in the registry (up
-        // to a configurable field limit)
-        queryStringQueryBuilder.defaultField("_all");
+        if (queryShardContext.getMapperService().fieldType("_all") != null) {
+          queryStringQueryBuilder.defaultField("_all");
```
bryanlb, line 148: Do we have a static somewhere that references `_all`? I know it wasn't that way in the existing code, but wondering about it now. 🤔

bryanlb (approved):
Approved, with a few optional suggestions.

bryanlb (approved):

PR #524, opened 2023-04-05
nocommit

vthacker:


kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -196,7 +202,8 @@ public void reloadSchema() {
               entry.getValue().name);
         }
       } catch (Exception e) {
-        LOG.error("Error parsing schema mapping for {}", entry.getValue().toString(), e);
+        throw new IllegalStateException(
```
bryanlb, line 205: I'm not sure this is what we want to do here? 

We need to provide protection against registering a field that already exists - this is a race condition that is expected to occur when new fields are being added.

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -379,15 +389,11 @@ private static boolean tryRegisterField(
...
-      try {
-        XContentBuilder mapping = fieldMapping(fieldName, buildField);
-        mapperService.merge(
-            "_doc",
-            new CompressedXContent(BytesReference.bytes(mapping)),
-            MapperService.MergeReason.MAPPING_UPDATE);
-      } catch (Exception e) {
-        LOG.error("Error doing map update", e);
-      }
+      XContentBuilder mapping = fieldMapping(fieldName, buildField);
```
bryanlb, line 392: This needs to catch the case where you attempt to register the same field multiple times.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -387,6 +396,7 @@ private static boolean tryRegisterField(
             MapperService.MergeReason.MAPPING_UPDATE);
       } catch (Exception e) {
         LOG.error("Error doing map update", e);
+        return false;
```
bryanlb, line 399: Whoops, missed that the first go-around.

kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -96,10 +97,15 @@ public class OpenSearchAdapter {
 
   private final QueryShardContext queryShardContext;
 
-  private final MapperService mapperService;
+  @VisibleForTesting protected final MapperService mapperService;
```
bryanlb, line 100: One thought here is just to change the `tryRegisterField` to not be static, and don't pass the mapper service in each call. That'd actually cleanup the `reloadSchema` method a bit, and then you wouldn't need to make the mapper service visible. You could just have the `tryRegisterField` as a protected.

kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/FieldLimitTest.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @AfterClass
+  public static void resetTotalFieldsLimitAfterTestRuns() {
+    OpenSearchAdapter.TOTAL_FIELDS_LIMIT = 5000;
+  }
+
+  @Test
+  public void checkIfMaxFieldLimitsIsEnforced() {
+    OpenSearchAdapter.TOTAL_FIELDS_LIMIT = 1;
```
bryanlb, line 28: Not sure I'm super keen on overriding this - making the `TOTAL_FIELDS_LIMIT` a static would be better. For purposes of the test you can just for loop and try to create 5001 text strings, and then you wouldn't need to mutate or reset this value. That also has a benefit of catching any other potential issues with registering / storing 5k fields we may not be aware of.

bryanlb (changes requested):

kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/FieldLimitTest.java
vthacker: that is a good idea!

kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/FieldLimitTest.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import org.junit.Test;
+
+public class FieldLimitTest {
+
+  @Test
+  public void checkIfMaxFieldLimitsIsEnforced() {
+    Map<String, LuceneFieldDef> fieldDefMap = getChunkSchema(5001);
+    OpenSearchAdapter openSearchAdapter = new OpenSearchAdapter(fieldDefMap);
+    assertThat(openSearchAdapter.tryRegisterFields()).isFalse();
```
bryanlb, line 16: We may want to slightly change this behavior, such that the first 5k are allowed (and still returns false). Then the following attempt to register 2 would fail because we already have 5k.

If we think of the situation where a cache node has 5001 fields, it would never register any of them instead of just dropping one that's over the limit, which I'm not sure is what we'd want to do.

vthacker: We could, however this `tryRegisterFields` only registers new fields - So in real life within one commit refresh we'd need to have 5k new fields get added?

bryanlb: This would be expected of a cache node, where it will have a single commit with all the fields populated at the initial load.

vthacker: So should we build our logic within `tryRegisterFields` to see if the number of new fields is not beyond X?

bryanlb: If `tryRegisterFields` is synchronized, we should be able to check the current length and then just attempt to insert however many is safe to add.

vthacker:
```
      CompressedXContent content = new CompressedXContent(BytesReference.bytes(createMappings(Map.of("_kaldb", b -> b.field("type", "keyword")))));
      DocumentMapper documentMapper = mapperService.documentMapperParser().parse("_doc", content);
      MappingLookup mappers = documentMapper.mappers();
      AtomicInteger count = new AtomicInteger();
      mappers.iterator().forEachRemaining(c -> count.getAndIncrement());
```

I figured if we do this, we can get the current count of the field defined. This involved staring at the internals quite a bit

Should we add this and then see if  `newFields.length + count> limit` ? 

Or did your comment just refer to ensuring `newFields.length > limit` ?

bryanlb: After looking a bit deeper into this, I think we should probably roll with what you have here in this PR and we may circle back on this later. This seems to be a bit more involved than we probably want to do right now, especially considering that it may change with our schema refactoring.

bryanlb (approved):

PR #525, opened 2023-04-10
Add index type to recovery task and node.

mansu:
Add index type field to `RecoveryTaskmetadata`. Added supported index types field to `RecoveryNodeMetadata`. Updated related code and tests.

I have also refactored the code to centralize the recovery task update code in `RecoveryNodeMetadataStore`. This simplifies any errors in copying attributes during state transitions. In future, if any new fields are added, there is one place in code to make those code changes.


bryanlb (changes requested):
After reviewing this PR in light of our current objectives, I believe this effort doesn't align well with our immediate priorities. We do not expect to support another index type other than Lucene in the near future. Additionally, we would like to have a larger discussion around what the upgrade path looks like for a major version of Lucene before further committing to this pattern.

We will leave the existing index type code in the repo that has already been merged (see #500), and this is linked to that previous effort now as well for future reference.

PR #526, opened 2023-04-10
Added min aggregator

kyle-sammons:
Adds the `min` aggregator to our suite of aggregators. There's also some formatting attached that happened automatically when I did a `mvn clean compile` to regenerate some of the protos.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapterTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapterTest.java
@@ -15,7 +15,6 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import org.apache.lucene.search.CollectorManager;
-import org.apache.lucene.search.MatchNoDocsQuery;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


kaldb/src/test/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapterTest.java
```diff
@@ -44,7 +47,9 @@ public String getType() {
         };
 
     openSearchAdapter.buildAggregatorUsingContext(
-        unknownAgg, logStoreAndSearcherRule.logStore.getSearcherManager().acquire());
+        unknownAgg,
+        logStoreAndSearcherRule.logStore.getSearcherManager().acquire(),
+        new MatchNoDocsQuery("testing"));
```
bryanlb, line 52: You could also pass `null` in these calls as long as the test isn't attempting a min/max agg.

bryanlb (changes requested):
The only thing I'd like to see changed on this is one extra test in the `LogIndexSearcherImpl` for a min agg - you can follow the example of the `testFullIndexSearch`. Do a query, pass a min agg, and then cast the internal agg to an InternalMin.

bryanlb (approved):

PR #529, opened 2023-04-12
Added the max aggregation

kyle-sammons:
Adds the max aggregation, tests for it, and also some formatting the got picked up in `mvn compile`. This implementation mirrors the min aggregation (and leads me to believe that we could likely abstract this out to a single implementation if we wanted to in the future).


bryanlb:
Rebased after merging the script support changes.

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
```diff
@@ -191,8 +192,21 @@ private static List<KaldbSearch.SearchRequest.SearchAggregation> getRecursive(Js
                                   KaldbSearch.SearchRequest.SearchAggregation.ValueSourceAggregation
                                       .newBuilder()
                                       .setField(getFieldName(min))
+                                      .setScript(SearchResultUtils.toValueProto(getScript(min)))
```
bryanlb, line 195: I missed a few spots setting the script on the min aggregator, so those are included in this PR as well.

vthacker (approved):

PR #541, opened 2023-05-03
Increase the number of mapping fields allowed by Opensearch

bryanlb:
This caused an issue with the cache nodes no longer returning results, and will need to be revised prior to merging again.

Reverts slackhq/kaldb#540


bryanlb:
As part of this PR we should attempt to combine the efforts of #535 and #534 (if necessary).

PR #543, opened 2023-05-11
Zookeeper/Metadata refactor

bryanlb:
This refactors our existing zookeeper/metadata code, replacing significant amounts of custom code in favor of using strongly typed curator models (see [modeled curator documentation](https://curator.apache.org/curator-x-async/modeled-components.html)). This also reduces several abstractions in preparation for new sub-pathing logic to be introduced in a later PR.

Additional items of note:
* Updates curator to 5.5 (latest)
* Addresses race condition todo around ZK change notifications
* Addresses previous case where cache could go stale on some ZK connection issues
* Moves uncached list api to use `childrenAsZNodes`, which provides better concurrency and error handling then the previous list method. This also addresses several unresolved todos within this method around known issues.


kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -163,14 +163,20 @@ public static SearchMetadata registerSearchMetadata(
                 snapshotName, cacheSearchContext.hostname),
             snapshotName,
             cacheSearchContext.toUrl());
-    searchMetadataStore.create(metadata).get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
+    searchMetadataStore
```
vthacker, line 166: can we document why this doesn't use createSync? I see the timeout is 5s vs the default timeout in createSync being 15. Maybe we can make that clear as to why we need a smaller timeout for this operation?

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -1,176 +1,200 @@
...
     try {
-      metadataStore.create(storeFolder, "", true).get();
-    } catch (ExecutionException exception) {
-      //noinspection StatementWithEmptyBody
-      if (exception.getCause() instanceof NodeExistsException) {
-        // ignore exception, since node creation is idemponent.
-      } else {
-        throw exception;
-      }
+      createAsync(metadataNode)
```
vthacker, line 79: could this be just createSync? timeout seems to be the same?

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -1,176 +1,200 @@
...
- * node.
+ * <p>Every method provides an async and a sync API. In general, use the async API you are
+ * performing batch operations and a sync if you are performing a synchronous operation on a node.
  */
-abstract class KaldbMetadataStore<T extends KaldbMetadata> {
-  protected final MetadataStore metadataStore;
-  protected final String storeFolder;
-  protected final MetadataSerializer<T> metadataSerializer;
-  protected final Logger logger;
+public class KaldbMetadataStore<T extends KaldbMetadata> implements Closeable {
```
vthacker, line 33: I think my biggest feedback from this class is, can this class return CompletableFuture instead of AsyncStage for the async operations? that way we still keep the ZK exposure minimum outside of this class?

kaldb/src/main/java/com/slack/kaldb/metadata/dataset/DatasetMetadataStore.java
```diff
@@ -1,23 +1,19 @@
...
     super(
+        curator,
+        CreateMode.PERSISTENT,
         shouldCache,
-        true,
-        DATASET_METADATA_STORE_ZK_PATH,
-        metadataStore,
-        new DatasetMetadataSerializer(),
-        LOG);
+        new DatasetMetadataSerializer().toModelSerializer(),
```
vthacker, line 16: this could just be a static?

kaldb/src/main/java/com/slack/kaldb/metadata/search/SearchMetadataStore.java
```diff
@@ -1,22 +1,26 @@
...
-        false,
-        SEARCH_METADATA_STORE_ZK_PATH,
-        metadataStore,
-        new SearchMetadataSerializer(),
-        LOG);
+        new SearchMetadataSerializer().toModelSerializer(),
+        SEARCH_METADATA_STORE_ZK_PATH);
+  }
+
+  @Override
```
vthacker, line 22: do we also need to override updateSync?

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
bryanlb: I think that we could return a `CompletionStage` here instead of an `AsyncStage` - that way it's part of the java concurrent package, isn't tied to Curator (like AsyncStage), and doesn't requiring converting these all to `CompleteableFutures`, since `AsyncStage` just extends `CompletionStage`.

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
bryanlb: I went back and looked at this, and it's not clear why we use `TIMEOUT_MS` throughout this class. I do see that it appears `DEFAULT_ZK_TIMEOUT_SECS` was added a week later, so this may be just the first implementation?

I added a deprecation notice to this variable, explaining that we should be switching to the sync methods as possible.

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
bryanlb: Made these all return `CompletionStage.`

vthacker (approved):

PR #548, opened 2023-05-17
use text blocks to make some of the JSON strings readable

vthacker:
Leverage the new text blocks feature available to us in java 17


benchmarks/src/main/java/com/slack/kaldb/IndexingBenchmark.java
```diff
@@ -59,7 +59,65 @@ public void createIndexer() throws Exception {
...
+        """
```
vthacker, line 62: verified by moving this to a separate test that we are able to parse the new JSON

bryanlb (approved):

PR #552, opened 2023-05-22
Force chunk rollover to prevent race condition in test

bryanlb:
###  Summary
Addresses an issue with the `IndexingChunkManagerTest.testFailedRollOverFinishesOnClose` test where due to rollover timing a test could intermittently fail, depending on the speed of the test env.


bryanlb:
This appears to help the issue, but not entirely. Closing, and will revisit after #543 is merged if it is still an issue with the new ZK implementation.

PR #553, opened 2023-05-31
ZK Partitioning Store

bryanlb:
###  Summary

Introduces a new ZK partitioning store. This works on the principle that the metadata classes will define a partitioning logic specific to their data, and then the new partitioning store will handle mapping these partitions to separate caches. This results in a slightly higher [[1]](#known-tradeoffs) memory pressure in favor of enabling scalability. For stores not operating on znodes by path this is an entirely transparent swap.

#### Example

SnapshotMetadata used to be stored to `/snapshot/{name}`. We now implement the following partitioning method:

```java
  @Override
  public String getPartition() {
    if (isLive(this)) {
      return "LIVE";
    } else {
      ZonedDateTime snapshotTime = Instant.ofEpochMilli(startTimeEpochMs).atZone(ZoneOffset.UTC);
      return String.format(
          "%s_%s",
          snapshotTime.getLong(ChronoField.EPOCH_DAY),
          snapshotTime.getLong(ChronoField.HOUR_OF_DAY));
    }
  }
```

This results in partitions that look like `/snapshot/19508_0/{name1}`, where the first portion is the days since epoch, and the second portion is the current hour of the day (0 - 23). For metadata that doesn't contain date information (for example, cache slots) a different partitioning logic can be used (ie, `/cacheSlot/{hostname}/{slotId}`).

#### Implementation

Switching a specific metadata store to use the partitioning should need the following steps performed:

1. Change the target metadata to extend a `KaldbPartionedMetadata` and implement the new `getPartition()` method
2. Change the metadata store to extend a `KaldbPartitioningMetadataStore`, making changes as required (constructor, some methods).
3. (_optional - recommended_) Change the zk path to something different, such as `/partitioned_{foo}`. This is because the new store is not backwards compatible with the existing ZK data.

#### Known tradeoffs
1. This will result in higher memory pressure. The entire dataset will be cached in memory regardless if the partitioning logic is used or not, but extra overhead comes from the separate modeled frameworks, and associated listeners.
2. This removes the strict guarantee preventing duplicate node names. Since these can exist in separate paths now, we do not atomically ensure that the sibling paths do not contain a matching node ID.

#### References
https://github.com/slackhq/kaldb/pull/543
https://github.com/slackhq/kaldb/pull/566
https://github.com/slackhq/kaldb/pull/562
https://github.com/slackhq/kaldb/pull/564
https://github.com/slackhq/kaldb/pull/567
https://github.com/slackhq/kaldb/pull/570

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
@@ -26,6 +26,7 @@ import java.nio.file.Path;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -60,10 +61,8 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
   private final String dataDirectoryPrefix;
   private final String s3Bucket;
   protected final SearchContext searchContext;
-  protected final String slotName;
-  private final String slotId;
+  protected final String slotId;
   private final CacheSlotMetadataStore cacheSlotMetadataStore;
-  private final CacheSlotMetadataStore cacheSlotListenerMetadataStore;
   private final ReplicaMetadataStore replicaMetadataStore;
   private final SnapshotMetadataStore snapshotMetadataStore;
   private final SearchMetadataStore searchMetadataStore;
@@ -103,7 +102,6 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
     this.executorService = executorService;
     this.searchContext = searchContext;
     this.slotId = UUID.randomUUID().toString();
-    this.slotName = String.format("%s-%s", searchContext.hostname, slotId);
 
     this.cacheSlotMetadataStore = cacheSlotMetadataStore;
     this.replicaMetadataStore = replicaMetadataStore;
@@ -112,19 +110,16 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
 
     CacheSlotMetadata cacheSlotMetadata =
         new CacheSlotMetadata(
-            slotName,
+            slotId,
             Metadata.CacheSlotMetadata.CacheSlotState.FREE,
             "",
             Instant.now().toEpochMilli(),
             List.of(Metadata.IndexType.LOGS_LUCENE9),
             searchContext.hostname);
     cacheSlotMetadataStore.createSync(cacheSlotMetadata);
+    cacheSlotMetadataStore.addListener(cacheSlotListener);
 
-    this.cacheSlotListenerMetadataStore =
-        new CacheSlotMetadataStore(curatorFramework, slotName, true);
-    cacheSlotListenerMetadataStore.addListener(cacheSlotListener);
     cacheSlotLastKnownState = Metadata.CacheSlotMetadata.CacheSlotState.FREE;
-
     chunkAssignmentTimerSuccess = meterRegistry.timer(CHUNK_ASSIGNMENT_TIMER, "successful", "true");
     chunkAssignmentTimerFailure =
         meterRegistry.timer(CHUNK_ASSIGNMENT_TIMER, "successful", "false");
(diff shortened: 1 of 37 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -98,6 +99,22 @@ public T getSync(String path) {
     }
   }
 
+  public CompletionStage<Stat> hasAsync(String path) {
+    if (cachedModeledFramework != null) {
+      return cachedModeledFramework.withPath(zPath.resolved(path)).checkExists();
```
vthacker, line 104: is there an advantage of using the cacheModelFramework if it's present? i.e can `hasAsync` just use modeledClient?

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStore.java
```diff
@@ -0,0 +1,249 @@
...
+    } catch (ExecutionException | InterruptedException | TimeoutException e) {
+      throw new InternalMetadataStoreException("Error listing nodes", e);
+    }
+  }
+
+  private KaldbMetadataStore<T> getOrCreateMetadataStore(String partition) {
+    return metadataStoreMap.computeIfAbsent(
+        partition,
+        (p1) -> {
+          LOG.info("Creating new metadata store for partition - {}", partition);
```
vthacker, line 204: will having the `path` to the output here be useful?

kaldb/src/main/java/com/slack/kaldb/metadata/core/MetadataSerializer.java
```diff
@@ -23,6 +23,10 @@ default ModelSerializer<T> toModelSerializer() {
     return new ModelSerializer<>() {
       @Override
       public byte[] serialize(T model) {
```
vthacker, line 25: not related to this change, but perhaps the variable `model` is confusing? Especially now that we have modelClient? Would `metadata` be a better name here?

kaldb/src/main/java/com/slack/kaldb/chunk/ReadOnlyChunkImpl.java
```diff
@@ -128,28 +126,29 @@ public ReadOnlyChunkImpl(
...
-              "Unexpected state transition from {} to {}", cacheSlotLastKnownState, newSlotState);
-        }
-        executorService.execute(() -> handleChunkAssignment(cacheSlotMetadata));
-      } else if (newSlotState.equals(Metadata.CacheSlotMetadata.CacheSlotState.EVICT)) {
-        LOG.info("Chunk - EVICT received");
-        if (!cacheSlotLastKnownState.equals(Metadata.CacheSlotMetadata.CacheSlotState.LIVE)) {
-          LOG.warn(
-              "Unexpected state transition from {} to {}", cacheSlotLastKnownState, newSlotState);
+  private KaldbMetadataStoreChangeListener<CacheSlotMetadata> cacheNodeListener() {
+    return (metadata) -> {
```
vthacker, line 130: nice! So we don't even need to call `getSync(slotId/slotName)` in this impl. 

We listen to all events changed and if the event matches the one we care about we execute some code.

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
bryanlb: The benefits here would be consistency (ie, you would not have an issue where `has` returns a different result than `get` or `list`), and performance (can just check the cache).

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStore.java
bryanlb: Good idea! Added.

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStoreTest.java
```diff
@@ -0,0 +1,539 @@
...
+    this.curatorFramework = CuratorBuilder.build(meterRegistry, zookeeperConfig);
+  }
+
+  @AfterEach
+  public void tearDown() throws IOException {
+    curatorFramework.unwrap().close();
+    testingServer.close();
+    meterRegistry.close();
+
+    System.setProperty(ZNODE_CONTAINER_CHECK_INTERVAL_MS, String.valueOf(checkInterval));
```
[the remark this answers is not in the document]
bryanlb: Good call - I think we might be able to do this in the constructor actually, immediately after instantiating the class as I don't believe it's used again.

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStore.java
```diff
@@ -0,0 +1,303 @@
...
+    return getOrCreateMetadataStore(findPartition(path)).getAsync(path);
+  }
+
+  /**
+   * Attempts to find the metadata without knowledge of the partition it exists in. Use of this
+   * should be avoided if possible, preferring the getSync.
+   *
+   * @see KaldbPartitioningMetadataStore#getSync(String, String)
+   */
+  public T findSync(String path) {
```
vthacker, line 181: what do you think of just putting this a a test util instead of exposing it here?

bryanlb: This is used in our production code path today as part of the `ReadOnlyChunkImpl.handleChunkAssignment`. In this case we have a `replicaId` which we need to first find, and then upon getting the `replicaId` we try to find the snapshot. In both of these cases we don't know the partition ahead of time.

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStoreTest.java
```diff
@@ -0,0 +1,538 @@
...
+      metadata1.setExtraField("val2");
+      store.updateSync(metadata1);
+
+      await().until(() -> counter.get() == 2);
+
+      store.removeListener(listener);
+      metadata1.setExtraField("val3");
+      store.updateSync(metadata1);
+
+      Thread.sleep(2000);
```
vthacker, line 534: you wanted to add this so that the next line assert can be added with more confidence? i.e in a scenario where we actually don't remove listener but the lag in the listener triggering might cause the test to pass but it shouldn't?

vthacker (approved):

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbPartitioningMetadataStoreTest.java
bryanlb: Fixed!

bryanlb: Updated this with a comment for the future, but you are correct. I didn't want to have an assert immediately after, as it could just be passing due to the test running fast.

PR #561, opened 2023-06-06
Improve test logging

bryanlb:
###  Summary

* Adds color highlighting to build logs
* Adds better logging when specific test methods are started and finished
* Suppresses an unnecessary log message generated by Apache Curator

Spun out of #553

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -26,8 +26,8 @@ jobs:
         distribution: 'temurin'
         cache: 'maven'
     - name: Ensure code is formatted
-      run: mvn com.spotify.fmt:fmt-maven-plugin:check --file kaldb/pom.xml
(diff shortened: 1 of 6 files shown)


.github/workflows/maven.yml
```diff
@@ -26,8 +26,8 @@ jobs:
         distribution: 'temurin'
         cache: 'maven'
     - name: Ensure code is formatted
-      run: mvn com.spotify.fmt:fmt-maven-plugin:check --file kaldb/pom.xml
+      run: mvn -B -Dstyle.color=always com.spotify.fmt:fmt-maven-plugin:check --file kaldb/pom.xml
```
vthacker, line 29: fancy!

vthacker (approved):

.github/workflows/maven.yml
bryanlb: For any Jenkins jobs that are invoked with this we probably should make sure we're doing the batch mode (`-B`) to prevent those progress messages from showing up.

PR #562, opened 2023-06-06
Add hostname to cache slot metadata

bryanlb:
###  Summary

Adds hostname to the cache slot metadata. This will enable two future use cases, including zookeeper partitioning and better slot/replica assignments.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadata.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadata.java
@@ -12,6 +12,7 @@ import java.util.List;
  * Make transitions more controlled via a state machine like API.
  */
 public class CacheSlotMetadata extends KaldbMetadata {
+  public final String hostname;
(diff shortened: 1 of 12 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadata.java
```diff
@@ -22,8 +23,10 @@ public CacheSlotMetadata(
       Metadata.CacheSlotMetadata.CacheSlotState cacheSlotState,
       String replicaId,
       long updatedTimeEpochMs,
-      List<Metadata.IndexType> supportedIndexTypes) {
+      List<Metadata.IndexType> supportedIndexTypes,
+      String hostname) {
     super(name);
+    checkArgument(hostname != null && !hostname.isEmpty(), "Hostname cannot be null or empty");
```
bryanlb, line 29: As cache slots are ephemeral, this should be safe to enforce in this PR as the code deploying will re-create the slots with the newly required hostname.

vthacker (approved):

PR #564, opened 2023-06-06
Refactor metadata store lifecycle to closeable lifecycle

bryanlb:
###  Summary

Refactors the signature of the existing metadata store lifecycle manager to work for any closable class. The existing approach did not specifically need a `KaldbMetadataStore` type, and this allows for pushing any closable into this list, which will be needed for #553.


kaldb/src/main/java/com/slack/kaldb/metadata/core/CloseableLifecycleManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  protected void shutDown() throws Exception {
+    LOG.info("Shutting down CloseableLifecycleManager for role - {}", nodeRole.toString());
+    closeableList.forEach(
+        closeable -> {
+          try {
+            closeable.close();
+          } catch (IOException e) {
+            throw new RuntimeException(e);
```
vthacker, line 39: Let's say we throw an IOException on the first closeable in the list. Should we still attempt to close others and then throw exception?

bryanlb: Good idea. I modified this so that it now only logs the exception, but continues with the shutdown process. It also includes a test for this behavior as well.

vthacker (approved):

PR #566, opened 2023-06-06
Rename metadata method, add deprecation warning

bryanlb:
###  Summary

This refactors the existing metadata store with the following method name changes:

* `getCachedSync` -> `listSync`
* `getCachedAsync` > `listAsync`
* `listSync` -> `listSyncUncached`
* `listAsync` -> `listAsyncUncached`

This also makes access to `listAsyncUncached` `private` now, and adds a deprecation notice to the `listSyncUncached`. Moving forward we will be enforcing cached-only access to the list methods.


vthacker (approved):
Thanks for breaking up the work!

PR #574, opened 2023-06-14
Delete successful chunks on recovery complete

bryanlb:
###  Summary

Addresses an issue where recovery nodes never delete the recovered chunks, resulting in out of disk errors


bryanlb:
Superseded by #690

PR #575, opened 2023-06-15
Increase cpu allocated for Lucene merge operations

bryanlb:
###  Summary

The default cpu utilization for merge operations is rather conservative, and can be increased. This moves to a `cores - 1` max, where the default is the larger of - 4, or half of the cores.


bryanlb:
Closing this as a won't do for now. While this does improve the hardware utilization, it doesn't help the significant amount of mostly idle time during S3 uploads. To address that we may look at either allowing recovery to pickup indexing the next task while still uploading to S3, or running two jobs in parallel.

Some other approaches for improving recovery speed are being addressed in #577, which should lower the overall time spent merging.

PR #577, opened 2023-06-16
make larger segments

vthacker:
###  Summary

Increase the default maxRamBufferSize for lucene. This will improve indexing performance for recovery nodes as they don't receive any search traffic when it's being built.

Indexer nodes could also gain some improvements but the with soft commits flushing segments it might not help much?

Also based on the ramBufferSize decide if we should use compound files or not.

> For best indexing speed you should flush by RAM usage with a large RAM buffer.  

From https://lucene.apache.org/core/7_0_1/core/org/apache/lucene/index/IndexWriter.html

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
@@ -62,6 +62,8 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
   private final io.micrometer.core.instrument.Timer commitsTimer;
   private final io.micrometer.core.instrument.Timer refreshesTimer;
 
+  private final String MAX_RAM_BUFFER_SIZE_MB = "maxRamBufferSizeMb";
+
   // TODO: Set the policy via a lucene config file.
   public static LuceneIndexStoreImpl makeLogStore(
       File dataDirectory, KaldbConfigs.LuceneConfig luceneConfig, MeterRegistry metricsRegistry)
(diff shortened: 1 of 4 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -62,6 +62,8 @@ public class LuceneIndexStoreImpl implements LogStore<LogMessage> {
   private final io.micrometer.core.instrument.Timer commitsTimer;
   private final io.micrometer.core.instrument.Timer refreshesTimer;
 
+  private final String MAX_RAM_BUFFER_SIZE_MB = "maxRamBufferSizeMb";
```
bryanlb, line 65: This seems unused?

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -162,6 +186,12 @@ private IndexWriterConfig buildIndexWriterConfig(
                         true)))
             .setIndexDeletionPolicy(snapshotDeletionPolicy);
 
+    // See
+    // https://lucene.apache.org/core/9_5_0/core/org/apache/lucene/index/IndexWriterConfig.html#setUseCompoundFile(boolean)
+    if (ramBufferSizeMb >= 128) {
```
bryanlb, line 191: Should this instead be checking the `useCFSFiles` boolean?

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -143,15 +145,37 @@ public void run() {
...
+    return targetBufferSize;
+  }
+
   private IndexWriterConfig buildIndexWriterConfig(
       Analyzer analyzer,
       SnapshotDeletionPolicy snapshotDeletionPolicy,
       LuceneIndexStoreConfig config,
       MeterRegistry metricsRegistry) {
+    long ramBufferSizeMb = getRAMBufferSizeMB(Runtime.getRuntime().maxMemory());
+    boolean useCFSFiles = ramBufferSizeMb <= 128;
```
bryanlb, line 172: Let's do a static final for the ram buffer size along with a comment why 128 is a good value to use.

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -143,15 +145,37 @@ public void run() {
...
+   * 10% matches that of the defaults of ES.
+   *
+   * @see {https://www.elastic.co/guide/en/elasticsearch/reference/current/indexing-buffer.html}
+   */
+  protected static long getRAMBufferSizeMB(long heapMaxBytes) {
+    long targetBufferSize = 256;
+    if (heapMaxBytes != Long.MAX_VALUE) {
+      targetBufferSize = Math.min(2048, Math.round(heapMaxBytes / 1e6 * 0.10));
+    }
+    LOG.warn(
```
bryanlb, line 159: Probably should be an info here instead of a warn.

bryanlb (changes requested):

bryanlb (approved):

PR #578, opened 2023-06-16
Dynamically allocate Lucene RAM buffer based on heap

bryanlb:
###  Summary

Dynamically allocate a larger ram buffer to Lucene, based on the heap size. Uses 10% of the heap as the target, maxing out at 2048 (see [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java#L238-L239), and [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java#LL136C46-L136C82) for why 2048). If heap is not set uses default of 256.

Original default is 16mb if not provided, which is significantly too many merges for our use case. Using this we would expect to set the heap on the recovery nodes to ~20gb to ensure we max out.


bryanlb:
Closing as we'll be addressing this in #577. cc @vthacker

PR #580, opened 2023-06-19
Add better preprocessor stream error handling

bryanlb:
###  Summary

There exists a scenario today where a stream application can shutdown due to an error, but this isn't reported back to Guava. When this happens the preprocessor continues to run but isn't consuming any messages. This PR adds an additional detection for this scenario and notifies Guava that the service has failed.

This is a somewhat short-term solution, as we expect to be removing stream processing in favor of http bulk ingest within the next few weeks.

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/preprocessor/PreprocessorServiceIntegrationTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/preprocessor/PreprocessorServiceIntegrationTest.java
@@ -4,6 +4,7 @@ import static com.slack.kaldb.server.KaldbConfig.DEFAULT_START_STOP_DURATION;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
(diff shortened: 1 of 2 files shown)


vthacker (approved):

kaldb/src/test/java/com/slack/kaldb/preprocessor/PreprocessorServiceIntegrationTest.java
```diff
@@ -50,6 +52,70 @@ public void teardown() throws Exception {
...
+
+    preprocessorService.startAsync();
+    preprocessorService.awaitRunning(DEFAULT_START_STOP_DURATION);
+    assertThat(MetricsUtil.getTimerCount(PreprocessorService.CONFIG_RELOAD_TIMER, meterRegistry))
+        .isEqualTo(1);
+
+    // restarting ZK should cause a stream application error due to missing source topics
+    zkServer.restart();
+
+    await()
```
vthacker, line 112: nice!

PR #581, opened 2023-06-19
Expose merge throttling parameter

vthacker:
###  Summary

On the recovery nodes we don't want to throttle merges. Throttling is useful when we want to balance indexing vs search performance.

However recovery nodes are not searchable while the recovery service is creating chunks. So we should be setting very high commit and refresh intervals ( which we have started doing last week ) and not throttling merges.

Another PR is tacking increasing the ram buffer size which also saw a significant improvement in overall throughput.

### Requirements


bryanlb:
See #624

PR #587, opened 2023-06-26
Add index searcher reference for opensearch query optimization

bryanlb:
###  Summary

Refactors code such that the opensearch adapter can make use of the index searcher for query optimizations.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
@@ -105,14 +105,14 @@ public class LogIndexSearcherImpl implements LogIndexSearcher<LogMessage> {
 
     Stopwatch elapsedTime = Stopwatch.createStarted();
     try {
-      Query query =
-          openSearchAdapter.buildQuery(dataset, queryStr, startTimeMsEpoch, endTimeMsEpoch);
-      span.tag("lucene_query", query.toString());
-
       // Acquire an index searcher from searcher manager.
       // This is a useful optimization for indexes that are static.
-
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -105,14 +105,14 @@ public SearchResult<LogMessage> search(
...
-      Query query =
-          openSearchAdapter.buildQuery(dataset, queryStr, startTimeMsEpoch, endTimeMsEpoch);
-      span.tag("lucene_query", query.toString());
-
       // Acquire an index searcher from searcher manager.
       // This is a useful optimization for indexes that are static.
-
       IndexSearcher searcher = searcherManager.acquire();
+
+      Query query =
```
vthacker, line 112: Can we change this to
```

      try {
        Query query =
                openSearchAdapter.buildQuery(
                        dataset, queryStr, startTimeMsEpoch, endTimeMsEpoch, searcher);
        span.tag("lucene_query", query.toString());
      } catch (Exception e) {
        span.tag("lucene_query_parse_exception", "true");
      }
```

We'll be able to see in the future what queries were we unable to parse?

vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcherImpl.java
bryanlb: These should already have exceptions for failure to parse, along with stack trace so I'm not sure having it on a trace is the best place for that? The overall trace should be marked as failed due to the exception as well.

PR #589, opened 2023-06-27
Increase total fields limit

bryanlb:
###  Summary

Increases the total field limit from 2500 to 3500, to provide some additional headroom. This still needs to be properly addressed, but provides a bit more time to make that happen.


vthacker (approved):

bryanlb:
Closing, we can address in #541

PR #591, opened 2023-06-27
pass index sort to the open search adapter

vthacker:
###  Summary

We want to pass the index sort to the open search adapter. That way it is aware of it and when it rewrites the query is able to leverage this info

Screenshot 2023-06-27 at 4 51 36 PM

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
+++ b/kaldb/src/main/java/com/slack/kaldb/logstore/opensearch/OpenSearchAdapter.java
@@ -330,6 +330,13 @@ public class OpenSearchAdapter {
             .put(IndexMetadata.SETTING_VERSION_CREATED, Version.V_2_7_0)
             .put(
                 MapperService.INDEX_MAPPING_TOTAL_FIELDS_LIMIT_SETTING.getKey(), TOTAL_FIELDS_LIMIT)
+
+            // Kaldb time sorts the indexes while building it
+            // {LuceneIndexStoreImpl#buildIndexWriterConfig}
+            // When we were using the lucene query parser the sort info was leveraged by lucene
+            // automatically ( as the sort info persists in the segment info ) at query time.
+            // However the OpenSearch query parser has a custom implementation which relies on the
+            // index sort info to be present as a setting here.
             .put("index.sort.field", LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName)
             .put("index.sort.order", "desc")
             .build();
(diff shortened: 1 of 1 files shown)


bryanlb (changes requested):
Let's also add comments that link this to the `LuceneIndexStoreImpl.buildIndexWriterConfig` where we set the sort - just in case something were to change in the future we'd be certain to get it updated in both places.

bryanlb (approved):

fxchen:
Prototype (from my CLI)

Here are the suggested improvements to the code:

1. In the `OpenSearchAdapter.java` file, the code can be improved by using a switch statement instead of multiple if-else statements. This will make the code more readable and maintainable. Here's an example:

```java
switch (entry.getValue().fieldType) {
  case TEXT:
    tryRegisterField(mapperService, entry.getValue().name, b -> b.field("type", "text"));
    break;
  case STRING:
    tryRegisterField(mapperService, entry.getValue().name, b -> b.field("type", "keyword"));
    break;
  // Add more cases for other field types as needed
  default:
    // Handle unknown field types
    break;
}
```

2. In the `KaldbDistributedQueryService.java` file, the code can be improved by removing the unused constant `LIMIT_SCHEMA_NODES_TO_QUERY`. Since it is not being used, it can be safely deleted.

3. In the same file, the code can be improved by removing the unnecessary logging statements. The log level can be changed from `debug` to `info` for the log statements in the `distributedSearch` and `distributedSchema` methods. Here's an example:

```java
LOG.info("Starting distributed search for request: {}", distribSearchReq);
```

4. In the `OpenSearchQueryParserTest.java` file, the code can be improved by using `Instant.parse` instead of `DateTimeFormatter.parse` to parse the bucket keys as `Instant` objects. Here's an example:

```java
long bucket1AsLong = Instant.parse(histogram.getBuckets().get(0).getKeyAsString()).toEpochMilli();
long bucket2AsLong = Instant.parse(histogram.getBuckets().get(1).getKeyAsString()).toEpochMilli();
```

5. In the `LogIndexSearcherImplTest.java` file, the code can be improved by using `DateTimeFormatter.ISO_INSTANT` to parse the bucket keys as `Instant` objects. This will make the code more concise and readable. Here's an example:

```java
DateTimeFormatter formatter = DateTimeFormatter.ISO_INSTANT;
long bucket1AsLong = Instant.from(formatter.parse(histogram.getBuckets().get(0).getKeyAsString())).toEpochMilli();
long bucket2AsLong = Instant.from(formatter.parse(histogram.getBuckets().get(1).getKeyAsString())).toEpochMilli();
```

These are some of the improvements that can be made to the code. Let me know if you have any questions!

PR #604, opened 2023-07-06
Add a new bulk ingest API for Kaldb

vthacker:
###  Summary

1. A bulk API which registers around the preprocessor.
2. The preprocessor will either register the pull API we have currently or the bulk API POST endpoint
3. The bulk ingest endpoint will write out to the preprocessor topic and will use regular Kafka producer APIs
4. We will only support docs against 1 index in a single request
5. We will apply rate limiting and either reject the whole request or allow it

### Requirements

* [ ] I've read and understood the [Contributing Guidelines](CONTRIBUTING.md) and have done my best effort to follow them.


kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -363,10 +365,12 @@ private static Set<Service> getServices(
           new ArmeriaService.Builder(serverPort, "kalDbPreprocessor", meterRegistry)
               .withRequestTimeout(requestTimeout)
               .withTracing(kaldbConfig.getTracingConfig())
+              .withAnnotatedService(
+                  new ElasticsearchBulkIngestAPI(
```
vthacker, line 369: todo: OpenSearchBulkIngestAPI

kaldb/src/main/java/com/slack/kaldb/server/ElasticsearchBulkIngestAPI.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * 1. Kaldb does not support the concept of "updates". It's always an add 2. The "index" is used
+   * as the span name
+   */
+  @Blocking
+  @Post("/_bulk")
+  public HttpResponse addDocument(String bulkRequest) {
+    try {
+      Map<String, List<Trace.Span>> docs = OpenSearchRequest.parseBulkHttpRequest(bulkRequest);
```
vthacker, line 109: allow only 1 index

kaldb/src/main/java/com/slack/kaldb/server/ElasticsearchBulkIngestAPI.java
```diff
@@ -0,0 +1,212 @@
...
+      Map<String, List<Trace.Span>> rateLimitedDocs = new HashMap<>();
+      for (Map.Entry<String, List<Trace.Span>> indexDoc : docs.entrySet()) {
+        final String index = indexDoc.getKey();
+        rateLimitedDocs.put(
+            index,
+            indexDoc.getValue().stream()
+                .filter(doc -> rateLimiterPredicate.test(index, doc))
```
vthacker, line 116: 1. Say after 10 docs we hit rate limit , we will drop the rest - make it all or nothing
2. function still reads `service_name` - We pass the key as index. make use of it

kaldb/src/main/java/com/slack/kaldb/server/ElasticsearchBulkIngestAPI.java
```diff
@@ -0,0 +1,212 @@
...
+        continue;
+      }
+      for (Trace.Span doc : indexDoc.getValue()) {
+        KafkaBlockingCallback producerReturnInfo =
+            new KafkaBlockingCallback(completeSignal, errorCount);
+
+        ProducerRecord<String, byte[]> producerRecord =
+            new ProducerRecord<>(
+                preprocessorConfig.getDownstreamTopic(), partition, index, doc.toByteArray());
+        kafkaProducer.send(producerRecord, producerReturnInfo);
```
vthacker, line 150: look into transactions

kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -363,10 +365,12 @@ private static Set<Service> getServices(
           new ArmeriaService.Builder(serverPort, "kalDbPreprocessor", meterRegistry)
               .withRequestTimeout(requestTimeout)
               .withTracing(kaldbConfig.getTracingConfig())
+              .withAnnotatedService(
```
vthacker, line 368: register either bulk ingest or preprocessor service

config/config.yaml
```diff
@@ -132,3 +132,5 @@ preprocessorConfig:
   dataTransformer: ${PREPROCESSOR_TRANSFORMER:-json}
   rateLimiterMaxBurstSeconds: ${PREPROCESSOR_RATE_LIMITER_MAX_BURST_SECONDS:-1}
   kafkaPartitionStickyTimeoutMs: ${KAFKA_PARTITION_STICKY_TIMEOUT_MS:-0}
+  bootstrapServers: ${KAFKA_BOOTSTRAP_SERVERS:-localhost:9092}
+  useBulkApi: ${KALDB_PREPROCESSOR_USE_BULK_API:-false}
```
bryanlb, line 136: Instead of a separate config here, can we just infer this based on the presence/absence of the stream configs / upstream topic?

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
```diff
@@ -37,9 +49,62 @@
...
+      } else {
+        LOG.warn(
+            "request=" + request + " of type " + request.opType().toString() + "not supported");
+      }
+    }
+    return indexRequests;
+  }
+
+  protected static IngestDocument convertRequestToDocument(IndexRequest indexRequest) {
+    String index = indexRequest.index();
```
bryanlb, line 95: Is there a reason we create all of these as separate variables?  Did we consider a static method on IngestDocument that has a `fromIndexRequest()`?

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.errors.AuthorizationException;
+import org.apache.kafka.common.errors.OutOfOrderSequenceException;
+import org.apache.kafka.common.errors.ProducerFencedException;
+import org.apache.kafka.streams.kstream.Predicate;
+import org.opensearch.action.index.IndexRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OpenSearchBulkIngestAPI extends AbstractService {
```
bryanlb, line 42: nit: Should be cased `Api` to match (most) the other uses

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
```diff
@@ -0,0 +1,237 @@
...
+      (datasetMetadata) -> load();
+
+  private final PreprocessorRateLimiter rateLimiter;
+  private Predicate<String, List<Trace.Span>> rateLimiterPredicate;
+  private List<DatasetMetadata> throughputSortedDatasets;
+
+  private final Timer configReloadTimer;
+
+  @Override
+  protected void doStart() {
```
bryanlb, line 60: Does this need to be a guava service? It appears all we're doing is adding a listener to the metadata store?

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
```diff
@@ -0,0 +1,237 @@
...
+      LOG.error("Request failed ", e);
+      BulkIngestResponse response = new BulkIngestResponse(0, 0, e.getMessage());
+      return HttpResponse.ofJson(INTERNAL_SERVER_ERROR, response);
+    }
+  }
+
+  public BulkIngestResponse produceDocuments(Map<String, List<Trace.Span>> indexDocs) {
+    int totalDocs = indexDocs.values().stream().mapToInt(List::size).sum();
+
+    String transactionId = String.join("_", indexDocs.keySet()) + "_" + System.currentTimeMillis();
```
bryanlb, line 167: Does this ever get excessively long? Is there a reason we can't use something like a GUID here?

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
```diff
@@ -0,0 +1,237 @@
...
+          kafkaProducer.initTransactions();
+          kafkaProducer.beginTransaction();
+          for (Trace.Span doc : indexDoc.getValue()) {
+
+            ProducerRecord<String, byte[]> producerRecord =
+                new ProducerRecord<>(
+                    preprocessorConfig.getDownstreamTopic(), partition, index, doc.toByteArray());
+            kafkaProducer.send(producerRecord);
+          }
+          kafkaProducer.commitTransaction();
```
bryanlb, line 193: Should this be part of the finally? What happens if an exception occurs and the transaction is not committed?

kaldb/src/main/java/com/slack/kaldb/writer/MurronLogFormatter.java
```diff
@@ -44,6 +50,41 @@ public static Trace.Span fromApiLog(Murron.MurronMessage murronMsg)
...
   }
 
+  public static Trace.Span fromIngestDocument(IngestDocument ingestDocument) {
+    ZonedDateTime timestamp =
+        (ZonedDateTime)
+            ingestDocument
+                .getIngestMetadata()
+                .getOrDefault("timestamp", ZonedDateTime.now(ZoneOffset.UTC));
+    // Trace.Span expects duration in microseconds today
+    long epochMicro = timestamp.toInstant().toEpochMilli() / 1000;
```
bryanlb, line 60: Microseconds should be milliseconds multiplied by 1000, not divided. We probably should add a test assert against our date handling too, as that'd be a bit of a pain to run down in the future if it regressed.

bryanlb (changes requested):

config/config.yaml
vthacker: it just seems safer to keep an explicit key. We anyways want to move to this and eventually remove the pull based option.

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
bryanlb: Disregard - I missed the fact that `IngestDocument` wasn't our class.

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
vthacker: Currently the service on init does call `load` and also adds a listener. The listener is also removed on shutdown through the guava service API

This is very similar to what we do in the PreprocessorService. I thought guava service was helpful because it cleans up resources on shutdown etc

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
vthacker: The hope is there is always one key in `indexDocs`. We actually today even error out otherwise.

I can definitely change it to GUID, but my initial reason for doing this was this _might_ have enough info if we were debugging Kafka server performance issues. So knowing timestamp and the index that is being used could be valuable. But looking at transaction id might not be even possible.

Happy to change it if you think this is too long

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestAPI.java
vthacker: good question. I forgot to add the close on finally . Updating PR now.

The commitTransaction can throw ProducerFencedException which we catch and cannot retry. However we don't retry anyways on the `KafkaException` catch block today

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
```diff
@@ -107,25 +107,100 @@ protected static RateLimiter smoothBurstyRateLimiter(
     }
   }
 
+  public static int getSpanBytes(List<Trace.Span> spans) {
+    return spans.stream().mapToInt(Trace.Span::getSerializedSize).sum();
+  }
+
+  public Predicate<String, List<Trace.Span>> createBulkIngestRateLimiter(
```
bryanlb, line 114: This is likely going to need some considerable refactoring once we switch to http-only. There's a few issues I can see with this I'm not a huge fan of:

* This still uses a Predicate, which is a Kafka Streams abstraction
* This current rate limiter doesn't have a transactional abstraction. This means that when attempting to reserve rate limits for a set of services, if any fail we don't return the unused rate limits when we fail the request. 

I don't think the second issue needs to be addressed in this PR, but we might want to see if the first is easy to resolve.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestApi.java
```diff
@@ -0,0 +1,230 @@
...
+          }
+          return new BulkIngestResponse(0, totalDocs, e.getMessage());
+        }
+      }
+    }
+
+    return new BulkIngestResponse(totalDocs, 0, "");
+  }
+
+  private int getPartition(String index) {
```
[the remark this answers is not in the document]
vthacker: I think we do this because it's throughput based? I actually don't understand the full value on how logs can belong to multiple service datasets ( in a practical sense ) which is why we use this technique.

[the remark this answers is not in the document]
bryanlb: This is fine per discussion - this is because a given service name can match multiple, depending on glob/regex matching.

kaldb/src/test/java/com/slack/kaldb/server/KaldbConfigTest.java
```diff
@@ -462,6 +464,9 @@ public void testParseKaldbYamlConfigFile() throws IOException {
     assertThat(preprocessorConfig.getDataTransformer()).isEqualTo("api_log");
     assertThat(preprocessorConfig.getRateLimiterMaxBurstSeconds()).isEqualTo(2);
 
+    assertThat(preprocessorConfig.getUseBulkApi()).isEqualTo(true);
```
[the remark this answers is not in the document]
vthacker: done!

kaldb/src/main/java/com/slack/kaldb/preprocessor/PreprocessorRateLimiter.java
vthacker: addressed 1 with BiPredicate

kaldb/src/main/java/com/slack/kaldb/preprocessor/ingest/OpenSearchBulkApiRequestParser.java
```diff
@@ -0,0 +1,125 @@
...
+    List<DocWriteRequest<?>> requests = bulkRequest.requests();
+    for (DocWriteRequest<?> request : requests) {
+      if (request.opType() == DocWriteRequest.OpType.INDEX) {
+
+        // The client makes a DocWriteRequest and sends it to the server
+        // IngestService#innerExecute is where the server eventually reads when request is an
+        // IndexRequest. It then creates an IngestDocument
+        indexRequests.add((IndexRequest) request);
+      } else {
+        LOG.warn(
```
bryanlb, line 119: These should use the token version (ie `{}`) instead of string concat

bryanlb (dismissed):

bryanlb (dismissed):

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestApi.java
```diff
@@ -32,11 +33,17 @@
...
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.errors.AuthorizationException;
+import org.apache.kafka.common.errors.OutOfOrderSequenceException;
 import org.apache.kafka.common.errors.ProducerFencedException;
 import org.opensearch.action.index.IndexRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * batching is important - if we send one doc a time we will create a transaction per request which
```
bryanlb, line 44: This comment could be expanded, describing what the entire class is for.

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestApi.java
```diff
@@ -163,41 +180,40 @@ public HttpResponse addDocument(String bulkRequest) {
...
-          continue;
+    for (Map.Entry<String, List<Trace.Span>> indexDoc : indexDocs.entrySet()) {
+      String index = indexDoc.getKey();
+      // call once per batch and use the same partition for better batching
+      int partition = getPartition(index);
+
+      // since there isn't a dataset provisioned for this service/index we will not index this set
+      // of docs
+      if (partition < 0) {
+        LOG.warn("index=" + index + " does not have a provisioned dataset associated with it");
```
bryanlb, line 193: This should be using the tokenized string, not concat

bryanlb (approved):

PR #607, opened 2023-07-12
Cap the amount of concurrent replica assignments per host

bryanlb:
###  Summary

Caps the number of concurrent downloads per cache node. This will prevent queuing a very large number of slots that would otherwise take precedence over the newer replicas.


vthacker (approved):
Nice!

PR #609, opened 2023-07-14
Revert "S3 CRT Client"

bryanlb:
Temporarily reverting, as this client appears to not respect the no_proxy env variable.

PR #610, opened 2023-07-14
S3 CRT Client

bryanlb:
Reverts slackhq/kaldb#609.

⚠️ Depends on `NO_PROXY` support before this can be merged - https://github.com/awslabs/aws-c-http/issues/413.

Today there exists no way to initialize the S3 CRT client from Java using no proxy, if the HTTP_PROXY/HTTPS_PROXY environment values are set. These are directly read from native C code when an empty proxy config is detected.

Attempting to initialize the proxy with null values does not bypass this behavior, nor does attempting to initialize the proxy with an empty string (throws errors in native code). Attempting to set the http.proxy/https.proxy Java values to empty also does not stop reading the ENV values, as this is also detected as an empty proxy config. The only way to avoid this behavior is to completely unset the proxy values from the system environment values, as mentioned in the issue above.

Initial benchmarks with this code (even running through the proxy) showed download speeds took less than 50% of the original time. As soon as some form of `no_proxy` support is added, or there exists a way to force disable reading of the env variables this should be revisited.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -34,6 +34,7 @@ s3Config:
   s3Region: ${S3_REGION:-us-east-1}
   s3EndPoint: ${S3_ENDPOINT:-http://localhost:9090}
   s3Bucket: ${S3_BUCKET:-test-s3-bucket}
+  s3TargetThroughputGbps: ${S3_TARGET_THROUGHPUT_GBPS:-25}
 
 tracingConfig:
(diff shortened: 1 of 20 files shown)


bryanlb:
We have a workaround in place for the time being, using a squid sidecar and then providing the env variables `aws.s3.crt.proxy.scheme` / `aws.s3.crt.proxy.host` / `aws.s3.crt.proxy.port`. This seems to be working as expected, and we can revert this portion once the fix is rolled into the upstream.

vthacker (approved):

PR #644, opened 2023-08-03
Add Kubernetes HPA metric support for Cache nodes

bryanlb:
###  Summary

Adds support for reporting HPA (horizontal pod autoscaling) Kubernetes metrics for the cache nodes. As we need to provide an application-aware implementation, this is calculated by the manager and stored in ZK as ephemeral nodes with the scaling decision. Cache nodes then read this value and report it as a pod-level metric. This allows the end user to deploy an HPA using the below configuration to enable auto-scaling on their cluster.

```yaml
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: ${APP}
  minReplicas: 1
  maxReplicas: ${MAX_REPLICAS}
  metrics:
    - type: Pods
      pods:
        metric:
          name: hpa_cache_demand_factor_${REPLICA_SET}
        target:
          type: AverageValue
          averageValue: 1.0
```

Manager node is responsible for doing calculation of target nodes using replica demand and cache slot availability, and ensures that scale-down operations are time-locked to a replica set using the Kubernetes implicit maintenance mode functionality.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
@@ -1,0 +1,173 @@
+package com.slack.kaldb.clusterManager;
+
+import com.google.common.util.concurrent.AbstractScheduledService;
+import com.slack.kaldb.metadata.cache.CacheSlotMetadataStore;
+import com.slack.kaldb.metadata.hpa.HpaMetricMetadata;
+import com.slack.kaldb.metadata.hpa.HpaMetricMetadataStore;
+import com.slack.kaldb.metadata.replica.ReplicaMetadata;
+import com.slack.kaldb.metadata.replica.ReplicaMetadataStore;
+import com.slack.kaldb.proto.metadata.Metadata;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The Cluster HPA metrics service is intended to be run on the manager node, and is used for making
+ * centralized, application-aware decisions that can be used to inform a Kubernetes horizontal pod
+ * autoscaler (HPA).
+ *
+ * @see <a
+ *     href="https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/">Kubernetes
+ *     HPA</a>
+ */
+public class ClusterHpaMetricService extends AbstractScheduledService {
+  private static final Logger LOG = LoggerFactory.getLogger(ClusterHpaMetricService.class);
+
+  // todo - consider making over-provision and lock duration configurable values
+  private static final Double CACHE_OVER_PROVISION = 1.1;
+
+  protected Duration CACHE_SCALEDOWN_LOCK = Duration.of(15, ChronoUnit.MINUTES);
+  protected static final String CACHE_HPA_METRIC_NAME = "hpa_cache_demand_factor_%s";
+
+  private final ReplicaMetadataStore replicaMetadataStore;
+  private final CacheSlotMetadataStore cacheSlotMetadataStore;
+  private final HpaMetricMetadataStore hpaMetricMetadataStore;
+  protected final Map<String, Instant> cacheScalingLock = new ConcurrentHashMap<>();
+
+  public ClusterHpaMetricService(
+      ReplicaMetadataStore replicaMetadataStore,
+      CacheSlotMetadataStore cacheSlotMetadataStore,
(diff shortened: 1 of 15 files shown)


kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
```diff
@@ -0,0 +1,172 @@
...
+   *     pods:
+   *       metric:
+   *         name: hpa_cache_demand_factor_REPLICASET
+   *       target:
+   *         type: AverageValue
+   *         averageValue: 1.0
+   * </pre>
+   */
+  private void publishCacheHpaMetrics() {
+    Set<String> replicaSets =
```
vthacker, line 92: qq: do we need `listSync` in all these places? I presume yes, but maybe we can add a small statement as to why?

kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
```diff
@@ -0,0 +1,172 @@
...
+              .count();
+      long totalReplicaDemand =
+          replicaMetadataStore.listSync().stream()
+              .filter(replicaMetadata -> replicaMetadata.getReplicaSet().equals(replicaSet))
+              .count();
+      double rawDemandFactor =
+          (double) (totalReplicaDemand + CACHE_SLOT_OVER_PROVISION) / (totalCacheSlotCapacity + 1);
+      double demandFactor = (double) Math.round(rawDemandFactor * 100) / 100;
+      LOG.info(
+          "Cache autoscaler for replicaSet '{}' calculated a demandFactor of '{}' - totalReplicaDemand: '{}', cacheSlotOverProvision: '{}', totalCacheSlotCapacity: '{}'",
```
vthacker, line 110: will this log line be too noisy? The scheduler runs every 30s and two replcasets.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @see <a
+ *     href="https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/">Kubernetes
+ *     HPA</a>
+ */
+public class ClusterHpaMetricService extends AbstractScheduledService {
+  private static final Logger LOG = LoggerFactory.getLogger(ClusterHpaMetricService.class);
+
+  // todo - consider making over-provision and lock duration configurable values
+  protected int CACHE_SLOT_OVER_PROVISION = 1000;
```
vthacker, line 35: so for each replica set we add 1000 to the totalReplicas to calculate demand. Is this value too high?

kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
bryanlb: After the ZK refactor, `listSync` uses the cache now. The only other option here is the listAsync but that's more of a syntax preference if you want to use the completable future or not.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
bryanlb: This adds 1000 slots, which would be only a handful of instances depending on the slot density. We typically run between 200 - 500 slots, so that's 2-5 pods extra.

This will eventually probably be a config, but we would ideally target a number slightly above x2 the slot capacity so that any one instance can go down and we'd still have capacity for rollovers.

kaldb/src/main/java/com/slack/kaldb/clusterManager/ClusterHpaMetricService.java
bryanlb: This is the easiest way to troubleshoot the behavior and determine what's going on with the logic - we may be able to reduce it in the future, but this would only be four lines per min which is negligible for our overall cluster.

kaldb/src/main/java/com/slack/kaldb/metadata/core/CuratorBuilder.java
```diff
@@ -70,6 +71,7 @@ public static AsyncCuratorFramework build(
                   && curatorEvent.getWatchedEvent().getState()
                       == Watcher.Event.KeeperState.Expired) {
                 LOG.warn("The ZK session has expired {}.", curatorEvent);
+                new RuntimeHalterImpl().handleFatal(new Throwable("ZK session expired."));
```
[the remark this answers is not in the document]
bryanlb: Correct, this is a fix that is not directly related to this PR but was exposed after deploying this PR to dev. The autoscaling kept growing when it wasn't expected to, and troubleshooting that issue revealed that cache slots were disappearing once ZK reconnected. This appears to have settled down the autoscaling to be more in line with what we were expecting, but we'll need some longer term data to ensure there's not additional issues causing the slots to disappear.

This would be an important fix to keep (and corresponding tests updates), regardless of the rest of this PR.

vthacker (approved):

PR #655, opened 2023-08-30
Add latch for ZK hydration initialization

### LINKED ISSUE #651: Indexer couldn't find all previous metadata snapshots
### Describe the bug
During indexer starts up, it was supposed to find all the existing metadata snapshots from ZK and only pull the newer data after the last snapshot.

But the listAsync() call from KaldbPartitionMetadataStore couldn't retrieve all existing snapshots from ZK (in fact most of the time it retrieved 0 snapshots), changing the listAsync() to listSync() seems fixing the problem.

See discussion thread here: https://slack-pde.slack.com/archives/C02028H9LHJ/p1692300820984549?thread_ts=1691511612.465979&cid=C02028H9LHJ

### Requirements (place an `x` in each of the `[ ]`)**
* [ ] I've read and understood the [Contributing guidelines](../CONTRIBUTING.md) and have done my best effort to follow them.
* [ ] I've searched for any related issues and avoided creating a duplicate issue.

#### Reproducible in:

KalDB version:

JVM version:

OS version(s):

bryanlb:
###  Summary

We currently do not block on cache initialization, which can cause downstream consumers to read incomplete/empty cache results. This now enforces the cache is completely hydrated before we allow any calls that rely on the cache to complete.

Resolves #651, closes #652

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
@@ -2,12 +2,14 @@ package com.slack.kaldb.metadata.core;
 
 import static com.slack.kaldb.server.KaldbConfig.DEFAULT_ZK_TIMEOUT_SECS;
 
+import com.slack.kaldb.util.RuntimeHalterImpl;
 import java.io.Closeable;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -32,6 +34,8 @@ public class KaldbMetadataStore<T extends KaldbMetadata> implements Closeable {
   protected final String storeFolder;
 
   private final ZPath zPath;
+
+  private final CountDownLatch cacheInitialized = new CountDownLatch(1);
 
   protected final ModeledFramework<T> modeledClient;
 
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -61,6 +65,7 @@ public KaldbMetadataStore(
 
     if (shouldCache) {
       cachedModeledFramework = modeledClient.cached();
+      cachedModeledFramework.listenable().addListener(getCacheInitializedListener());
       cachedModeledFramework.start();
```
vthacker, line 69: why don't we add `awaitCacheInitialized();` here and wait till the cache has been initialized? vs awaiting before the async ZK operations?

bryanlb: Two reasons; first, there exists a race condition where this can result in a deadlock when attempting to initialize multiple copies of a cached store; second, we only need to guarantee a hydrated cache for _some_ of these methods, which results in unnecessary blocking on initialization if this is in the constructor.

vthacker: I see the point in 2) i.e only the get operations need to wait for cache to be initialized. but it shouldn't take very long? Seems like we can miss calls to await if we add new read APIs

But I don't understand the deadlock aspect. Either way it would be good to document it in code?

kaldb/src/test/java/com/slack/kaldb/metadata/core/KaldbMetadataStoreTest.java
```diff
@@ -394,4 +394,56 @@ public TestMetadataStore() {
...
+    int testMetadataInitCount = 10;
+    try (KaldbMetadataStore<TestMetadata> init = new FastMetadataStore()) {
+      for (int i = 0; i < testMetadataInitCount; i++) {
+        init.createSync(new TestMetadata("name" + i, "value" + i));
+      }
+    }
+
+    try (KaldbMetadataStore<TestMetadata> init = new SlowMetadataStore()) {
+      List<TestMetadata> metadata = init.listSync();
+      assertThat(metadata.size()).isEqualTo(testMetadataInitCount);
```
vthacker, line 446: cool! this assert fails when I commented out `awaitCacheInitialized` from the listAsync method :)

PR #656, opened 2023-08-31
Add the ability to have the indexer start from head

### LINKED ISSUE #620: Create recovery task when no existing offset is persisted
### Description

Today we start from the beginning of the topic retention, which can cause significant amounts of lag / delay before we are ingesting current data. We should always start with the current in this case, and create one or more recovery tasks to catch up the missing data.

kyle-sammons:
###  Summary
There are certain situations when you have data in Kafka, but no metadata information, and the Indexer is coming up clean (think creating a new cluster from a Kafka cluster with data in it, or various disaster recovery scenarios). In these situations you may want to have the Indexer not start from as far back as Kafka possibly can go, but instead just start reading from head. There are yet other scenarios where you want the Indexer to start reading from head and to have it create recovery tasks for all of the data between the tail and head. This PR attempts to allow both of those use cases.

Resolves #620.

### Requirements

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -19,18 +19,19 @@
         <http.client.version>4.5.14</http.client.version>
         <http.core.version>4.4.16</http.core.version>
         <protobuf.version>3.23.2</protobuf.version>
-        <grpc.version>1.55.1</grpc.version>
+        <grpc.version>1.57.2</grpc.version>
         <micrometer.version>1.11.0</micrometer.version>
-        <armeria.version>1.24.3</armeria.version>
+        <armeria.version>1.25.2</armeria.version>
         <kafka.version>3.4.0</kafka.version>
         <jackson.version>2.15.2</jackson.version>
         <jackson.databind.version>2.15.1</jackson.databind.version>
         <lucene.version>9.5.0</lucene.version>
         <opensearch.version>2.7.0</opensearch.version>
-        <curator.version>5.4.0</curator.version>
+        <curator.version>5.5.0</curator.version>
         <log4j.version>2.20.0</log4j.version>
         <aws.sdk.version>2.20.116</aws.sdk.version>
         <error.prone.version>2.16</error.prone.version>
+        <junit.jupiter.version>5.10.0</junit.jupiter.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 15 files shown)


bryanlb:
Closes #520

kaldb/src/main/java/com/slack/kaldb/server/KaldbIndexer.java
```diff
@@ -110,7 +112,10 @@ private long indexerPreStart() throws Exception {
             meterRegistry);
 
     long currentHeadOffsetForPartition = kafkaConsumer.getEndOffSetForPartition();
```
bryanlb, line 114: I'm not a huge fan of calling these head/tail, especially since we seem to have inverted the common definitions. Tail is generally the most recent messages, head the oldest. I realize you haven't started this pattern, but we are extending this further now.

Maybe using similar terms as Kafka is a good plan here - end / beginning.

kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
```diff
@@ -184,7 +188,40 @@ public long determineStartingOffset(long currentHeadOffsetForPartition) {
 
     if (highestDurableOffsetForPartition <= 0) {
       LOG.info("There is no prior offset for this partition {}.", partitionId);
-      return highestDurableOffsetForPartition;
+
+      // If the user wants to start at the current offset in Kafka and _does not_ want to create
```
bryanlb, line 192: You might wanna reformat this comment block, as the auto-formatter seems to have wrecked it a bit.

kaldb/src/main/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumer.java
```diff
@@ -42,6 +42,7 @@ public class KaldbKafkaConsumer {
   private static final Logger LOG = LoggerFactory.getLogger(KaldbKafkaConsumer.class);
   public static final int KAFKA_POLL_TIMEOUT_MS = 250;
   private final LogMessageWriterImpl logMessageWriterImpl;
+  private KaldbConfigs.KafkaConfig kafkaConfig;
```
bryanlb, line 45: This should be able to be final

kaldb/src/main/proto/kaldb_configs.proto
```diff
@@ -85,6 +85,11 @@ message QueryServiceConfig {
   string managerConnectString = 3;
 }
 
+enum KafkaOffsetLocation {
+  BEGINNING = 0;
```
bryanlb, line 89: Here's another place where we introduce more terms describing where the offset is at. To recap:

Kafka configs use `earliest` and `latest`, which is read into `beginningOffset` and `endOffset` respectively.
Kaldb configs use `beginning` and `current`, which is read into `tailOffset` and `headOffset` respectively. 

We should consolidate these to use the Kafka pattern for describing these.

bryanlb (changes requested):
Just a few small things, otherwise this looks good.

bryanlb (approved):

PR #649, opened 2023-08-21
Use lucene directory abstraction to compute file size

vthacker:
###  Summary

When we use the regular file abstractions to calculate the size of the lucene index, lucene might be merging segments. This means the function has a race condition between the time it fetches the list and the time the calculation is done. 

This is harmless in practice as we just try again and keep the old value. But it leads to noisy logs. When we use Lucene's FS abstraction it has a list of pendingDeletes which helps avoid this race condition.

Another advantage of this PR is during the happy case i.e when the java file abstraction computes the size it can take the pending deletes into account. So the disk size is smaller than what we want to. This will address that problem as well as noted by Bryan

PR #676, opened 2023-09-15
Add error handling when reaching max keys listing S3 objects

bryanlb:
###  Summary

There exists an issue today where when attempting to list files from S3, it can fail to exit the loop listing files. This is because there are no checks against reaching the default max keys of 1000. The new AWS CRT client has different behavior when attempting to read the continuation token of a listing that has reached max, which causes this to infinitely block downloading the chunk.

This increases the default of 1000 to 2500, and adds error checking for this edge case. The current chunk sizes of 15GB can occasionally just exceed 1k files which eventually causes the cluster to become unhealthy. Improvements such as #649 are resulting in better utilizing the disk space which can increase the likelihood of this occurring.


vthacker (approved):
nice find!

PR #678, opened 2023-09-19
Upgrade to java 21

vthacker:
###  Summary

Upgrade to Java 21. This change means code will need to be built with Java 21 so your local dev environment needs to have Java 21 as JAVA_HOME when compiling kaldb locally

This PR also explicitly defines the version of bytebuddy being used. The ones that get packaged as a transitive dependency don't have the latest update that works with java 21

From [this](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.htm)
> . You can always guarantee a version by declaring it explicitly in your project's POM


.github/workflows/maven.yml
```diff
@@ -19,15 +19,15 @@ jobs:
...
       uses: actions/setup-java@v3
       with:
-        java-version: '17'
-        distribution: 'temurin'
+        java-version: '21'
+        distribution: 'corretto'
         cache: 'maven'
     - name: Ensure code is formatted
-      run: mvn -B -Dstyle.color=always com.spotify.fmt:fmt-maven-plugin:check --file kaldb/pom.xml
+      run: mvn -B -Dstyle.color=always spotless:check --file kaldb/pom.xml
```
bryanlb, line 29: Looks like `fmt-mvn-plugin` just got an update this morning. What do you think about giving this a shot? https://github.com/spotify/fmt-maven-plugin/pull/179

.github/workflows/benchmark.yml
```diff
@@ -23,11 +23,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@v3
-      - name: Set up JDK 17
+      - name: Set up JDK 21
         uses: actions/setup-java@v3
         with:
-          java-version: '17'
-          distribution: 'temurin'
+          java-version: '21'
+          distribution: 'corretto'
```
bryanlb, line 30: We initially used `temurin` due to build caches, but since we use `corretto` in production it's not awful to build & test against that.

kaldb/pom.xml
```diff
@@ -438,6 +439,12 @@
             <version>5.2.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
bryanlb, line 442: Maybe add a comment here why we have this specific version pinned, and then add a comment to the Mockito version that we can remove Bytebuddy probably after the next Mockito version bump.

kaldb/src/test/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumerTest.java
```diff
@@ -223,6 +223,7 @@ public void testConsumeMessagesBetweenOffsets() throws Exception {
       // Assign doesn't create a consumer group.
       assertThat(kafkaServer.getConnectedConsumerGroups()).isEqualTo(0);
     }
+
```
bryanlb, line 226: Looks like an errant return here

bryanlb (changes requested):

kaldb/src/test/java/com/slack/kaldb/writer/kafka/KaldbKafkaConsumerTest.java
vthacker: this was the only file that changes because of the new google format version bump

.github/workflows/maven.yml
vthacker: that seems to do the job!

bryanlb (approved):

PR #679, opened 2023-09-20
ZSTD codec

vthacker:
###  Summary

Describe the goal of this PR. Mention any related Issue numbers.

### Requirements

* [ ] I've read and understood the [Contributing Guidelines](CONTRIBUTING.md) and have done my best effort to follow them.

### DIFF
--- a/kaldb/pom.xml
+++ b/kaldb/pom.xml
@@ -27,6 +27,7 @@
         <jackson.databind.version>2.15.1</jackson.databind.version>
         <lucene.version>9.5.0</lucene.version>
         <opensearch.version>2.7.0</opensearch.version>
+        <zstd.jni.version>1.5.5-5</zstd.jni.version>
         <curator.version>5.5.0</curator.version>
         <log4j.version>2.20.0</log4j.version>
         <aws.sdk.version>2.20.116</aws.sdk.version>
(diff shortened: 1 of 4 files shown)


bryanlb:
I looked a bit more into this, and I'm not convinced that ZSTD is the direction we want to go. The discussion from OpenSearch [here](https://github.com/opensearch-project/OpenSearch/issues/9422) specifically raises some concerns:

> One of those reasons is the hard dependency on native compiled code, which often leads to portability issues due to glibc differences.

There's also currently issues with memory leaks, and aside from that the performance (especially at query time) is a mixed bag. If we were already running zlib this may be more interesting, but we currently use the default of BEST_SPEED.

https://github.com/apache/lucene/issues/9784#issuecomment-1223937672

> |Codec|Indexing time (ms)|Disk usage (MB)|Retrieval time per 10k docs (ms)|
> |-|-|-|-|
> |BEST_SPEED (LZ4 with small blocks)|35383|90.175|190.17524|
> |BEST_COMPRESSION (vanilla zlib, DEFLATE level=6)|76671|58.682|1910.42106|
> |BEST_COMPRESSION (Cloudflare zlib, DEFLATE level=6)|54791|58.601|1395.53593|
> |ZSTD dict (level=1)|24687|63.324|928.73997|
> |ZSTD dict (level=2)|24934|63.722|977.29911|
> |ZSTD dict (level=3)|28285|62.072|938.10886|
> |ZSTD dict (level=4)|37863|60.427|969.18655|
> |ZSTD dict (level=5)|45479|59.317|941.20922|
> |ZSTD dict (level=6)|57842|58.481|881.69049|
> |ZSTD dict (level=7)|65796|58.107|886.42249|

I think this is something we can table for now, and potentially revisit in the future.

PR #680, opened 2023-09-20
Add AWS S3 Transfer Manager

bryanlb:
###  Summary

Adds the S3 Transfer manager for uploading and downloading files and folders. If a folder is detected on upload or download this will automatically switch to using the directory APIs which have much better concurrency and throughput.

This also adds a missing test suite from when we split off the CRT client, and fixes a previously unknown exception issue. As mentioned in the new CRT docblock, we need to revisit our entire blobfs abstraction as the current design isn't working well with the direction of the APIs we need to support.


kaldb/src/test/java/com/slack/kaldb/elasticsearchApi/OpenSearchBulkEndpointTest.java
```diff
@@ -268,6 +269,7 @@ record ->
   }
 
   @Test
+  @Disabled("Flaky test")
```
bryanlb, line 272: This is intermittently getting stuck in execution, even with a higher timeout. Disabling for now as it's also failing master - https://github.com/slackhq/kaldb/actions/runs/6242013937/job/16945192608

kaldb/src/main/java/com/slack/kaldb/blobfs/s3/S3CrtBlobFs.java
```diff
@@ -491,21 +513,84 @@ public void copyToLocalFile(URI srcUri, File dstFile) throws Exception {
...
+    if (isDirectory(srcUri)) {
+      CompletedDirectoryDownload completedDirectoryDownload =
+          transferManager
+              .downloadDirectory(
+                  DownloadDirectoryRequest.builder()
+                      .destination(dstFile.toPath())
+                      .bucket(srcUri.getHost())
+                      .listObjectsV2RequestTransformer(
+                          builder -> {
+                            builder.maxKeys(2500);
```
vthacker, line 526: Can we use LIST_MAX_KEYS ?

bryanlb: Ah yes, good call. Fixed.

vthacker (approved):

PR #681, opened 2023-09-20
Change HPA to account for Kubernetes tolerances

bryanlb:
###  Summary

This changes the HPA logic to account for the Kubernetes tolerances (see https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details).

This also changes the implicit maintenance window to instead return a nominal value, as the implicit maintenance mode effectively locks out autoscaling until the next deploy for that replicaset.


bryanlb:
Should be merged after #680

vthacker (approved):

PR #682, opened 2023-09-21
Fix flaky test - testDocumentInKafkaTransactionError

vthacker:
###  Summary

I hardened the test to ensure timing issue doesn't mess with asserts with offsets.

Previously the assert for consumer offset depends on when kafka producer flushes documents to the broker. 

This PR also adds linger.ms to the producer to match what we have today


kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestApi.java
```diff
@@ -241,6 +241,7 @@ private KafkaProducer<String, byte[]> createKafkaTransactionProducer(String tran
     props.put("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
     props.put("value.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer");
     props.put("transactional.id", transactionId);
+    props.put("linger.ms", 250);
```
bryanlb, line 244: As discussed, we'll followup with supporting `KAFKA_ADDITIONAL_PROPS` here as well in some format as we probably want to support `snappy`, etc.

bryanlb (approved):

PR #687, opened 2023-09-23
use virtual threads for blocking tasks

vthacker:
###  Summary

Verified via break points, that the virtual thread pool gets uses for the zipkin service and the grpc services


vthacker:
If https://github.com/line/armeria/issues/4911 is merged, we don't need this.

bryanlb:
Closing as discussed - the current plan is to move towards a model where we'll use virtual threads for our code, but won't attempt to change executors on existing code that haven't considered virtual threads.

bryanlb:
Reference #701

Issue #611, opened 2023-07-24
Indexers never expire LIVE snapshots


bryanlb:
### Describe the bug

Indexers currently never expire live snapshots. For long running indexers, this can cause disk saturation issues and also causes failures to delete once these reach the snapshot deletion service time.

```
Starting delete of snapshot SnapshotMetadata{name='LIVE_log_1689833167_dc25ce42-abbf-41ca-955e-e0888a66fe75', snapshotPath='LIVE', snapshotId='LIVE_log_1689833167_dc25ce42-abbf-41ca-955e-e0888a66fe75', startTimeEpochMs=1689228850484, endTimeEpochMs=1689837644041, maxOffset=45729812334, partitionId='83', indexType=LOGS_LUCENE9}
```
```
Exception deleting snapshot
error_root_cause_stack_trace:
    java.lang.IllegalArgumentException: Parameter 'Bucket' must not be null at software.amazon.awssdk.protocols.xml.internal.marshall.SimpleTypePathMarshaller.lambda$static$0(SimpleTypePathMarshaller.java:46) at software.amazon.awssdk.protocols.xml.internal.marshall.XmlProtocolMarshaller.doMarshall(XmlProtocolMarshaller.java:104) at software.amazon.awssdk.protocols.xml.internal.marshall.XmlProtocolMarshaller.marshall(XmlProtocolMarshaller.java:80) at software.amazon.awssdk.protocols.xml.internal.marshall.XmlProtocolMarshaller.marshall(XmlProtocolMarshaller.java:49) at software.amazon.awssdk.services.s3.transform.ListObjectsV2RequestMarshaller.marshall(ListObjectsV2RequestMarshaller.java:51) at software.amazon.awssdk.services.s3.transform.Li
```

### To Reproduce
Long running indexers will show high stale snapshot counts on reboots. Also an increase in exception attempting to delete LIVE snapshots.

```
Deleting 38 stale snapshots: [SnapshotMetadata{name='LIVE_log_1690035728_cf55...
```

### Expected behavior
Indexers should expire LIVE snapshots after a configurable time, or once cache node(s) start serving the content.

vthacker:
this fixes the symptom - https://github.com/slackhq/kaldb/pull/650

Probably not going to go with this

bryanlb:
[...]

PR #690, opened 2023-09-25
Fixed Indexer/Recover nodes' handling of chunks and snapshots

kyle-sammons:
###  Summary
This PR attempts to fix up how Indexer and Recovery nodes handle chunks and snapshots. Specifically, it reworks `ChunkCleanerService` to be event driven (i.e. only triggered on chunk rollover) rather than time based (i.e. running ever 2 hours) and also allows users to set a configurable amount of max chunks they're comfortable with being on disk; this should prevent the edge case where we are ingesting data too quickly and possibly filling up disk before cleaning it up. Further, this PR has Recovery nodes actually use the `ChunkCleanerService`, preventing us from building up stale data on disk. And lastly, it addresses a long standing bug where we'd try to clean up `LIVE` snapshots in the `SnapshotDeletionService`, ultimately failing due to an expectation that the data is in S3 (when `LIVE` snapshots only have their data on disk).

Resolves https://github.com/slackhq/kaldb/issues/611

### Requirements


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkCleanerService.java
```diff
@@ -1,54 +1,89 @@
...
+    final List<Chunk<T>> chunksToDelete = sortedChunks.subList(0, totalChunksToDelete);
+
+    LOG.info("Number of chunks past limit of {} is {}", limit, chunksToDelete.size());
+    chunkManager.removeStaleChunks(chunksToDelete);
+    return chunksToDelete.size();
   }
 
-  @VisibleForTesting
-  public int deleteStaleChunksPastCutOff(long staleDataCutOffMs) {
+  private int deleteStaleChunksPastCutOff(long staleDataCutOffMs) {
```
bryanlb, line 80: In general taking `Instant` instead of a `long` when referring to time is a much safer design - as you don't have to worry about someone in the future passing an incorrect time unit.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkCleanerService.java
```diff
@@ -1,54 +1,89 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(ChunkCleanerService.class);
 
+  private final ChunkManager<T> chunkManager;
+  private final int limit;
+
   private final Duration staleDelayDuration;
-  private final IndexingChunkManager chunkManager;
 
-  public ChunkCleanerService(IndexingChunkManager chunkManager, Duration staleDelayDuration) {
+  public ChunkCleanerService(ChunkManager<T> chunkManager, int limit, Duration staleDelayDuration) {
```
bryanlb, line 26: I'm not sure I like the design of this in light of the new changes we've made. Have we considered getting rid of this `ChunkCleanerService` class entirely, and moving these stale chunks to the indexer implementations specifically?

The idea that we need to keep multiple stale ones only applies to the indexer anyhow - we're just setting it to "0" for the recovery nodes to reuse the code.

It seems like we could move all this complexity internal to the indexing chunk manager, and then the recovery indexer has a much simpler cleanup logic. Then we also don't expose the extra methods to the chunk manager interface.

bryanlb (changes requested):

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkCleanerService.java
bryanlb: I think that removing the abstraction would be a good idea here. The problem is that for one context (indexer) we have the idea of a limit, delay duration - and the only other context we invoke this method (recovery) we just pass empty values (ie `new ChunkCleaner<>(this, 0, Duration.ZERO);`). 

This also helps solves the other issue of needing to expose these to the `ChunkManager` interface and then implementing them with `UnsupportedOperationException` in the cache. You wouldn't need to add any methods to the interface then, and the the behavior is sufficiently different between the recovery and index nodes that it'll make it easier to read if they're separate implementations.

kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManager.java
```diff
@@ -13,4 +15,8 @@ public interface ChunkManager<T> {
   SearchResult<T> query(SearchQuery query, Duration queryTimeout);
 
   Map<String, FieldType> getSchema();
+
+  List<Chunk<T>> getChunkList();
```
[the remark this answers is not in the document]
bryanlb: As discussed in https://github.com/slackhq/kaldb/pull/690#discussion_r1378268291 I think that we can avoid this by removing the abstraction. Resoling this comment for now so we can continue the conversation there.

bryanlb (approved):

PR #691, opened 2023-09-26
Update disk or message rollover to use SegmentInfos

bryanlb:
###  Summary

Updates the disk or message rollover strategy to use the Lucene `SegmentInfos` API. This helps account for temporary files, pending deletes, and pending merges influencing the observed directory size. 

This should allow our chunk size on disk to be much closer to the target value. Note that this does use the last commit for this value, so long commit durations can result in overshooting the target value. Also note that any pending merges may still influence the final size on disk.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
+++ b/kaldb/src/main/java/com/slack/kaldb/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
@@ -7,12 +7,13 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.kaldb.proto.config.KaldbConfigs;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


kaldb/src/main/java/com/slack/kaldb/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
```diff
@@ -106,27 +107,24 @@ public static long calculateDirectorySize(AtomicReference<FSDirectory> activeChu
...
+            .sum();
       }
+    } catch (IndexNotFoundException ignored) {
+      // no committed index found (may be brand new)
     } catch (Exception e) {
       LOG.error("Error calculating the directory size", e);
+      return -1;
     }
-    return -1;
+    return 0;
```
vthacker, line 127: We have this code comment in the class today. Is returning two different values for different exceptions helpful?


```
          // in case the method fails to calculate we return -1 so don't update the old value
          if (dirSize > 0) {
            approximateDirectoryBytes.set(dirSize);
          }
```

vthacker: I guess you need it for the test

vthacker (approved):

PR #693, opened 2023-09-29
Register OpenSearchBulkIngestApi as a service itself

vthacker:
###  Summary

Small fixes
1. Register OpenSearchBulkIngestApi as a service itself so that the `doStart/doStop` methods get invoked
2. some safety checks to ensure we always start with correct configs

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
+++ b/kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
@@ -393,10 +393,11 @@ public class Kaldb {
           new CloseableLifecycleManager(
               KaldbConfigs.NodeRole.PREPROCESSOR, List.of(datasetMetadataStore)));
 
-      LOG.info("TEST: getUseBulkApi=" + preprocessorConfig.getUseBulkApi());
       if (preprocessorConfig.getUseBulkApi()) {
-        armeriaServiceBuilder.withAnnotatedService(
-            new OpenSearchBulkIngestApi(datasetMetadataStore, preprocessorConfig, meterRegistry));
+        OpenSearchBulkIngestApi openSearchBulkApiService =
+            new OpenSearchBulkIngestApi(datasetMetadataStore, preprocessorConfig, meterRegistry);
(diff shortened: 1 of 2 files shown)


kaldb/src/main/java/com/slack/kaldb/server/Kaldb.java
```diff
@@ -394,8 +394,10 @@ private static Set<Service> getServices(
               KaldbConfigs.NodeRole.PREPROCESSOR, List.of(datasetMetadataStore)));
 
       if (preprocessorConfig.getUseBulkApi()) {
-        armeriaServiceBuilder.withAnnotatedService(
-            new OpenSearchBulkIngestApi(datasetMetadataStore, preprocessorConfig, meterRegistry));
+        OpenSearchBulkIngestApi openSearchBulkApiService =
+            new OpenSearchBulkIngestApi(datasetMetadataStore, preprocessorConfig, meterRegistry);
+        armeriaServiceBuilder.withAnnotatedService(openSearchBulkApiService);
+        services.add(openSearchBulkApiService);
```
bryanlb, line 400: Let's add a todo here to revisit this - I'm not sure I'm a huge fan of mixing the Armeria service with a guava service in the same class.

kaldb/src/main/java/com/slack/kaldb/server/OpenSearchBulkIngestApi.java
```diff
@@ -97,6 +99,9 @@ public void load() {
       List<DatasetMetadata> datasetMetadataToProcesses =
           filterValidDatasetMetadata(datasetMetadataList);
 
+      checkArgument(
```
bryanlb, line 102: These should be `checkState` instead of `checkArgument` - the argument version is only intended for checking params to the calling method.

```
  /**
   * Ensures the truth of an expression involving one or more parameters to the calling method.
   *
   * @param expression a boolean expression
   * @throws IllegalArgumentException if {@code expression} is false
   */
  public static void checkArgument(boolean expression) {
```

bryanlb (changes requested):

bryanlb (approved):

PR #703, opened 2023-10-19
Update AWS SDK, remove fixed executor for cache assignments

bryanlb:
###  Summary
Updates the AWS SDK to disable our temp proxy bypass, removed fixed executor for cache assignments as this is now handled in the manager during assignments.


vthacker (approved):
nice!

PR #704, opened 2023-10-19
Switch from executors to virtual threads & structured concurrency where possible

bryanlb:
###  Summary

This switches a large amount of ThreadPoolExecutors, ListenableFutures, and CompletableFutures to use virtual threads and structured concurrency. As part of this work fixed executors pools have been refactored to use Semaphores where appropriate, and several synchronized blocks have been replaced with ReentrantLocks to enable additional use of virtual threads in the future.

The bulk of this rewrite focuses around the query path, the chunk and distributed query concurrency. Also included are a few improvements to query path visibility (tracing and error results).


kaldb/src/main/java/com/slack/kaldb/chunkManager/ChunkManagerBase.java
```diff
@@ -107,72 +84,88 @@ public SearchResult<T> query(SearchQuery query, Duration queryTimeout) {
...
+                                })))
+                .toList();
         try {
-          Futures.allAsList(queries).get(0, TimeUnit.SECONDS);
-        } catch (Exception e) {
-          throw new IllegalArgumentException(e);
+          scope.joinUntil(Instant.now().plusSeconds(queryTimeout.toSeconds()));
+        } catch (TimeoutException timeoutException) {
+          scope.shutdown();
+          scope.join();
```
vthacker, line 114: maybe we can add some docs as to why we call `join` here? This definitely is something I hadn't realized in the past :)

vthacker (approved):

PR #708, opened 2023-10-24
Rewrite Lucene queries for better caching compatibility

bryanlb:
###  Summary

This moves to rewriting the Lucene queries such that when the query window exceeds the start, end, or both those clauses are not included in the query provided to Lucene. This allows for better optimization of the query, and should improve cache hits.

Note: For Lucene it only completely removes the query if both are null, otherwise sets them to max/min long.


kaldb/src/main/java/com/slack/kaldb/logstore/search/LogIndexSearcher.java
```diff
@@ -5,5 +5,5 @@
 
 public interface LogIndexSearcher<T> extends Closeable {
   SearchResult<T> search(
-      String dataset, String query, long minTime, long maxTime, int howMany, AggBuilder aggBuilder);
+      String dataset, String query, Long minTime, Long maxTime, int howMany, AggBuilder aggBuilder);
```
vthacker, line 8: Should these two variables also be called startTimeMsEpoch/endTimeMsEpoch?

bryanlb: As discussed we'll leave these as-is for now.

vthacker (approved):

PR #710, opened 2023-10-25
Refactor Zipkin API, fix duration bug

bryanlb:
###  Summary

Refactors the Zipkin API to use custom response objects. This brings our response in line with the Zipkin spec (https://zipkin.io/zipkin-api/#/), as directly serializing the proto to string resulted in losing some type information (see https://github.com/protocolbuffers/protobuf/issues/1823). This overall result is much cleaner than the previous as well, a we can rely on Jackson for complete object serialization instead of string concatenation. 

The driving factor for this change however was to fix the duration bug - previously duration was reported in millis instead of micros, which caused the results to be off by a factor of x1000.

Screenshot 2023-10-25 at 11 27 02 AM
Screenshot 2023-10-25 at 11 23 04 AM


kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinSpanResponse.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+/**
+ * Span response object for Zipkin API
+ *
+ * @see <a href="https://zipkin.io/zipkin-api/#/">Zipkin API Spec</a>
+ */
+@SuppressWarnings("unused")
```
vthacker, line 11:
```
record ZipkinSpanResponse(String id, String traceId, String parentId, String name, Long timestamp,
                          String localEndpoint, ZipkinEndpointResponse remoteEndpoint,
                          int duration, String kind,  Map<String, String> tags) { }
```

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
```diff
@@ -92,24 +95,20 @@ protected static String convertLogWireMessageToZipkinSpan(List<LogWireMessage> m
...
-              Optional.ofNullable(parentId),
-              message.getId(),
-              Optional.ofNullable(name),
-              Optional.ofNullable(serviceName),
-              messageConvertedTimestamp,
-              duration,
-              messageTags);
-      String spanJson = printer.print(span);
-      traces.add(spanJson);
+      final ZipkinSpanResponse span = new ZipkinSpanResponse(message.getId(), messageTraceId);
```
vthacker, line 98:
```
final ZipkinSpanResponse span = new ZipkinSpanResponse(message.getId(), messageTraceId, parentId, name,
              convertToMicroSeconds(message.getTimestamp()), null, remoteEndpoint, Math.toIntExact(duration), null, messageTags);
```

vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/zipkinApi/ZipkinService.java
bryanlb: We elected not to go with this approach as we lose the benefit of required/optional field enforcement. Today the constructor requires both traceId and id, but the rest of the fields are optional.

PR #711, opened 2023-10-25
Fix rounding error in HPA autoscaler

bryanlb:
###  Summary

Fixes a rounding issue in the current HPA autoscaling logic when running in sufficiently large clusters. This HPA approach has some significant limitations as compared to a proper Kubernetes operator, which may be the route we end up taking long term.


vthacker (approved):
LGTM

PR #719, opened 2023-10-31
Fix bug in bulk ingest API when document has special chars

vthacker:
###  Summary

This was a fun one to track down - We need to pass the correct string length which needs to take into account special chars


kaldb/src/main/java/com/slack/kaldb/preprocessor/ingest/OpenSearchBulkApiRequestParser.java
```diff
@@ -105,8 +105,8 @@ public static List<IndexRequest> parseBulkRequest(String postBody) throws IOExce
     List<IndexRequest> indexRequests = new ArrayList<>();
     BulkRequest bulkRequest = new BulkRequest();
     // calls parse under the hood
-    bulkRequest.add(
-        postBody.getBytes(StandardCharsets.UTF_8), 0, postBody.length(), null, XContentType.JSON);
+    byte[] bytes = postBody.getBytes(StandardCharsets.UTF_8);
+    bulkRequest.add(bytes, 0, bytes.length, null, MediaTypeRegistry.JSON);
```
vthacker, line 109: fix was to pass `bytes.length` instead of `postBody.length()` . The test will fail without the fix

bryanlb (approved):

PR #726, opened 2023-11-01
Logging adjustments, error prone bump

bryanlb:
###  Summary

This audits our existing logging for any potential PII that may be logged out by KalDB. Additionally it reduces several log lines that shouldn't be needed at LOG level anymore down to DEBUG. 

This also includes a bump to our error prone version and a few additional checks & related fixes. I was looking into options for error prone that would help with log consistency, but the only ones currently available target Flogger, which we don't use.

Also see https://github.com/slackhq/kaldb/issues/727


kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
```diff
@@ -114,7 +114,7 @@ public IndexingChunkManager(
       MeterRegistry registry,
       BlobFs blobFs,
       String s3Bucket,
-      ListeningExecutorService rollOverExecutorService,
+      ListeningExecutorService rolloverExecutorService,
```
vthacker, line 117: any reason why we change 
```
rollOverExecutorService to rolloverExecutorService,
```

but in the previous file we use camelCase?

```
chunkRolloverStrategy -> chunkRollOverStrategy
```

kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -97,7 +97,6 @@ private EsSearchResponse doSearch(KaldbSearch.SearchRequest searchRequest) {
     KaldbSearch.SearchResult searchResult = searcher.doSearch(searchRequest);
 
     span.tag("requestDataset", searchRequest.getDataset());
-    span.tag("requestQueryString", searchRequest.getQueryString());
```
vthacker, line 100: good one!

kaldb/src/main/java/com/slack/kaldb/chunkManager/IndexingChunkManager.java
bryanlb: The rule for errorprone is they just need to be consistent, so I fixed the locations they flagged. Personally I don't like splitting rollover into two words but that's for another day. :D

vthacker (approved):

PR #735, opened 2023-11-16
Transaction batching Kafka producer

bryanlb:
###  Summary

This introduces a new kafka producer that can batch bulk ingest request into larger transactional blocks using a single threaded consumer.

Also: 
* Adds supports support for Kafka additional properties on bulk ingest
* Adds metrics for bulk ingest
* Switches from blocking to virtual thread implementation for better throughput
* Fixes java format to include support for new syntax (https://github.com/spotify/fmt-maven-plugin/pull/185)
* Refactors bulk ingest class structure to match other API endpoints


kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestApi.java
```diff
@@ -0,0 +1,103 @@
...
+      for (Map.Entry<String, List<Trace.Span>> indexDocs : docs.entrySet()) {
+        final String index = indexDocs.getKey();
+        if (!datasetRateLimitingService.tryAcquire(index, indexDocs.getValue())) {
+          BulkIngestResponse response = new BulkIngestResponse(0, 0, "rate limit exceeded");
+          future.complete(HttpResponse.ofJson(TOO_MANY_REQUESTS, response));
+          return HttpResponse.of(future);
+        }
+      }
+
+      Thread.ofVirtual()
```
vthacker, line 81: Discussed this offline, could we for now just leave a comment saying long term we might want to remove the `Thread.ofVirtual` from here and just have the blocking task executor be be backed by virtual threads.

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestKafkaProducer.java
```diff
@@ -0,0 +1,298 @@
...
+        "Kafka bootstrapServers must be provided");
+
+    checkArgument(!kafkaConfig.getKafkaTopic().isEmpty(), "Kafka topic must be provided");
+
+    this.datasetMetadataStore = datasetMetadataStore;
+    this.pendingRequests = new LinkedBlockingQueue<>();
+
+    // todo - consider making this a configurable value or removing the config
+    this.producerSleep =
+        Integer.parseInt(System.getProperty("kalDb.bulkIngest.producerSleep", "50"));
```
vthacker, line 91: perhaps we can add a unit suffix?

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestKafkaProducer.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+  }
+
+  public BulkIngestRequest submitRequest(Map<String, List<Trace.Span>> inputDocs) {
+    BulkIngestRequest request = new BulkIngestRequest(inputDocs);
+    pendingRequests.add(request);
+    return request;
+  }
+
+  protected Map<BulkIngestRequest, BulkIngestResponse> transactionCommit(
```
vthacker, line 159: maybe a better name could be `produceDocsAndCommit` ?

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestKafkaProducer.java
```diff
@@ -0,0 +1,298 @@
...
+      // KafkaProducer does not allow creating multiple transactions from a single object -
+      // rightfully so.
+      // Till we fix the producer design to allow for multiple /_bulk requests to be able to
+      // write to the same txn
+      // we will limit producing documents 1 thread at a time
+      for (Trace.Span doc : indexDoc.getValue()) {
+        ProducerRecord<String, byte[]> producerRecord =
+            new ProducerRecord<>(kafkaConfig.getKafkaTopic(), partition, index, doc.toByteArray());
+
+        // we intentionally supress FutureReturnValueIgnored here in errorprone - this is because
```
vthacker, line 250: It was a typo initially as well `s/supress/suppress` :)

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestKafkaProducer.java
bryanlb: I ended up adding a suffix to the variable name so it matches our current configs approach. I think long term we would benefit with a duration implementation that allows us to use arbitrary units in our configs (ie `50ms`), but a problem for another day.

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/BulkIngestKafkaProducer.java
bryanlb: Done - `produceDocumentsAndCommit`

vthacker (approved):

PR #760, opened 2024-02-01
parse timestamp field safely

vthacker:
###  Summary

1. Parse @timestamp first since that should take precendece.
2. parse timestamp field in a try-catch since we make the assumption it will be a long and in millis


bryanlb (changes requested):
We got an extra test assert to go along with this? 😁

bryanlb (approved):

PR #762, opened 2024-02-06
Index the `_id` field with encoding to be searchable

vthacker:
###  Summary

We rely on OpenSearch to query for documents. Which means we translate our schema to their mappings API and then let the query parser form the query.

For `_id` fields, the query parser encodes the query string value ( which it also does during indexing )

This PR adds an ID field, currently just for the `_id` field to use and encodes the value so that queries work on it.

We want the least resistance in reworking our schema layer to play nicely with OpenSearch. In a future effort we have to rethink this and redo a lot of how to make fields configurable, how can it sit closer to OpenSearch etc.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldType.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldType.java
@@ -28,6 +28,7 @@ import org.apache.lucene.index.Term;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.util.BytesRef;
+import org.opensearch.index.mapper.Uid;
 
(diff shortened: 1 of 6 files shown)


bryanlb (approved):

kaldb/src/main/java/com/slack/kaldb/metadata/schema/FieldType.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.util.BytesRef;
+import org.opensearch.index.mapper.Uid;
```
bryanlb, line 31: This is something we'll have to consider as we design the new schema approach.

bryanlb (approved):

PR #764, opened 2024-02-07
Add detection of invalid recovery tasks on recovery nodes

bryanlb:
###  Summary

Add detection for previously created bad recovery tasks. This should be solved in several places, but should unblock current deploys were bad recovery tasks exist.


kaldb/src/main/java/com/slack/kaldb/server/RecoveryTaskCreator.java
```diff
@@ -218,6 +218,8 @@ public long determineStartingOffset(
       } else if (indexerConfig.getCreateRecoveryTasksOnStart()
           && indexerConfig.getReadFromLocationOnStart()
               == KaldbConfigs.KafkaOffsetLocation.LATEST) {
+        // todo - this appears to be able to create recovery tasks that have a start and end
```
vthacker, line 221: could we change the comment to specify this happens during next cluster creation?

bryanlb: Done! Also added a detail that it seems to be worse with indexers that are created ant not immediately used.

vthacker (approved):

PR #765, opened 2024-02-07
make rate limit error code configurable

vthacker:
###  Summary

We want to be able to configure how rate limit error HTTP code is sent back to clients. 

Some clients might want to drop logs ( say to prioritize fresh logs ) and some might retry. So making the code configurable


kaldb/src/test/java/com/slack/kaldb/server/BulkIngestApiTest.java
```diff
@@ -208,49 +217,30 @@ public void testBulkApiBasic() throws Exception {
...
+    try {
+      BulkIngestResponse httpResponseObj =
+          JsonUtil.read(httpResponse.contentUtf8(), BulkIngestResponse.class);
+      assertThat(httpResponseObj.totalDocs()).isEqualTo(1);
+      assertThat(httpResponseObj.failedDocs()).isEqualTo(0);
+    } catch (IOException e) {
+      fail("", e);
+    }
+
+    httpResponse = bulkApi.addDocument(request1).aggregate().join();
```
vthacker, line 232: this test was failing previously ( even without any of my changes ).

The reason this still worked was we don't implement `handle` and the AssertionError gets caught by the default `handle` and gets suppressed.

bryanlb (approved):

PR #769, opened 2024-02-12
Refactor read only chunk timeout

bryanlb:
###  Summary

Refactors a deprecated class-level timeout, in favor of another global deprecated timeout. We need to remove both, but this prevents us from having separate values - especially since the one is set so aggressively.


bryanlb:
Combined with #768

PR #772, opened 2024-02-14
Add persistent ephemeral node support to metadata store

bryanlb:
###  Summary

This adds an optional **opt-in** mode to run the `KaldbMetadataStore` using a combination of `PersistentNode` and `PersistentWatcher` for ephemeral nodes. This system property flag will be removed once the logic is validated, at which point this will no longer be able to be opt-out.

The goal of this PR is to support using ephemeral nodes across ZK session disconnects (https://github.com/slackhq/kaldb/pull/644#discussion_r1295013333), and uses the Curator `PersistentNode` in order to recreate nodes when they disappear, and `PersistentWatcher` to subscribe to updates to ephemeral nodes by other instances (ie - manager updating cache/recovery ephemerals). 

https://curator.apache.org/docs/recipes-persistent-node
https://curator.apache.org/docs/recipes-persistent-watcher

This PR also includes:
* Passing the meter registry into the metadata stores - in part for testing of this PR, but it will also enable us to start adding meters where appropriate to this class.
* Fixes to both unclosed metadata stores when they should have been closed (mostly tests), and fixes to closing metadata stores when they shouldn't have been (ie, passed as a param)
* Deduplication of events on the recovery service. This should probably have a further refactor to match the approach used by the manager, with an eventAggregationWindow given that ZK is _eventually consistent_.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadataStore.java
+++ b/kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadataStore.java
@@ -4,6 +4,7 @@ import com.google.common.util.concurrent.JdkFutureAdapters;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.slack.kaldb.metadata.core.KaldbPartitioningMetadataStore;
 import com.slack.kaldb.proto.metadata.Metadata;
+import io.micrometer.core.instrument.MeterRegistry;
 import java.time.Instant;
 import org.apache.curator.x.async.AsyncCuratorFramework;
 import org.apache.zookeeper.CreateMode;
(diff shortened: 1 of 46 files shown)


kaldb/src/main/java/com/slack/kaldb/metadata/cache/CacheSlotMetadataStore.java
```diff
@@ -15,12 +16,14 @@ public class CacheSlotMetadataStore extends KaldbPartitioningMetadataStore<Cache
    * Initializes a cache slot metadata store at the CACHE_SLOT_ZK_PATH. This should be used to
    * create/update the cache slots, and for listening to all cache slot events.
    */
-  public CacheSlotMetadataStore(AsyncCuratorFramework curatorFramework) throws Exception {
+  public CacheSlotMetadataStore(AsyncCuratorFramework curatorFramework, MeterRegistry meterRegistry)
```
mansu, line 19: A large part of this diff if adding `meterRegistry` parameter. It would be far more easier to review the diff if `meterRegistry` change is done in a separate PR.

kaldb/src/main/java/com/slack/kaldb/metadata/core/KaldbMetadataStore.java
```diff
@@ -23,18 +26,31 @@
...
  *
  * <p>Every method provides an async and a sync API. In general, use the async API you are
  * performing batch operations and a sync if you are performing a synchronous operation on a node.
+ *
+ * <p><a href="https://curator.apache.org/docs/recipes-persistent-node">Persistent node recipie</a>
  */
 public class KaldbMetadataStore<T extends KaldbMetadata> implements Closeable {
+  private static final Logger LOG = LoggerFactory.getLogger(KaldbMetadataStore.class);
+
+  public static String PERSISTENT_EPHEMERAL_PROPERTY = "kaldb.metadata.persistentEphemeral";
```
mansu, line 43: It seems that `Persistent Node` is a lot of complexity esp when we can use a non-ephermeral node add for cleaner tests. Can we use a regular node instead since this is mostly needed for HPA?

PR #773, opened 2024-02-16
Add re-entrant lock protection around adding message

bryanlb:
###  Summary

This should address an exception we occasionally see, where the `addDocument` fails due to the IndexWriter already being closed.

```java
org.apache.lucene.store.AlreadyClosedException: this IndexWriter is closed
	at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:908)
	at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:921)
	at org.apache.lucene.index.IndexWriter.updateDocuments(IndexWriter.java:1542)
	at org.apache.lucene.index.IndexWriter.updateDocument(IndexWriter.java:1830)
	at org.apache.lucene.index.IndexWriter.addDocument(IndexWriter.java:1470)
	at com.slack.kaldb.logstore.LuceneIndexStoreImpl.addMessage(LuceneIndexStoreImpl.java:261)
	at com.slack.kaldb.logstore.LuceneIndexStoreImpl.addMessage(LuceneIndexStoreImpl.java:42)
	at com.slack.kaldb.chunk.ReadWriteChunk.addMessage(ReadWriteChunk.java:153)
	at com.slack.kaldb.chunkManager.IndexingChunkManager.addMessage(IndexingChunkManager.java:179)
	at com.slack.kaldb.writer.LogMessageWriterImpl.insertRecord(LogMessageWriterImpl.java:106)
	at com.slack.kaldb.writer.kafka.KaldbKafkaConsumer.consumeMessages(KaldbKafkaConsumer.java:234)
	at com.slack.kaldb.writer.kafka.KaldbKafkaConsumer.consumeMessages(KaldbKafkaConsumer.java:199)
	at com.slack.kaldb.server.KaldbIndexer.run(KaldbIndexer.java:137)
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1.lambda$doStart$1(AbstractExecutionThreadService.java:59)
	at com.google.common.util.concurrent.Callables.lambda$threadRenaming$3(Callables.java:105)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```


vthacker (approved):

kaldb/src/main/java/com/slack/kaldb/logstore/LuceneIndexStoreImpl.java
```diff
@@ -255,6 +261,7 @@ private void handleNonFatal(Throwable ex) {
 
   @Override
   public void addMessage(LogMessage message) {
+    pendingWritesPhaser.register();
```
mansu, line 264: Would adding a phaser impact here the ingestion throughput? Do we have any extra logging around the message write failure like the actual message(timestamp etc..) and the chunk metadata? From the PR description if the chunk is being closed while documents are being added, makes it sound like a higher level chunk manager logic issue. Could it be late arriving data or a bug in chunk selection logic? The extra logging can help determine that.

bryanlb: This does not impact ingestion, as the [phaser](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/Phaser.html) is non-blocking in this approach. You can see a multithreaded demo in the provided `phaserDemoTest.`

The issue here is the close method is being invoked via the `ChunkManagerBase`, which uses an `AbstractIdleService`. This runs in a separate thread than that of the `addMessage`, and when throughput is high enough there is a race condition in the existing indexing logic around these lines:

```java
@Override
public void addMessage(LogMessage message) {
  try {
      if (indexWriter.isPresent()) {
        indexWriter.get().addDocument(documentBuilder.fromMessage(message));
      } else {
 ...
 
@Override
public void close() {
...
  try {
    indexWriter.get().close();
  } catch (IllegalStateException | IOException | NoSuchElementException e) {
    LOG.error("Error closing index " + id, e);
  }
  indexWriter = Optional.empty();     
```

As you can see from the provided stacktrace the current issue is `indexWriter` is present, but is already closed. The only scenario this happens is when close is actively running in a separate thread to `addMessage` and has shut down the index writer but not set it to an optional empty.

A phaser here is actually a bandaid around the fact the `LuceneIndexStoreImpl` class is not threadsafe. A broader rewrite to make these indexer classes threadsafe is preferred, but this PR is in case we continue to see indexer concurrency issues.

mansu: I am a bit confused about how we arrive at this condition.

IIRC, the ingest path works as follows (unsure if it's changed now, but code looks the same) is we create an active chunk and add messages to it. We keep it active until a rollover and close it about ~15 minutes after the chunk is rolled over. The delay helps with uploading the chunk to the remote store and also to account for any late arriving messages. If any messages arrive late we drop them since they are outside the window. 

Since there is so much delay between `addMessage` call and the `close()` call, I am confused why messages are being added this late to a chunk. Per application logic we should never have arrived at this state, since the chunkManager shouldn't be adding a message to this chunk to start with?

mansu: Also, it would be worth running the benchmark with and without the change to confirm that there is no impact to ingestion throughput.

PR #776, opened 2024-02-20
Parse binary data as-is

vthacker:
###  Summary

Some fields that are nested in the request, get encoded as binary data. This PR changes that to atleast be the string representation of the values.

In the future we want to support nested fields, though we have some support today it's not fully baked in to the bulk request parsing layer

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
+++ b/kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
@@ -195,7 +195,7 @@ public class SpanFormatter {
       } else if (valueType == 3) {
         jsonMap.put(key, tag.getVFloat64());
       } else if (valueType == 4) {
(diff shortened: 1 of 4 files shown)


kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
```diff
@@ -195,7 +195,7 @@ public static LogMessage toLogMessage(Trace.Span span) {
       } else if (valueType == 3) {
         jsonMap.put(key, tag.getVFloat64());
       } else if (valueType == 4) {
-        jsonMap.put(key, encodeBinaryTagValue(tag.getVBinary()));
+        jsonMap.put(key, tag.getVBinary().toStringUtf8());
```
vthacker, line 198: the fix - otherwise an input of `[]` would get encoded to ` W10=`

bryanlb (approved):

PR #778, opened 2024-02-22
Allow for user defined schema and enrich docs in the preprocessor 

vthacker:
Summary
Part 4 and 5 - Introduce a schema, add info to the documents(Trace.Span) in the preprocesssor

The preprocessor today creates Trace.Span objects. In this PR we will add field type info for each field and we want the ability to reference that to create a lucene document. That will be the next PR

The schema file from the file system, however we have future plans to put this file in ZK and expose mappings This also eliminates the need to add field type to the proto objects.

**Kaldb Schema PR breakdown** - 5 part series to introduce schema configs to Kaldb

1. introduce a concept of int and float to trace.proto
2. remove the concept of index transformer - it's always trace.span
3. use trace.span to create lucene documents instead of LogMessage
4. insert schema info to trace.proto in the preprocessor
5. leverage schema in the indexer


kaldb/src/main/proto/trace.proto
```diff
@@ -29,8 +31,10 @@ message KeyValue {
   string    v_str     = 3;
   bool      v_bool    = 4;
   int64     v_int64   = 5;
+  int32     v_int32   = 9;
   double    v_float64 = 6;
-  bytes     v_binary  = 7;
+  float     v_float32 = 10;
+  slack.proto.kaldb.schema.SchemaFieldType fieldType = 8;
```
mansu, line 37: Isn't schema a property of the dataset? If so, why is it in a `KeyValue` type?

kaldb/src/main/proto/trace.proto
```diff
@@ -18,9 +19,10 @@ import "google/protobuf/empty.proto";
 enum ValueType {
   STRING  = 0;
   BOOL    = 1;
+  INT32   = 5;
   INT64   = 2;
   FLOAT64 = 3;
-  BINARY  = 4;
```
mansu, line 23: Is there a reason to drop the Binary type? I think it has a lot of value to keep raw data unmodified in Lucene. We should keep it.

kaldb/src/main/proto/trace.proto
```diff
@@ -18,9 +19,10 @@ import "google/protobuf/empty.proto";
...
+  FLOAT32 = 6;
```
mansu, line 25: Since Lucene doesn't support 32 bit or 64 bit values, why do we need 32 bit values in the value type here?

kaldb/src/main/proto/trace.proto
```diff
@@ -67,6 +71,8 @@ message Span {
 
   // A list of key value pairs.
   repeated KeyValue tags = 7;
+
+  string service_name = 8;
```
mansu, line 75: Is there a reason to make `service_name` a first class citizen and not a tag?

kaldb/src/main/proto/schema.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+
+package slack.proto.kaldb.schema;
+
+option java_package = "com.slack.kaldb.proto.schema";
+
+message KaldbSchema {
```
mansu, line 7: Instead of a `KalDBSchema` a better name would be `DatasetSchema` and associate a schema with a dataset.

kaldb/src/test/resources/schema/test_schema.yaml
```diff
@@ -0,0 +1,27 @@
+fields:
+  host:
+    type: KEYWORD
```
vthacker, line 3: `target_type`

kaldb/src/main/proto/trace.proto
vthacker: Schema will be part of the dataset. That will continue to be in the indexer. Here we just want to inject field type info from the preprocessor that the indexer can leverage.

kaldb/src/main/proto/trace.proto
vthacker: fixed!

kaldb/src/main/proto/schema.proto
vthacker: yup. Will change

kaldb/src/main/proto/trace.proto
vthacker: I think it does? https://lucene.apache.org/core/9_5_0/core/org/apache/lucene/document/Field.html

mansu: Can you please point to the specific field type where 32 bit ints or floats are specially handled?

kaldb/src/main/proto/trace.proto
```diff
@@ -35,6 +36,7 @@ message KeyValue {
   bytes     v_binary  = 7;
   int32     v_int32   = 8;
   float     v_float32 = 9;
+  slack.proto.kaldb.schema.SchemaFieldType fieldType = 10;
```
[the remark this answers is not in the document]
vthacker: > Also, drop the 32 bit fields since they are not needed?

Why are they not needed?

kaldb/src/main/proto/schema.proto
```diff
@@ -0,0 +1,35 @@
...
+  // other field definitions in the future
+}
+
+// https://opensearch.org/docs/2.4/opensearch/supported-field-types/index/
+// Add the remaining types as needed
+enum SchemaFieldType {
+  KEYWORD  = 0;
+  TEXT  = 1;
+  IP  = 2;
+  // expose format in the future
```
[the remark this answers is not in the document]
vthacker: sounds good!

kaldb/src/main/proto/trace.proto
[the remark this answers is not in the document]
mansu: Lucene doesn't have special handling for 32 bit fields. So, it's best to leave those fields from the trace.proto.

config/schema.yaml
```diff
@@ -0,0 +1 @@
+fields:
```
bryanlb, line 1: Let's add a commented-out example or two?

kaldb/src/main/java/com/slack/kaldb/metadata/schema/SchemaUtil.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.commons.text.StringSubstitutor;
+import org.apache.commons.text.lookup.StringLookup;
+
+public class SchemaUtil {
+
+  public static Schema.PreprocessorSchema parseSchema(Path schemaPath) throws IOException {
+    String filename = schemaPath.getFileName().toString();
+    if (filename.endsWith(".yaml")) {
+      return parseSchemaYaml(Files.readString(schemaPath), System::getenv);
+    } else if (filename.endsWith(".json")) {
```
bryanlb, line 22: We can leave the json for now, but I think we'll be removing json support more broadly moving forward for our config files.

kaldb/src/main/java/com/slack/kaldb/metadata/schema/SchemaUtil.java
```diff
@@ -0,0 +1,51 @@
...
+
+    Object obj = yamlReader.readValue(substitute.replace(yamlStr), Object.class);
+    return parseJsonSchema(jsonWriter.writeValueAsString(obj));
+  }
+
+  @VisibleForTesting
+  public static Schema.PreprocessorSchema parseJsonSchema(String jsonStr)
+      throws InvalidProtocolBufferException {
+    Schema.PreprocessorSchema.Builder kaldbSchemaBuilder = Schema.PreprocessorSchema.newBuilder();
+    JsonFormat.parser().merge(jsonStr, kaldbSchemaBuilder);
```
bryanlb, line 46: What happens if this (or the yaml) version throw an exception (ie, malformed)? Do we continue without a schema or do we fatal?

kaldb/src/main/proto/kaldb_configs.proto
```diff
@@ -274,4 +274,6 @@ message PreprocessorConfig {
   // Set this to 429 for clients to retry the request after a delay
   // Only used when we use the bulk API
   int32 rate_limit_exceeded_error_code = 11;
+
+  string schema_file = 12;
```
bryanlb, line 278: This should have a comment for what it's intended use case is - even if this is just "temporary" until we move to ZK.

kaldb/src/main/proto/schema.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
+package slack.proto.kaldb.schema;
+
+option java_package = "com.slack.kaldb.proto.schema";
+
+// We want to call this Schema and then remove the .schema package
+// But if we do that today, the field types interfere with existing enums
+// which we don't want addressing just yet ( kaldb_search.proto )
+message PreprocessorSchema {
```
bryanlb, line 10: Since this is consumed by both indexer and preprocessor, maybe "IngestSchema" is a better fit? Up to you.

kaldb/src/test/java/com/slack/kaldb/schema/SchemaConfigTest.java
```diff
@@ -0,0 +1,97 @@
...
+import com.slack.kaldb.proto.schema.Schema;
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class SchemaConfigTest {
+
+  @Test
+  public void testParsingEmptySchema() {
```
bryanlb, line 16: We should add one for a malformed schema as well, and one that points to a file that doesn't exist.

kaldb/src/test/resources/schema/test_schema.yaml
```diff
@@ -0,0 +1,27 @@
+fields:
```
bryanlb, line 1: Some thoughts of how this can evolve over time:

```yaml
default:
  type: TEXT
  fields:
    keyword:
      type: KEYWORD
fields:
  host:
    type: TEXT
    fields: // derived fields (constraint - must be prefixed with original field name)
      keyword:
        type: KEYWORD       // host.keyword
        ignoreAbove: 256
  responseCode:       // responseCode.int
    type: TEXT
    fields:
      int:
        type: INTEGER
  message:
    type: TEXT
...
```

bryanlb: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-params.html

config/schema.yaml
vthacker: I'll do that within the docs folder. We'll be evolving it to ZK most likely so I am leaning to not adding it to the main repo yet.

kaldb/src/main/java/com/slack/kaldb/metadata/schema/SchemaUtil.java
vthacker: nice! added a test for that

PR #780, opened 2024-02-22
We should enforce these required inputs by returning a useful error and 400 response. right now it is 500

autata:
###  Summary
I think we should set a reasonable default for when size is missing (10? - i think that is what ES uses when missing). Opening a draft PR as a discussion before going further in this direction.  There may be a better way to more closely follow the ES API spec. ([_msearch](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-multi-search.html), [range query docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html))

For `lte` and `gte` I think if they are missing, we should return a 400 (Bad Request) and a helpful error string.

Right now, if any of these are missing, we get a 500 (Internal Server Error) response.

### Requirements

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
+++ b/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
@@ -85,14 +85,20 @@ public class OpenSearchRequest {
     return queryString;
   }
 
+  // This is required. consider returning a more useful error to the user than 500
+  // or setting a reasonable default (default seems a good option here)
   private static int getHowMany(JsonNode body) {
     return body.get("size").asInt();
   }
 
+  // This is required. consider returning a more useful error to the user than 500
+  // or setting a reasonable default (not sure what that would be)
   private static long getStartTimeEpochMs(JsonNode body) {
     return body.get("query").findValue("gte").asLong();
   }
 
+  // This is required. consider returning a more useful error to the user than 500
+  // or setting a reasonable default (not sure what that would be)
   private static long getEndTimeEpochMs(JsonNode body) {
     return body.get("query").findValue("lte").asLong();
   }


kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/OpenSearchRequest.java
```diff
@@ -85,14 +85,20 @@ private static String getQueryString(JsonNode body) {
     return queryString;
   }
 
+  // This is required. consider returning a more useful error to the user than 500
+  // or setting a reasonable default (default seems a good option here)
   private static int getHowMany(JsonNode body) {
     return body.get("size").asInt();
   }
 
+  // This is required. consider returning a more useful error to the user than 500
```
autata, line 94: probably something like this:
```
probably should return something like this instead of 500
{
"error": {
"code": 400,
"message": "Bad Request",
"details": "The 'query.gte' field is required but missing in the request."
}
}
```

bryanlb:
@autata we are looking to next quarter or the following to move to using the Opensearch library for the entire request parsing, which should allow us better compatibility with these fields and options. Currently we do this for a subsection of the request (aggregations), and would like to extend this to the entire request body. This should allow us to support more complex request types that are not currently issued via Grafana today.

autata:
@bryanlb this is helpful context. Do you think it makes sense to open a PR with slightly better return codes (as suggested in these comments) or potentially change this to an issue since following Opensearch API is in the roadmap for the next quarter or two and this would be fixed if moving to Opensearch full functionality.

bryanlb:
@autata I think that moving this to an issue makes the most sense here. If you still want to PR a change to these response codes that would be fine, but if all goes according to plan this would only be in place for a few months before we can entirely rework the request parsing.

autata:
@bryanlb opened an issue for tracking: https://github.com/slackhq/astra/issues/798

PR #783, opened 2024-02-28
Remove transformer in the indexer

vthacker:
###  Summary

Part 2 - Remove transformer in the indexer and always assume incoming data is a Trace.Span object.

We will revisit bringing your own transformer if future project users would find it useful. When we revisit, we need to think of that Transformer directly converting the input bytes to lucene document i.e the transformer should be aware of schema of their documents and also expose it as an API ( so that our query service mapping API can leverage it )

**Kaldb Schema PR breakdown** - 5 part series to introduce schema configs to Kaldb

1. introduce a concept of int and float to trace.proto
2. remove the concept of index transformer - it's always trace.span
3. use trace.span to create lucene documents instead of LogMessage
4. insert schema info to trace.proto in the preprocssor
5. leverage schema in the indexer


kaldb/src/test/java/com/slack/kaldb/testlib/MessageUtil.java
```diff
@@ -169,9 +170,60 @@ public static List<LogMessage> makeMessagesWithTimeDifference(
     return result;
   }
 
-  // TODO: Move this to TestKafkaServer class.
-  public int getPort() throws IOException {
-    ServerSocket socket = new ServerSocket(0);
-    return socket.getLocalPort();
+  public static List<Trace.Span> makeMessagesWithTimeDifferenceTS(
```
vthacker, line 173: by the end of the next PR all tests will start using this and we'll be able to remove `makeMessagesWithTimeDifference` and rename this method

benchmarks/src/main/java/com/slack/kaldb/IndexAPILog.java
```diff
@@ -1,187 +1,188 @@
...
-          "testKey",
-          testMurronMsg.toByteString().toByteArray());
-    } catch (Exception e) {
-      System.out.println("skipping - cannot parse input" + e);
-      skipCount++;
-      return null;
-    }
-  }
-}
+// package com.slack.kaldb;
```
bryanlb, line 1: Instead of commenting these out delete these files?

kaldb/src/main/proto/trace.proto
```diff
@@ -21,6 +21,8 @@ enum ValueType {
   INT64   = 2;
   FLOAT64 = 3;
   BINARY  = 4;
+  INT32   = 5;
```
bryanlb, line 24: Split to #782

kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
```diff
@@ -196,6 +196,10 @@ public static LogMessage toLogMessage(Trace.Span span) {
         jsonMap.put(key, tag.getVFloat64());
       } else if (valueType == 4) {
         jsonMap.put(key, tag.getVBinary().toStringUtf8());
+      } else if (valueType == 5) {
```
bryanlb, line 199: We should add some extra asserts in the `SpanFormatterTest` for these types.

bryanlb (changes requested):
Couple small things to address, but like the direction this is going. 👍

kaldb/src/main/java/com/slack/kaldb/writer/SpanFormatter.java
vthacker: I'm going to hold off doing anything on this. I feel like we need more tests that do the following - Take Spans and convert them to lucene documents and assert field types

SpanFormatterTest seems to convert Spans to LogMessage which we are getting rid of in the next PR

bryanlb: Ah, that makes sense. 👍

bryanlb (approved):

bryanlb (approved):

benchmarks/src/main/java/com/slack/kaldb/QueryBenchmark.java
```diff
@@ -1,169 +0,0 @@
...
-import org.apache.kafka.common.record.TimestampType;
-import org.openjdk.jmh.annotations.Benchmark;
-import org.openjdk.jmh.annotations.Level;
-import org.openjdk.jmh.annotations.Scope;
-import org.openjdk.jmh.annotations.Setup;
-import org.openjdk.jmh.annotations.State;
-import org.openjdk.jmh.annotations.TearDown;
-
-@State(Scope.Thread)
-public class QueryBenchmark {
```
mansu, line 42: Why is benchmark code deleted as part of this change?

vthacker: good question! We'll be brining them back once some of our current schema work is done. We haven't used these in years and some of it needs to be redone anyways

We also want to remove stuff like api log / murron and other slack specific constructs before our GA

PR #815, opened 2024-03-19
Docs initial tree structure, setup general configs

bryanlb:
###  Summary

Sets up the initial tree structure for the docs site, adds links/logos/etc to header & footer that were missing from the initial template.

The pages are mostly empty but should allow folks to start PR'ing content as appropriate.

image


docs/astra.tree
```diff
@@ -5,7 +5,30 @@
...
+            <toc-element topic="Initial-RFC.md"/>
+        </toc-element>
+    </toc-element>
+    <toc-element topic="Dependencies.md">
+        <toc-element topic="Kafka.md"/>
+        <toc-element topic="Zookeeper.md"/>
+    </toc-element>
+    <toc-element topic="Troubleshooting.md"/>
     <toc-element topic="Roadmap.md"/>
 </instance-profile>
```
vthacker, line 34: can we just add the newlines :)

bryanlb: Hah, this is a generated file so I don't believe manual changes would persist. 😢

vthacker (approved):

PR #817, opened 2024-03-20
remove top level fields from tags(if present)

vthacker:
###  Summary

Tags is a list, and if any top level trace.proto fields ends up existing as tags we'll try to recreate the field twice in the lucene document leading to an error ( cannot index same field twice )

This is a quick and dirty way to fix it. We need to evolve trace.proto to be a better fit for our internal data transport so we'll revisit this again

### DIFF
--- a/kaldb/src/test/java/com/slack/kaldb/schema/SpanTagFormatterWithSchemaTest.java
+++ b/kaldb/src/test/java/com/slack/kaldb/schema/SpanTagFormatterWithSchemaTest.java
@@ -1,18 +1,72 @@
 package com.slack.kaldb.schema;
 
+import static com.slack.kaldb.logstore.LuceneIndexStoreImpl.MESSAGES_FAILED_COUNTER;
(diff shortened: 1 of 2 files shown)


bryanlb (changes requested):
You know what I'm gonna ask for - we got a test for this? :D

kaldb/src/test/java/com/slack/kaldb/schema/SpanTagFormatterWithSchemaTest.java
```diff
@@ -266,4 +321,22 @@ public void testSimpleWithoutSchema() {
     assertThat(kv.getFieldType()).isEqualTo(Schema.SchemaFieldType.KEYWORD);
     assertThat(kv.getVStr()).isEqualTo("e30=");
   }
+
+  @Test
+  public void testDuplicateFieldAsTag() {
```
vthacker, line 326: this fails if I comment out the fix

bryanlb (approved):

bryanlb (approved):

PR #818, opened 2024-03-20
standardize one field type

vthacker:
###  Summary

remove the field type definition in `kaldb_search.proto` . We use that to convert the internal `FieldType` enum class and we can swap it for the new schema field type.

This will cause less confusion and ambiguity when working around the schema work. I think the change should be backwards compatible but I'll test internally and confim

### DIFF
--- a/kaldb/src/main/proto/kaldb_search.proto
+++ b/kaldb/src/main/proto/kaldb_search.proto
@@ -4,6 +4,8 @@ package slack.proto.kaldb;
 
 option java_package = "com.slack.kaldb.proto.service";
 
+import "schema.proto";
+
 message SearchRequest {
   // Data sets or chunk_ids to be searched
   string dataset = 1;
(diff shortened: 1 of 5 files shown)


kaldb/src/main/proto/kaldb_search.proto
```diff
@@ -238,21 +240,9 @@ message SchemaResult {
...
-  FLOAT = 4;
-  BOOLEAN = 5;
-  DOUBLE = 6;
-  ID = 7;
-}
-
 message SchemaDefinition {
   // Field type as defined by KalDB
-  FieldType type = 1;
+  slack.proto.kaldb.schema.SchemaFieldType type = 1;
```
bryanlb, line 245: This wouldn't be backward compatible, because the enum values have changed. For example ID used to be "7" in the previous option, but would now be "14".

vthacker: It wouldn't. But we don't use it for inter node communiction. This is just a proto and we always translate the JSON back to the HTTP API. So we don't technically need to be back compat

bryanlb (approved):

Issue #798, opened 2024-03-12
[FEATURE] Support Opensearch API


autata:
### Description
Opening this issue to track support for the Opensearch API. This is to replace the draft PR: https://github.com/slackhq/astra/pull/780

I'm new to this project, so I can't outline each of the places support is missing, but it would be valuable to support the spec. In particular, ([_msearch](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-multi-search.html) and [range query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html)) in their entirety.

For example if either `lte` or `gte` are missing, astra return a 500 (Internal Server Error) response due to null pointer exception. In Opensearch these are not required. Also a 500 is not ideal. Perhaps 400 (Bad Request) and a helpful error string would be better if not implementing the API.

For now, I've closed the draft PR that shows this issue. Because it looks like Opensearch will be implemented in a few months, it probably doesn't make sense to do one-off changes to the return codes.

### Requirements (place an `x` in each of the `[ ]`)
* [x] I've read and understood the [Contributing guidelines](../CONTRIBUTING.md) and have done my best effort to follow them.
* [x] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

PR #819, opened 2024-03-21
Improve ES query compatibility for Grafana 10

bryanlb:
###  Summary

Improves ES query support for Grafana 10. This is to enable easier datasource migrations from ES to Astra.

This improves the current state of #798, but a greater rework of query parsing is still needed.

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
+++ b/kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
@@ -61,6 +61,27 @@ public class ElasticsearchApiService {
     this.searcher = searcher;
   }
 
+  /** Returns metadata about the cluster */
+  @Get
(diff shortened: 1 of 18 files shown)


kaldb/src/main/java/com/slack/kaldb/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -61,6 +61,27 @@ public ElasticsearchApiService(KaldbQueryServiceBase searcher) {
     this.searcher = searcher;
   }
 
+  /** Returns metadata about the cluster */
+  @Get
+  @Path("/")
+  public HttpResponse clusterMetadata() {
```
bryanlb, line 67: Due to https://github.com/grafana/grafana/commit/4d7c0904ef3455a2fc8db0ba813463c222d6c350

vthacker (approved):

PR #829, opened 2024-03-26
deal with muliple tags with the same key

vthacker:
###  Summary

deal with muliple tags with the same key

Without the fix, if there are multiple tags with the same key we throw this error
```
java.lang.IllegalStateException: Duplicate key tag1 (attempted merging values key: "tag1"
v_str: "value1"
 and key: "tag1"
v_str: "value1"
)

	at java.base/java.util.stream.Collectors.duplicateKeyException(Collectors.java:135)
	at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:182)
...

```

### DIFF
--- a/kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/opensearch/BulkApiRequestParser.java
+++ b/kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/opensearch/BulkApiRequestParser.java
@@ -112,14 +112,22 @@ public class BulkApiRequestParser {
     sourceAndMetadata.remove(IngestDocument.Metadata.ID.getFieldName());
     sourceAndMetadata.remove(IngestDocument.Metadata.INDEX.getFieldName());
 
-    sourceAndMetadata.forEach(
-        (key, value) -> spanBuilder.addTags(SpanFormatter.convertKVtoProto(key, value, schema)));
-    spanBuilder.addTags(
-        Trace.KeyValue.newBuilder()
-            .setKey(SERVICE_NAME_KEY)
-            .setVType(Trace.ValueType.STRING)
-            .setVStr(index)
-            .build());
+    boolean tagsContainServiceName = false;
+    for (Map.Entry<String, Object> kv : sourceAndMetadata.entrySet()) {
+      if (!tagsContainServiceName && kv.getKey().equals(SERVICE_NAME_KEY)) {
+        tagsContainServiceName = true;
+      }
+      spanBuilder.addTags(SpanFormatter.convertKVtoProto(kv.getKey(), kv.getValue(), schema));
+    }
+    if (!tagsContainServiceName) {
+      spanBuilder.addTags(
+          Trace.KeyValue.newBuilder()
+              .setKey(SERVICE_NAME_KEY)
+              .setVType(Trace.ValueType.STRING)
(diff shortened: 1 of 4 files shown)


kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/opensearch/BulkApiRequestParser.java
```diff
@@ -112,14 +112,22 @@ public static Trace.Span fromIngestDocument(
...
-            .setVStr(index)
-            .build());
+    boolean tagsContainServiceName = false;
+    for (Map.Entry<String, Object> kv : sourceAndMetadata.entrySet()) {
+      if (!tagsContainServiceName && kv.getKey().equals(SERVICE_NAME_KEY)) {
+        tagsContainServiceName = true;
+      }
+      spanBuilder.addTags(SpanFormatter.convertKVtoProto(kv.getKey(), kv.getValue(), schema));
+    }
+    if (!tagsContainServiceName) {
```
vthacker, line 122: making sure we don't add `service_name` twice as a tag. We will anyways pick the first entry in the indexer, but this is for lesser ambiguity in the future.

bryanlb: A more readable method of doing this might be:

```java
    sourceAndMetadata.forEach((key, value) -> spanBuilder.addTags(SpanFormatter.convertKVtoProto(key, value, schema)));
    if (sourceAndMetadata.entrySet().stream().noneMatch((entry) -> entry.getKey().equals(SERVICE_NAME_KEY))) {
      spanBuilder.addTags(
          Trace.KeyValue.newBuilder()
              .setKey(SERVICE_NAME_KEY)
              .setVType(Trace.ValueType.STRING)
              .setFieldType(Schema.SchemaFieldType.KEYWORD)
              .setVStr(index)
              .build());
    }
```

kaldb/src/main/java/com/slack/kaldb/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
```diff
@@ -430,7 +430,7 @@ public Document fromMessage(Trace.Span message) throws JsonProcessingException {
     Map<String, Trace.KeyValue> tags =
         message.getTagsList().stream()
             .map(keyValue -> Map.entry(keyValue.getKey(), keyValue))
-            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (k1, _) -> k1));
```
bryanlb, line 433: Perhaps a better variable name than `k1`?

kaldb/src/main/java/com/slack/kaldb/bulkIngestApi/opensearch/BulkApiRequestParser.java
bryanlb: Note that in the above example it would require traversing the tags up to twice, which may not be desired since this is a hot path.

bryanlb (approved):

bryanlb (approved):

PR #838, opened 2024-04-01
adjust the grafana config with the correct TS field and the test index used in the running instructions

autata:
###  Summary

These instructions https://github.com/slackhq/astra/discussions/799 make the database name `test`. From querying the data, it looks like the supported timestamp field is `_timesinceepoch`

Note `@timestamp` is only filled in if that was added in the request. but `_timesinceepoch` is always filled in.
image

Explore works by default with this configuration:
image

### Requirements


bryanlb (approved):

bryanlb:
@allcontributors please add @autata for code

PR #839, opened 2024-04-01
support multi-fields

vthacker:
###  Summary

We want the ability to add multi-fields. The way it works is for each field we can define a list of fields. For those multi fields the value will be the same but the type will be different


docs/topics/Schema.md
```diff
@@ -0,0 +1,79 @@
+# Schema
```
bryanlb, line 1: 😍

PR #849, opened 2024-04-08
Update docker compose to work with astra grafana plugin

autata:
###  Summary
Updates docker compose to allow the astra grafana plugin to work. (note, the [slack-astra-app](https://github.com/slackhq/slack-astra-app) repo must be cloned in the same directory where the astra project is cloned. Also the build instructions need to be followed as well.

TODO: fix other flags. current `HEAD` renamed some flags since this update. 

TODO: provision the datasource. Configuration that works:
image

### Requirements

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,13 +28,13 @@ services:
       - zookeeper
 
   grafana:
-    image: 'grafana/grafana:8.0.3'
+    image: 'grafana/grafana:10.3.1'
     container_name: dep_grafana
     ports:
       - '3000:3000'
     volumes:
       - ./config/grafana/provisioning:/etc/grafana/provisioning
-      - ../slack-kaldb-app:/var/lib/grafana/plugins
+      - ../slack-astra-app:/var/lib/grafana/plugins
     environment:
       GF_LOG_MODE: "console file"
       GF_LOG_LEVEL: "info"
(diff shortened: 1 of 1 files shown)


bryanlb:
Hey @autata, looks like this may have been created against a branch prior to the rename. I see that the rename was already done, but we are currently missing the required Grafana version bump you have here.

We may be able to get the datasource configured automatically in the https://github.com/slackhq/astra/blob/master/config/grafana/provisioning/datasources/automatic.yml, but I'm not sure how that works when the plugin isn't enabled by default in the docker compose. 🤔

autata:
@bryanlb yeah this was created before the rename (why it is in draft still). I need to rebase and fix the other flags still so docker compose keeps working. Ah yeah and I need to enable it. I did manually, but should be specified in docker compose for sure.

Another question: do you have a timeline when the plugin will be available via [grafana installation (signed?)](https://grafana.com/grafana/plugins/) vs sideloading/unsigned it (which I'm doing here)?

bryanlb:
@autata we do not currently have a timeline on signing the plugin. This requires some administration stuff on our end to setup and we would need to go through the administration with Grafana to get a signing key. This is something we'd like to do, but realistically isn't something we expect to do soon.

We have pivoted internally to focusing more on supporting the drop-in replacement use case, and with the latest builds of Astra you should be able to connect an ES datasource in Grafana to Astra and have it work with most aggregations.

PR #909, opened 2024-05-01
Add true date field support and also index "@timestamp" as a date

vthacker:
###  Summary

PR does two things.

1. Add true date field support  - which means any schema defined date field will get stored as a date and the indexer will use it appropriately. i.e you can do date range queries but the values will show up a a string an not a long since epoch like it was previously.
2. Index "@timestamp" as a date field if the document contains it. For now just treat this as an additional field and then later solidify our plan our reserved fields.

### DIFF
--- a/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
+++ b/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
@@ -9,10 +9,6 @@ import com.slack.astra.writer.SpanFormatter;
 import com.slack.service.murron.trace.Trace;
 import java.io.IOException;
 import java.time.Instant;
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -55,15 +51,12 @@ public class BulkApiRequestParser {
     try {
       if (sourceAndMetadata.containsKey(ReservedFields.TIMESTAMP)) {
         String dateString = (String) sourceAndMetadata.get(ReservedFields.TIMESTAMP);
-        LocalDateTime localDateTime =
-            LocalDateTime.parse(dateString, DateTimeFormatter.ISO_DATE_TIME);
-        Instant instant = localDateTime.toInstant(ZoneOffset.UTC);
+        Instant instant = Instant.parse(dateString);
         return instant.toEpochMilli();
       }
 
       // assumption that the provided timestamp is in millis
-      // at some point both th unit and field need to be configurable
-      // when we do that, remember to change the called to appropriately remove the field
+      // at some point both the unit and field need to be configurable
       if (sourceAndMetadata.containsKey("timestamp")) {
         return (long) sourceAndMetadata.get("timestamp");
       }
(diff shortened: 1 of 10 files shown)


astra/src/main/java/com/slack/astra/metadata/schema/FieldType.java
```diff
@@ -133,7 +136,22 @@ public Schema.SchemaFieldType toSchemaFieldType() {
...
+        }
+        if (fieldDef.isStored) {
+          doc.add(new StoredField(name, localDateTime.toString()));
+        }
+        if (fieldDef.storeDocValue) {
+          doc.add(new NumericDocValuesField(name, timeSinceEpoch));
+        }
+      } else {
+        // back-compat
+        LONG.addField(doc, name, value, fieldDef);
```
vthacker, line 153: so when we deploy the indexer first, docs will keep getting added as this. Once we deploy preprocessor the object will become LocalDateTime

astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
```diff
@@ -104,6 +105,29 @@ public static Trace.Span fromIngestDocument(
...
         TimeUnit.MICROSECONDS.convert(timestampInMillis, TimeUnit.MILLISECONDS));
 
+    // Today @timestamp is parsed in getTimestampFromIngestDocument and set as timestampInMillis
+    // However we store it as _timesinceepoch in the trace.proto
+    // Let's also store the original field name and classify it as a date
+    if (ingestDocument.hasField("@timestamp")) {
+      String dateString = ingestDocument.getFieldValue("@timestamp", String.class);
+      LocalDateTime localDateTime =
+          LocalDateTime.parse(dateString, DateTimeFormatter.ISO_DATE_TIME);
+      // As I understand frontend UIs expect this to be constant at UTC and they adjust it based on
```
bryanlb, line 115: Instant can always be considered UTC (ie, https://www.baeldung.com/java-instant-vs-localdatetime#:~:text=The%20easiest%20way%20to%20think,in%20the%20UTC%20time%20zone). You can see this also in the javadocs for `Instant.atOffset()`.

As such we'd always want to store this at UTC, and you are correct that timezone conversions are done at query time - https://github.com/slackhq/astra/blob/master/astra/src/main/java/com/slack/astra/logstore/opensearch/OpenSearchAdapter.java#L985

bryanlb (changes requested):
Consider also adding a test for a timestamp that comes in with a timezone other than UTC, and ensure it's converted correctly. Just to make sure this doesn't regress.

vthacker:
> Consider also adding a test for a timestamp that comes in with a timezone other than UTC, and ensure it's converted correctly. Just to make sure this doesn't regress.

I believe that this will break today. We use the ISO_INSTANT formatter which only accepts time in UTC. I believe a good way to improve this in the future is to paraterize this and let the user pass their data format.

astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
```diff
@@ -49,11 +50,11 @@ public static Map<String, List<Trace.Span>> parseRequest(
    * in millis 2. Check if a field called `@timestamp` exists and parse that as a date (since
    * logstash sets that) 3. Use the current time from the ingestMetadata
    */
-  public static long getTimestampFromIngestDocument(IngestDocument ingestDocument) {
+  public static long getTimestampFromIngestDocument(Map<String, Object> sourceAndMetadata) {
 
     try {
-      if (ingestDocument.hasField("@timestamp")) {
```
vthacker, line 55: hasField basically gets sourceAndMetadata so the change shouldn't have any effect.

bryanlb (dismissed):

bryanlb (dismissed):

PR #910, opened 2024-05-02
remove special handling for "type" and remove some reserved fields that were not being used

vthacker:
###  Summary

We don't need to do any special handling for the type tag. This also means user definition for type is respected and not overriden.

### DIFF
--- a/astra/src/main/java/com/slack/astra/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
+++ b/astra/src/main/java/com/slack/astra/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
@@ -6,7 +6,6 @@ import static com.slack.astra.writer.SpanFormatter.DEFAULT_LOG_MESSAGE_TYPE;
 import static com.slack.astra.writer.SpanFormatter.isValidTimestamp;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.google.common.collect.ImmutableMap;
 import com.slack.astra.logstore.DocumentBuilder;
 import com.slack.astra.logstore.FieldDefMismatchException;
 import com.slack.astra.logstore.LogMessage;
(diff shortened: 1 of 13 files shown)


astra/src/main/java/com/slack/astra/logstore/schema/SchemaAwareLogDocumentBuilderImpl.java
```diff
@@ -381,7 +313,7 @@ public Document fromMessage(Trace.Span message) throws JsonProcessingException {
           doc,
           LogMessage.SystemField.ID.fieldName,
           message.getId().toStringUtf8(),
-          Schema.SchemaFieldType.KEYWORD,
+          Schema.SchemaFieldType.ID,
```
vthacker, line 316: this is important. previosuly since `_id` was predefined, we had it wrongly set to KEYWORD which meant all tests were failing.

bryanlb (dismissed):

astra/src/test/java/com/slack/astra/chunk/IndexingChunkImplTest.java
```diff
@@ -499,16 +499,7 @@ public void tearDown() throws IOException, TimeoutException {
...
-          Trace.Span.newBuilder()
-              .setId(ByteString.copyFromUtf8("1"))
-              .addTags(
-                  Trace.KeyValue.newBuilder()
-                      .setVInt32(123)
-                      .setKey(LogMessage.ReservedField.MESSAGE.fieldName)
-                      .setFieldType(Schema.SchemaFieldType.INTEGER)
-                      .build())
-              .build();
+      Trace.Span invalidSpan = Trace.Span.newBuilder().build();
```
vthacker, line 502: old way of creating invalid span doesn't work. Earlier we took advantage of the fact `message` was predefined as a string but we said in the span it's Integer.

So throw error

New way is just to create empty span - we fail without ID.

astra/src/test/java/com/slack/astra/elasticsearchApi/ElasticsearchApiServiceTest.java
```diff
@@ -137,7 +137,7 @@ public void testSchemaFields() throws Exception {
         objectMapper.convertValue(
             jsonNode.get("test").get("mappings").get("properties"), Map.class);
     assertThat(map).isNotNull();
-    assertThat(map.size()).isEqualTo(31);
+    assertThat(map.size()).isEqualTo(25);
```
vthacker, line 140: These are the 6 pre-defined fields that won't show up in the mapping API.
hostname
duration_ms
package
payload
tag
type


Previously they would have shown up even if no one had ever added a record for it. So this is a good change

astra/src/test/java/com/slack/astra/server/AstraTest.java
```diff
@@ -275,14 +274,6 @@ public void testDistributedQueryOneIndexerOneQueryNode() throws Exception {
             indexerMeterRegistry);
     indexer.serviceManager.awaitHealthy(DEFAULT_START_STOP_DURATION);
 
-    await()
```
vthacker, line 278: we wait for the service to be healthy. That should be enough?

We don't pre-register  any fields so this counter is 0

astra/src/test/java/com/slack/astra/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -1354,7 +1354,7 @@ public void testDisabledFullTextSearch() {
                         "1", LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName, "1s"))
                 .hits
                 .size())
-        .isEqualTo(3);
+        .isEqualTo(2);
```
vthacker, line 1357: this should always have been 2. The 3rd is matching on a `message` field which is KEYWORD. But earlier it would match it since `message` was a predefined field TEXT

astra/src/test/java/com/slack/astra/logstore/search/LogIndexSearcherImplTest.java
```diff
@@ -1339,7 +1339,7 @@ public void testDisabledFullTextSearch() {
                         "1", LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName, "1s"))
                 .hits
                 .size())
-        .isEqualTo(2);
+        .isEqualTo(0);
```
vthacker, line 1342: same with this

bryanlb (approved):

PR #921, opened 2024-05-07
disk based chunk rollover should also have a time based limit

vthacker:
###  Summary

This is especially useful in dev where the log volume is low to have a time based rollover on top of disk size.

We see scenarios where a chunk is several hours long and during an indexer deploy messages are missing till the recovery task completes.

In a separate task we should try to write out the current chunk during an indexer deploy

### DIFF
--- a/astra/src/main/java/com/slack/astra/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
+++ b/astra/src/main/java/com/slack/astra/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
@@ -7,9 +7,12 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.slack.astra.proto.config.AstraConfigs;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.lucene.index.IndexNotFoundException;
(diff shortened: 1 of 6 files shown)


config/config.yaml
```diff
@@ -3,6 +3,7 @@ nodeRoles: [${NODE_ROLES:-QUERY,INDEX,CACHE,MANAGER,RECOVERY,PREPROCESSOR}]
 indexerConfig:
   maxMessagesPerChunk: ${INDEXER_MAX_MESSAGES_PER_CHUNK:-100000}
   maxBytesPerChunk: ${INDEXER_MAX_BYTES_PER_CHUNK:-1000000}
+  maxTimePerChunkMins: ${INDEXER_MAX_TIME_PER_CHUNK_MINS:-30}
```
bryanlb, line 6: This should probably default to 60.

astra/src/main/java/com/slack/astra/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
```diff
@@ -44,20 +47,36 @@ public class DiskOrMessageCountBasedRolloverStrategy implements ChunkRollOverStr
 
   private final AtomicReference<FSDirectory> activeChunkDirectory = new AtomicReference<>();
   private final AtomicLong approximateDirectoryBytes = new AtomicLong(0);
+  private final AtomicBoolean maxTimePerChunksMinsReached = new AtomicBoolean(false);
+  private final Instant chunkStartTime;
 
   public static DiskOrMessageCountBasedRolloverStrategy fromConfig(
```
bryanlb, line 53: We should probably (eventually?) rename this. Also I don't think the max messages rollover is helpful at all for the indexer today, so that probably could also (eventually) be removed.

bryanlb (changes requested):
Update the default value to something like 60 or 90 and this is good to go.

config/config.yaml
vthacker: done!

astra/src/main/java/com/slack/astra/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
vthacker: makes sense. We'll do it in a followup PR

bryanlb (approved):

astra/src/main/java/com/slack/astra/chunkrollover/DiskOrMessageCountBasedRolloverStrategy.java
```diff
@@ -44,28 +47,58 @@ public class DiskOrMessageCountBasedRolloverStrategy implements ChunkRollOverStr
...
+            long dirSize = calculateDirectorySize(activeChunkDirectory);
+            // in case the method fails to calculate we return -1 so don't update the old value
+            if (dirSize > 0) {
+              approximateDirectoryBytes.set(dirSize);
+            }
+            if (!maxTimePerChunksMinsReached.get()
+                && Instant.now()
+                    .isAfter(rolloverStartTime.plus(maxTimePerChunksSeconds, ChronoUnit.SECONDS))) {
+              LOG.info(
+                  "`Max time p`er chunk reached. chunkStartTime: {} currentTime: {}",
```
bryanlb, line 95: This looks like a typo potentially

bryanlb (changes requested):

bryanlb (approved):

PR #923, opened 2024-05-09
Rollover current chunk during shutdown

vthacker:
###  Summary

Every time we do an an indexer deploy, data goes missing for ~10 mins till the recovery task completes. Our users notice this and we want to address this.

Our k8's timeout for  a pod graceful shutdow can be set to 180s max internally, after which the pod will be killed forcibly.

Astra now uses S3 native library (CRT client) which makes uploading/downloading chunks from S3 at close to the theoretical max of the underlying host.

Quick math:

We configure a chunk rollover size is 15GB. Let's assume the worst case we are almost at the limit when the shutdown hook is called.

We run our indexers on `r5d.24xlarge` nodes which have a 25 Gbps Network Bandwidth. Let's assume we have 25 indexer pods on this host that need to upload their chunks before shutdown.

meaning each pod gets roughly 1Gbps on average. 15GB @ 1Gbps comes to exactly 2 mins.

Worst case scenario of 15GB chunk we may or may not make it. But if you take an average chunk size of 10GB when the shutdown hook is called we have a very good chance of succeeding.


astra/src/main/java/com/slack/astra/server/Astra.java
```diff
@@ -446,7 +446,7 @@ public void failure(Service service) {
   void shutdown() {
     LOG.info("Running shutdown hook.");
     try {
-      serviceManager.stopAsync().awaitStopped(30, TimeUnit.SECONDS);
+      serviceManager.stopAsync().awaitStopped(180, TimeUnit.SECONDS);
```
vthacker, line 449: this has to be higher than any service shutdown timeouts.

mansu: It typically takes a few minutes to roll over a chunk. We couldn't do this before since the k8s timeout for container shutdown was on the order of ~30s which was not enough for roll over. Is that not an issue anymore? Or is this only applicable to low volume use cases?

vthacker: That's a very good question! So the k8's timeout has been 120s internally for us. After which the pod will be killed forcibly.  

We're revising this because every time we do an an indexer deploy, data goes missing for ~10 mins till the recovery task completes. Our users notice this and have raised escals hence we are revisiting this.

Bryan did some really cool work a few months ago to use the S3 native library (CRT client) which makes Kaldb upload/download chunks from S3 at close to the host's network bandwidth.

Our chunk rollover size is 15GB. In the worst case we are almost at the limit when the shutdown hook is called. 

We run our indexers on `r5d.24xlarge` nodes which have a `25 Gbps` Network Bandwidth. Let's assume we have 25 indexers on this host that need to upload their chunks before shutdown.

So each pod gets roughly 1Gbps on average. 15GB @ 1Gbps comes to exactly 2 mins. 

So worst case scenario of 15GB chunk we may or may not make it. But if you take an average chunk size of 10GB when the shutdown hook is called we have a very good chance.

I'll be testing it this week and seeing what happens

vthacker:
Currently the effort is blocked by https://github.com/aws/aws-sdk-java-v2/issues/3963

mansu:
In the past this change was not possible since the k8s timeout couldn't be longer than 30secs after a shutdown was issued. Are longer timeouts on k8s not longer an issue?

Alternatively, a better solution is to incrementally upload lucene segments to S3? That would solve 2 problems for us:
* On a shutdown we would need to upload newly created chunks and not all the chunks. This makes the shutdowns and deployments fast.
* This incremental chunk upload functionality can be used for features like making the indexer highly available in the future.

PR #924, opened 2024-05-10
Add logging at main API endpoints (bulk ingest, msearch) when 500 is returned

autata:
###  Summary
Add 2 more places to log errors where astra returns a 500. Without this, nothing is logged in these cases when astra encounters an internal server error. I picked error level because other places where a 500 is returned, those are also logged as errors. These may not be as severe and more implementation choices. I'm curious what level the reviewers would prefer here.

In the case of `request must contain only 1 unique index`, debug may be a more suitable log level as there could be a lot of these and error logs could be spammy.

For multisearch query, an example where this is triggered would be shown in the PR: https://github.com/slackhq/astra/pull/780. (Null pointer exception due to json deserialization logic) Potentially, this could be better suited as a warning log?

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/bulkIngestApi/BulkIngestApi.java
+++ b/astra/src/main/java/com/slack/astra/bulkIngestApi/BulkIngestApi.java
@@ -81,6 +81,7 @@ public class BulkIngestApi {
         BulkIngestResponse response =
             new BulkIngestResponse(0, 0, "request must contain only 1 unique index");
         future.complete(HttpResponse.ofJson(INTERNAL_SERVER_ERROR, response));
+        LOG.error("request must contain only 1 unique index");
         return HttpResponse.of(future);
       }
 
--- a/astra/src/main/java/com/slack/astra/elasticsearchApi/ElasticsearchApiService.java
+++ b/astra/src/main/java/com/slack/astra/elasticsearchApi/ElasticsearchApiService.java
@@ -100,18 +100,18 @@ public class ElasticsearchApiService {
       CurrentTraceContext currentTraceContext = Tracing.current().currentTraceContext();
       try (var scope = new StructuredTaskScope<EsSearchResponse>()) {
         List<StructuredTaskScope.Subtask<EsSearchResponse>> requestSubtasks =
-          openSearchRequest.parseHttpPostBody(postBody).stream()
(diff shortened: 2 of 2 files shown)


astra/src/main/java/com/slack/astra/bulkIngestApi/BulkIngestApi.java
```diff
@@ -81,6 +81,7 @@ public HttpResponse addDocument(String bulkRequest) {
         BulkIngestResponse response =
             new BulkIngestResponse(0, 0, "request must contain only 1 unique index");
         future.complete(HttpResponse.ofJson(INTERNAL_SERVER_ERROR, response));
+        LOG.error("request must contain only 1 unique index");
```
bryanlb, line 84: Ah, this is a good add.

astra/src/main/java/com/slack/astra/elasticsearchApi/ElasticsearchApiService.java
```diff
@@ -92,24 +92,30 @@ public HttpResponse clusterMetadata() {
...
-              Map.of("traceId", getTraceId()));
-      return HttpResponse.of(
+            0,
+            requestSubtasks.stream().map(StructuredTaskScope.Subtask::get).toList(),
+            Map.of("traceId", getTraceId()));
+        return HttpResponse.of(
           HttpStatus.OK, MediaType.JSON_UTF_8, JsonUtil.writeAsString(responseMetadata));
+      }
+    } catch (Exception e) {
+      LOG.error("Error fulfilling request for multisearch query", e);
```
bryanlb, line 117: I believe we should already be logging these, as we've enabled Armeria request logging here: https://github.com/slackhq/astra/blob/2d7c87f2e7d044474d2c3e051277bd485b5c66b6/astra/src/main/java/com/slack/astra/server/ArmeriaService.java#L72-L84

This should automatically log all successful and failed requests at the log levels indicated. This does make use of `responseContentSanitizer` which has since been deprecated in Armeria.

I'm not sure if this currently logs out the stack trace, but instead of adding try/catches to the individual endpoints I think that looking into the `LoggingService` decorator would be a better approach.

bryanlb (changes requested):

PR #940, opened 2024-05-30
Add sizeInBytes field

kx-chen:
###  Summary
Add sizeInBytes field to `SnapshotMetadata` and `ChunkInfo`. Used to keep track of size of chunks in preparation for dynamic chunk sizes.

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/chunk/ChunkInfo.java
+++ b/astra/src/main/java/com/slack/astra/chunk/ChunkInfo.java
@@ -42,7 +42,8 @@ public class ChunkInfo {
         chunkInfo.getDataEndTimeEpochMs(),
         chunkInfo.maxOffset,
         chunkInfo.kafkaPartitionId,
-        Metadata.IndexType.LOGS_LUCENE9);
+        Metadata.IndexType.LOGS_LUCENE9,
+        chunkInfo.sizeInBytes);
   }
 
   /* A unique identifier for a the chunk. */
@@ -77,6 +78,9 @@ public class ChunkInfo {
 
   // Path to S3 snapshot.
   private String snapshotPath;
+
+  // Size of chunk in bytes
+  private long sizeInBytes;
 
   public ChunkInfo(
       String chunkId, long chunkCreationTimeEpochMs, String kafkaPartitionId, String snapshotPath) {
@@ -164,6 +168,14 @@ public class ChunkInfo {
     return snapshotPath;
   }
 
+  public long getSizeInBytes() {
+    return sizeInBytes;
+  }
+
+  public void setSizeInBytes(long sizeInBytes) {
+    this.sizeInBytes = sizeInBytes;
+  }
+
   public void updateMaxOffset(long newOffset) {
     maxOffset = Math.max(maxOffset, newOffset);
   }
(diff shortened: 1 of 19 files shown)


astra/src/main/java/com/slack/astra/chunk/ChunkInfo.java
```diff
@@ -78,6 +79,9 @@ public static SnapshotMetadata toSnapshotMetadata(ChunkInfo chunkInfo, String ch
   // Path to S3 snapshot.
   private String snapshotPath;
 
+  // Size of chunk in bytes
+  private long sizeInBytes;
```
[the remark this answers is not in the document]
bryanlb: Since this isn't a required param in the constructor, set a sensible default value (ie, `null`, `0`, `-1`) so that it's apparent if someone doesn't set it and then tries to read it what they should expect.

astra/src/main/java/com/slack/astra/chunk/ReadWriteChunk.java
```diff
@@ -89,6 +89,7 @@ public abstract class ReadWriteChunk<T> implements Chunk<T> {
 
   // TODO: Move this flag into LogStore?.
   private boolean readOnly;
+  protected long size;
```
bryanlb, line 92: Is this unused?

astra/src/main/java/com/slack/astra/chunk/ReadWriteChunk.java
```diff
@@ -243,14 +245,17 @@ public boolean snapshotToS3(String bucket, String prefix, BlobFs blobFs) {
       // Upload files
       logger.info("{} active files in {} in index", filesToUpload.size(), dirPath);
       for (String fileName : filesToUpload) {
-        logger.debug("File name is {}}", fileName);
+        File fileToCopy = new File(dirPath.toString(), fileName);
```
bryanlb, line 248: You may want to use something like `FileUtils.size()` here, or `Files.size()` so that you don't have to create a new file object just to get the bytes.

astra/src/test/java/com/slack/astra/metadata/snapshot/SnapshotMetadataSerializerTest.java
```diff
@@ -20,7 +20,8 @@ public void testSnapshotMetadataSerializer() throws InvalidProtocolBufferExcepti
     final String partitionId = "1";
 
     SnapshotMetadata snapshotMetadata =
-        new SnapshotMetadata(name, path, startTime, endTime, maxOffset, partitionId, LOGS_LUCENE9);
+        new SnapshotMetadata(
```
bryanlb, line 23: Consider adding a test for what happens when you try to deserialize a snapshot that doesn't have size on populated. You can construct a new `Metadata.SnapshotMetadata` (protobuf) object using the builder, and leave the size off then try to deserialize that into a `SnapshotMetadata`.

The existing ZK snapshots won't have size on them, so we want to make sure we can still construct these SnapshotMetdata objects for the time it'll take for that size to populate to all snapshots.

bryanlb (changes requested):

bryanlb (approved):

PR #951, opened 2024-06-12
special handling for duration_ms within the zipkin api

vthacker:


bryanlb (approved):
A test would be neat too 😂

PR #958, opened 2024-06-20
Cache Node Assignment Service

kx-chen:
###  Summary
This PR adds a basic draft of the cache node assignment service + tests. Several other changes were added in order to support the service, they include:
* Adding `CacheNodeAssignment` and `CacheNodeMetadata` classes (plus serializers, ZK store classes, proto file changes)
* Having the `CachingChunkManager` listen for changes in assignments and download data as appropriate

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/chunk/ReadOnlyChunkImpl.java
+++ b/astra/src/main/java/com/slack/astra/chunk/ReadOnlyChunkImpl.java
@@ -8,6 +8,8 @@ import com.slack.astra.logstore.search.LogIndexSearcher;
 import com.slack.astra.logstore.search.LogIndexSearcherImpl;
 import com.slack.astra.logstore.search.SearchQuery;
 import com.slack.astra.logstore.search.SearchResult;
+import com.slack.astra.metadata.cache.CacheNodeAssignment;
+import com.slack.astra.metadata.cache.CacheNodeAssignmentStore;
 import com.slack.astra.metadata.cache.CacheSlotMetadata;
 import com.slack.astra.metadata.cache.CacheSlotMetadataStore;
 import com.slack.astra.metadata.core.AstraMetadataStoreChangeListener;
@@ -56,6 +58,7 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
   private Path dataDirectory;
   private ChunkSchema chunkSchema;
   private Metadata.CacheSlotMetadata.CacheSlotState cacheSlotLastKnownState;
+  private Metadata.CacheNodeAssignment.CacheNodeAssignmentState lastKnownAssignmentState;
 
   private final String dataDirectoryPrefix;
   private final String s3Bucket;
@@ -65,8 +68,11 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
   private final ReplicaMetadataStore replicaMetadataStore;
   private final SnapshotMetadataStore snapshotMetadataStore;
   private final SearchMetadataStore searchMetadataStore;
+  private CacheNodeAssignmentStore cacheNodeAssignmentStore;
   private final MeterRegistry meterRegistry;
   private final BlobFs blobFs;
+  private String assignmentId;
+  private String cacheNodeId;
 
   public static final String CHUNK_ASSIGNMENT_TIMER = "chunk_assignment_timer";
   public static final String CHUNK_EVICTION_TIMER = "chunk_eviction_timer";
@@ -80,6 +86,42 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
       this::cacheNodeListener;
 
   private final ReentrantLock chunkAssignmentLock = new ReentrantLock();
+
+  public ReadOnlyChunkImpl(
+      AsyncCuratorFramework curatorFramework,
+      MeterRegistry meterRegistry,
+      BlobFs blobFs,
+      SearchContext searchContext,
+      String s3Bucket,
+      String dataDirectoryPrefix,
+      String replicaSet,
+      CacheSlotMetadataStore cacheSlotMetadataStore,
+      ReplicaMetadataStore replicaMetadataStore,
+      SnapshotMetadataStore snapshotMetadataStore,
+      SearchMetadataStore searchMetadataStore,
+      CacheNodeAssignmentStore cacheNodeAssignmentStore,
+      String assignmentId,
+      String cacheNodeId)
+      throws Exception {
+    this(
+        curatorFramework,
+        meterRegistry,
+        blobFs,
+        searchContext,
+        s3Bucket,
+        dataDirectoryPrefix,
+        replicaSet,
+        cacheSlotMetadataStore,
+        replicaMetadataStore,
+        snapshotMetadataStore,
+        searchMetadataStore);
+    this.assignmentId = assignmentId;
+    this.cacheNodeId = cacheNodeId;
+    this.cacheNodeAssignmentStore = cacheNodeAssignmentStore;
+    this.lastKnownAssignmentState = Metadata.CacheNodeAssignment.CacheNodeAssignmentState.LOADING;
+
+    downloadChunkData();
+  }
 
   public ReadOnlyChunkImpl(
       AsyncCuratorFramework curatorFramework,
@@ -106,17 +148,19 @@ public class ReadOnlyChunkImpl<T> implements Chunk<T> {
     this.snapshotMetadataStore = snapshotMetadataStore;
     this.searchMetadataStore = searchMetadataStore;
 
-    CacheSlotMetadata cacheSlotMetadata =
-        new CacheSlotMetadata(
-            slotId,
-            Metadata.CacheSlotMetadata.CacheSlotState.FREE,
-            "",
-            Instant.now().toEpochMilli(),
-            List.of(Metadata.IndexType.LOGS_LUCENE9),
-            searchContext.hostname,
-            replicaSet);
-    cacheSlotMetadataStore.createSync(cacheSlotMetadata);
-    cacheSlotMetadataStore.addListener(cacheSlotListener);
+    if (!Boolean.getBoolean("astra.ng.dynamicChunkSizes")) {
+      CacheSlotMetadata cacheSlotMetadata =
+          new CacheSlotMetadata(
+              slotId,
+              Metadata.CacheSlotMetadata.CacheSlotState.FREE,
+              "",
+              Instant.now().toEpochMilli(),
+              List.of(Metadata.IndexType.LOGS_LUCENE9),
+              searchContext.hostname,
+              replicaSet);
+      cacheSlotMetadataStore.createSync(cacheSlotMetadata);
+      cacheSlotMetadataStore.addListener(cacheSlotListener);
+    }
 
     cacheSlotLastKnownState = Metadata.CacheSlotMetadata.CacheSlotState.FREE;
     chunkAssignmentTimerSuccess = meterRegistry.timer(CHUNK_ASSIGNMENT_TIMER, "successful", "true");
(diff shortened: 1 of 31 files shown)


mansu:
We already have code that does cache node assignment. How is this different from current logic?

kx-chen:
Hi Suman! Great to see you. Good question!

Yes, we are adding dynamic chunks. However, the goal is not just to support more chunks per cache node but also to improve disk space utilization.

I'm sure you already know, but chunks are not always a fixed size. While we aim for uniform size, they can vary. Since slots assume each chunk is the same size, this can lead to under-utilization of disk space if chunks are smaller than the assumed size.

For example, with a 3TB disk and 15GB per slot, we get 200 slots. If a chunk is only 10GB, there’s still 5GB of unused space allocated for that slot.

The change in assignment algorithm addresses this. We plan to use a bin-packing algorithm to pack all chunks into the least number of cache nodes, optimizing disk space usage and reducing costs.

Let me know if that answers your question! And apologies for the lack of context in the PR—it's still a work in progress 😅

mansu:
Hi Kai, great to see you again! 

Thanks for the context. It would ideal to write a design doc or a discussion for this before we start coding to document the goals of this change. 

It seems that the goal of this change is improve disk utilization. This is a great goal. I am however curious why the there is so much wasted disk space unused in practice. If you are using size based roll over, at any given time only a small percentage of chunks would not be the same size as the cache slot but in steady state the chunks should be of the same size. If it is happening regularly, is it not a tuning issue? Also, even if the temporarily pack smaller chunks better(say after a deployment), would we not get back to a the same configuration as a fixed slot config once the roll over happens in an hour? I am asking since I may be missing a case where there are longer term advantages of this change. 

Also, the current implementation of static number of slots results employs a constant work paradigm for the entire cluster. This makes cluster operations mechanical, predictable and easy to automate. 

The dynamic slot model proposed here is much more dynamic and makes the system less predictable over all. This is big sacrifice from a design perspective. 

To carefully evaluate the effect of these changes, it may be better to break this change down into a few smaller changes:
* The most basic feature needed here is ability to create dynamic slots on cache nodes instead of fixed slots. 
* Once we can dynamically change number of slots, the next feature to consider is if we should allow nodes other than cache node(ex cluster manager) to create the slots. 
* There may be a lot of ways in which we can optimize the cache assignment algorithm. So, we should probably make the assignment algorithm change independent of these dynamic nodes. 

It is possible that we can get back disk utilization after the first 2 changes. It seems we are also changing the cache assignment nodes in this PR. I think that change may be better done independent of this change. 

Overall, I think it may be ideal if we approach this change in a more incremental manner.

.run/Astra_cache.run.xml
bryanlb, line 1: Not sure these two run files should be deleted.

astra/src/main/java/com/slack/astra/chunk/ReadOnlyChunkImpl.java
```diff
@@ -106,17 +149,19 @@ public ReadOnlyChunkImpl(
...
-            slotId,
-            Metadata.CacheSlotMetadata.CacheSlotState.FREE,
-            "",
-            Instant.now().toEpochMilli(),
-            List.of(Metadata.IndexType.LOGS_LUCENE9),
-            searchContext.hostname,
-            replicaSet);
-    cacheSlotMetadataStore.createSync(cacheSlotMetadata);
-    cacheSlotMetadataStore.addListener(cacheSlotListener);
+    if (!Boolean.getBoolean("astra.ng.dynamicChunkSizes")) {
```
bryanlb, line 152: Consider putting this string into a static final somewhere, and then using that throughout the code. This helps prevent any typos from slipping through unnoticed.

astra/src/test/java/com/slack/astra/chunk/ReadOnlyChunkImplTest.java
```diff
@@ -483,6 +485,105 @@ public void shouldUseOptimizedQueryStartEndTime() {
...
+                "*:*",
+                Instant.now().minus(1, ChronoUnit.MINUTES).toEpochMilli(),
+                Instant.now().toEpochMilli(),
+                500,
+                new DateHistogramAggBuilder(
+                    "1", LogMessage.SystemField.TIME_SINCE_EPOCH.fieldName, "1s"),
+                Collections.emptyList()));
+    assertThat(logMessageSearchResult.hits.size()).isEqualTo(10);
+
+    //    await()
```
bryanlb, line 560: Should this be commented out?

astra/src/test/java/com/slack/astra/chunkManager/CachingChunkManagerTest.java
```diff
@@ -160,6 +168,81 @@ public void testAddMessageIsUnsupported() throws TimeoutException {
...
+
+    await()
+        .timeout(10000, TimeUnit.MILLISECONDS)
+        .until(() -> cachingChunkManager.getChunksMap().size() == 1);
+    assertThat(cachingChunkManager.getChunksMap().size()).isEqualTo(1);
+
+    // assert state is correct
+    // assert searchMetadata is correct
+    // assert can query
+    System.setProperty("astra.ng.dynamicChunkSizes", "false");
```
bryanlb, line 186: I like the resetting at the end - consider trying to reset it back to the previous value. It would be safer to read the current value at the beginning, and then reset it to whatever that was at the end (true, false, or empty/deleted).

astra/src/test/java/com/slack/astra/metadata/cache/CacheNodeAssignmentSerializerTest.java
```diff
@@ -0,0 +1,3 @@
+package com.slack.astra.metadata.cache;
+
+public class CacheNodeAssignmentSerializerTest {}
```
bryanlb, line 3: Needs tests here

astra/src/test/java/com/slack/astra/metadata/cache/CacheNodeAssignmentStoreTest.java
```diff
@@ -0,0 +1,3 @@
+package com.slack.astra.metadata.cache;
+
+public class CacheNodeAssignmentStoreTest {}
```
bryanlb, line 3: Needs tests here

astra/src/test/java/com/slack/astra/metadata/cache/CacheNodeAssignmentTest.java
```diff
@@ -0,0 +1,3 @@
+package com.slack.astra.metadata.cache;
+
+public class CacheNodeAssignmentTest {}
```
bryanlb, line 3: Needs tests here

astra/src/test/java/com/slack/astra/metadata/cache/CacheNodeMetadataStoreTest.java
```diff
@@ -0,0 +1,3 @@
+package com.slack.astra.metadata.cache;
+
+public class CacheNodeMetadataStoreTest {}
```
bryanlb, line 3: Needs tests here

astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
```diff
@@ -0,0 +1,568 @@
...
+                    cacheNodeBin.getTotalCapacityBytes() - cacheNodeBin.getRemainingCapacityBytes())
+            .sum();
+
+    return (double) totalBytesRequiringAssignment / totalCacheNodeCapacityBytes;
+  }
+
+  /** Updates or inserts an (ephemeral) HPA metric for the cache nodes. This is NOT threadsafe. */
+  private static void persistCacheConfig(
+      HpaMetricMetadataStore hpaMetricMetadataStore, String replicaSet, Double demandFactor) {
+    String key = String.format(CACHE_HPA_METRIC_NAME, replicaSet);
```
bryanlb, line 509: Make sure that the other HPA publishing service isn't running at the same time, otherwise they'll end up fighting each other.

astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
```diff
@@ -0,0 +1,568 @@
...
+   * @param replicaSet the replica set for which to persist assignments
+   * @return the number of successful assignments
+   */
+  private static int createAssignments(
+      CacheNodeAssignmentStore cacheNodeAssignmentStore,
+      Map<String, CacheNodeBin> newAssignments,
+      String replicaSet,
+      Map<String, ReplicaMetadata> replicasBySnapshotId) {
+    int numAssigned = 0;
+    for (Map.Entry<String, CacheNodeBin> entry : newAssignments.entrySet()) {
```
bryanlb, line 322: Something you should keep in mind is HashMap `entrySet` isn't guaranteed to have any order. If order is important for any of the assignment logic you should look into using a different map implementation, or a list.

astra/src/main/java/com/slack/astra/chunkManager/CachingChunkManager.java
```diff
@@ -126,4 +180,76 @@ public void addMessage(Trace.Span message, long msgSize, String kafkaPartitionId
...
+      try {
+        if (chunksMap.containsKey(assignment.assignmentId)) {
+          ReadOnlyChunkImpl<T> chunk = chunksMap.get(assignment.assignmentId);
+
+          if (chunkStateChangedToEvict(assignment, chunk)) {
+            LOG.info(
+                "Starting eviction for assignment {} from node {}",
+                assignment.assignmentId,
+                cacheNodeId);
+            Thread.ofVirtual().start(() -> chunk.evictChunk(assignment));
```
bryanlb, line 201: Doing this async may end up causing some issues - because you've only send the evictChunk to be executed, but it hasn't yet been completed. Note the `onAssignmentHandler` already runs async as part of the zookeeper event handler.

astra/src/main/java/com/slack/astra/chunk/ReadOnlyChunkImpl.java
```diff
@@ -128,6 +173,142 @@ public ReadOnlyChunkImpl(
     LOG.debug("Created a new read only chunk - zkSlotId: {}", slotId);
   }
 
+  /*
+  ======================================================
+  All methods below RELATED to astra.ng.dynamicChunkSizes
+  ======================================================
+   */
+
+  public void evictChunk(CacheNodeAssignment cacheNodeAssignment) {
```
bryanlb, line 182: Try to limit visibility of these methods as much as possible (ie, use `private`, `protected` can help in writing tests). If it's only public to facilitate tests, consider adding the `@VisibleForTesting` annotation.

astra/src/main/java/com/slack/astra/chunkManager/CachingChunkManager.java
```diff
@@ -63,41 +90,67 @@ protected void startUp() throws Exception {
...
+              LOG.error("Error closing readonly chunk", e);
+            }
+          });
+    }
 
     cacheSlotMetadataStore.close();
     searchMetadataStore.close();
     snapshotMetadataStore.close();
     replicaMetadataStore.close();
+    cacheNodeAssignmentStore.close();
```
bryanlb, line 153: Generally these should be reverse order of initialization

astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
```diff
@@ -0,0 +1,556 @@
...
+  /**
+   * Persists cache node assignments in the cache node assignment store.
+   *
+   * @param cacheNodeAssignmentStore the store to persist cache node assignments
+   * @param newAssignments a map of new assignments keyed by cache node ID
+   * @param replicaSet the replica set for which to persist assignments
+   * @param maxConcurrentAssignments the maximum amount of assignments to create for each cache node
+   * @return the number of successful assignments
+   */
+  // pass in current cacheNodeId => loading assignments
```
bryanlb, line 316: Is this a leftover comment?

astra/src/main/java/com/slack/astra/clusterManager/ClusterHpaMetricService.java
```diff
@@ -143,6 +177,23 @@ private void publishCacheHpaMetrics() {
     }
   }
 
+  private static double calculateDemandFactor(
+      long totalCacheSlotCapacity,
+      long totalReplicaDemand,
+      long totalCacheNodeCapacityBytes,
+      long totalAssignedBytes) {
+    // Attempt to calculate hpa value from ng dynamic chunk cache nodes if no cache slot capacity
+    if (totalCacheSlotCapacity == 0) {
```
bryanlb, line 186: Not a bad approach 👍

astra/src/main/java/com/slack/astra/clusterManager/ClusterHpaMetricService.java
```diff
@@ -160,15 +211,38 @@ protected static double calculateDemandFactor(
...
+        "Calculating demand factor from ng cache nodes: bytes needed: {}, capacity: {}, demandFactor: {}",
+        totalBytesRequiringAssignment,
+        totalCacheNodeCapacityBytes,
+        Math.ceil(rawDemandFactor * 100) / 100);
+    return Math.ceil(rawDemandFactor * 100) / 100;
+  }
+
   /** Updates or inserts an (ephemeral) HPA metric for the cache nodes. This is NOT threadsafe. */
-  private void persistCacheConfig(String replicaSet, Double demandFactor) {
+  private synchronized void persistCacheConfig(String replicaSet, Double demandFactor) {
```
bryanlb, line 231: Why do you use both synchronized and a reentrant lock here? The latter is generally a pattern to use now, in case something was ever refactored in the future and we wanted to run this in a virtual thread.

astra/src/main/java/com/slack/astra/clusterManager/ClusterMonitorService.java
```diff
@@ -54,9 +166,147 @@ public ClusterMonitorService(
         "cached_service_nodes_size", datasetMetadataStore, store -> store.listSync().size());
   }
 
+  private void updatePerPodMetrics() {
+    Thread.ofVirtual()
```
bryanlb, line 170: Why are we doing two separate virtual threads here?

astra/src/main/java/com/slack/astra/clusterManager/ClusterMonitorService.java
```diff
@@ -54,9 +166,147 @@ public ClusterMonitorService(
...
+                                                calculateFreeSpaceForPod(cacheNodeMetadata.id)))))
+                        .collect(Collectors.toUnmodifiableList()),
+                    true));
+  }
+
+  private void cacheNodeAssignmentListener(CacheNodeAssignment assignment) {
+    // only update metrics if something has become live
+    if (assignment.state == Metadata.CacheNodeAssignment.CacheNodeAssignmentState.LIVE) {
+      LOG.info("updating per pod statistics due to assignment change");
+      Thread.ofVirtual().start(() -> updateLiveChunksPerPod(assignment));
```
bryanlb, line 209: Same here, is there a reason to use multiple virtual threads?

config/config.yaml
```diff
@@ -62,16 +62,17 @@ metadataStoreConfig:
 cacheConfig:
   slotsPerInstance: ${ASTRA_CACHE_SLOTS_PER_INSTANCE:-10}
   replicaSet: ${ASTRA_CACHE_REPLICA_SET:-rep1}
-  dataDirectory: ${ASTRA_CACHE_DATA_DIR:-/tmp}
+  dataDirectory: ${ASTRA_CACHE_DATA_DIR:-/astra-data}
```
bryanlb, line 65: Let's leave this default to `/tmp` for now

bryanlb (commented):
Looking pretty good! This is a pretty huge change, but you've done a good job of isolating the new behavior behind a "default false" system property. 

I took a look over the code coverage as well, and it looks like you did a decent job of exercising the new paths. I think with a couple small tweaks this is probably good to merge up.

astra/src/main/java/com/slack/astra/clusterManager/ClusterHpaMetricService.java
kx-chen: Heh this was a remnant of my attempts to debug the manager crashing. I've removed it for now!

astra/src/main/java/com/slack/astra/clusterManager/ClusterMonitorService.java
kx-chen: When we redeploy cache nodes, there ends up being a huge barrage of ZK events as assignments for the old cache nodes are deleted. This is partially what was causing the manager to crash previously. The virtual thread prevents the manager from getting overwhelmed and killed from failing health checks. 

I'm honestly not a huge fan of this approach; I was thinking it might be worth it at this point to make this service a scheduled one and add some debouncing for metric updating? Open to any other ideas!

bryanlb (approved):
Couple nits, but I would consider addressing them in a future PR. I'm not seeing anything that sticks outs that we haven't discussed, and you have already successfully deployed this to several test environments.

PR #968, opened 2024-06-24
make kafka transactions optional and disable by default

vthacker:
###  Summary

make kafka transactions optional and disable by default. Our internal kafka cluster is quite under-provisioned to handle acks=all which is mandatory when using transactions. 

So we want the ability to disable it. Which we'll do for all application log clusters to increase throughput. 

This PR adds some missing docs, a counter for error tracking

### DIFF
--- a/astra/src/test/java/com/slack/astra/server/BulkIngestApiTest.java
+++ b/astra/src/test/java/com/slack/astra/server/BulkIngestApiTest.java
@@ -116,22 +116,10 @@ public class BulkIngestApiTest {
 
     datasetRateLimitingService =
         new DatasetRateLimitingService(datasetMetadataStore, preprocessorConfig, meterRegistry);
-    bulkIngestKafkaProducer =
-        new BulkIngestKafkaProducer(datasetMetadataStore, preprocessorConfig, meterRegistry);
 
     datasetRateLimitingService.startAsync();
-    bulkIngestKafkaProducer.startAsync();
 
     datasetRateLimitingService.awaitRunning(DEFAULT_START_STOP_DURATION);
-    bulkIngestKafkaProducer.awaitRunning(DEFAULT_START_STOP_DURATION);
-
-    bulkApi =
-        new BulkIngestApi(
-            bulkIngestKafkaProducer,
-            datasetRateLimitingService,
-            meterRegistry,
-            400,
-            Schema.IngestSchema.newBuilder().build());
   }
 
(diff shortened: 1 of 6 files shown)


docs/topics/System-properties-reference.md
```diff
@@ -14,6 +14,20 @@ How long Astra will wait before starting the next bulk batch if the last batch w
...
+Enable this to ensure that all documents are written using exactly once semantics 
+and all kafka brokers for a partition acknowledge a write. Defaults to false.
+
+```bash
+-Dastra.bulkIngest.useKafkaTransactions=true
+```
+
+{style="narrow"}
+defaultValue
+: 50
```
bryanlb, line 29: I think this might be a copy-paste error, default value isn't an option here?

docs/topics/Config-options.md
```diff
@@ -57,6 +57,16 @@ indexerConfig:
   maxBytesPerChunk: 1000000
 ```
 
+### maxTimePerChunkSeconds
+Maximum time that a <tooltip term="chunk">chunk</tooltip> can be open before closing and uploading to S3. Defaults to 90 minutes.
+THis configuration is useful for ensuring that chunks are uploaded to S3 within a set time frame,
```
bryanlb, line 62: Typo, `THis`

astra/src/test/java/com/slack/astra/server/BulkIngestApiTest.java
```diff
@@ -66,6 +69,23 @@ public class BulkIngestApiTest {
 
   private static String DOWNSTREAM_TOPIC = "test-topic-out";
 
+  private static boolean useKafkaTransactions;
+
+  @BeforeAll
+  public static void useTransactions() {
+    Random random = new Random();
```
vthacker, line 76: This is the first time we'll be using this sort of paradigm. i.e depending on the test run one code path is executed. 

The other option is to have the `testDocumentInKafkaSimple`  test twice. Thoughts @bryanlb ?

bryanlb: I like the idea of testing it twice personally

bryanlb (approved):

bryanlb:
Note - when rolling this out we'll need to ensure that the kafka props for `isolation.level: read_committed` is removed, as otherwise it will be only reading transactional commits.

PR #969, opened 2024-06-25
indexer should read the timestamp correctly

vthacker:
###  Summary

Followup from https://github.com/slackhq/astra/pull/959 , where the indexer should also be aware that timestamp is in micros


vthacker:
This doesn't matter because the search still takes millis - https://github.com/slackhq/astra/blob/bdf13f684cc6178fd359584fdb85520bf733aeba/astra/src/main/java/com/slack/astra/logstore/search/LogIndexSearcherImpl.java#L84

PR #973, opened 2024-06-28
Add tests for bulk ingest bad documents, more edge case guarding

bryanlb:
###  Summary

This expands the hotfixes done in #972 and #970 to further improve scenarios where documents are sent with malformed data.

### DIFF
--- a/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
+++ b/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
@@ -14,7 +14,6 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.UUID;
 import org.opensearch.action.DocWriteRequest;
 import org.opensearch.action.bulk.BulkRequest;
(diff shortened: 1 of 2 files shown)


astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
```diff
@@ -72,13 +71,25 @@ public static Trace.Span fromIngestDocument(
 
     // See https://blog.mikemccandless.com/2014/05/choosing-fast-unique-identifier-uuid.html on how
     // to improve this
-    String id =
-        Optional.ofNullable(sourceAndMetadata.get(IngestDocument.Metadata.ID.getFieldName()))
-            .map(String::valueOf)
-            .orElse(UUID.randomUUID().toString());
+    String id = UUID.randomUUID().toString();
```
vthacker, line 74: calling `UUID.randomUUID()` could be expensive? Perhaps we should only do it if id is not present?

bryanlb: I think this ends up being the default case, but I've went ahead and moved it to only set if not provided.

vthacker (approved):

Issue #949, opened 2024-06-06
`min_doc_count` isn't respected outside of histogram aggregations


kyle-sammons:
### Describe the bug
The `min_doc_count` field on an aggregation isn't respected outside of histogram aggregations. This results in us getting fields with `0` hits back, despite us explicitly requesting a `min_doc_count` of `1` in our `terms` aggregation.

### Requirements (place an `x` in each of the `[ ]`)**
* [X] I've read and understood the [Contributing guidelines](../CONTRIBUTING.md) and have done my best effort to follow them.
* [X] I've read and agree to the [Code of Conduct](https://slackhq.github.io/code-of-conduct).
* [X] I've searched for any related issues and avoided creating a duplicate issue.

### To Reproduce
0. Submit a query for a (non histogram) aggregation that returns multiple fields
1. Set the `min_doc_count` to be higher than one of the values
2. Observe the field still being returned in the aggregation

### Expected behavior
An field with a value lower than `min_doc_count` is filtered out of an aggregation.

#### Screenshots
N/A

#### Reproducible in:

Astra version: All

JVM version: All

OS version(s): All

#### Additional context
N/A

bryanlb:
@allcontributors please add @kyle-sammons for bug

PR #974, opened 2024-06-28
Add missing min doc count param to terms agg

bryanlb:
###  Summary

Adds a missing param on the terms aggregation where min doc count wasn't being correctly set. Adds tests to verify that param is set as expected now.


bryanlb:
Closes #949

PR #987, opened 2024-07-10
Refactored query parsing to support all OpenSearch queries

kyle-sammons:
###  Summary
This PR refactors query parsing to support all OpenSearch queries by making use of their query parsing libraries. This still maintains support for the old query path should people want to use that, but that path will likely be remove in the near future.

While this doesn't totally resolve our OpenSearch API compatibility issues (see: https://github.com/slackhq/astra/issues/798), it does get us remarkably closer

### Requirements

### DIFF
--- a/astra/pom.xml
+++ b/astra/pom.xml
@@ -18,8 +18,8 @@
         <uberjar.name>astra</uberjar.name>
         <http.core.version>4.4.16</http.core.version>
         <protobuf.version>3.25.3</protobuf.version>
-        <grpc.version>1.65.0</grpc.version>
-        <micrometer.version>1.13.1</micrometer.version>
+        <grpc.version>1.65.1</grpc.version>
+        <micrometer.version>1.13.2</micrometer.version>
         <armeria.version>1.27.3</armeria.version>
         <kafka.version>3.5.0</kafka.version>
         <jackson.version>2.17.0</jackson.version>


astra/src/test/java/com/slack/astra/elasticsearchApi/OpenSearchRequestTest.java
```diff
@@ -2,17 +2,27 @@
...
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 public class OpenSearchRequestTest {
 
+  private static ObjectMapper objectMapper;
+
+  @BeforeAll
+  public static void beforeClass() {
+    objectMapper = new ObjectMapper();
```
bryanlb, line 23: Does this need re-initialized for each test?

astra/src/test/java/com/slack/astra/logstore/opensearch/OpenSearchAdapterTest.java
```diff
@@ -78,6 +80,7 @@ public OpenSearchAdapterTest() throws IOException {
             true,
             true));
     openSearchAdapter = new OpenSearchAdapter(fieldDefBuilder.build());
+    System.setProperty("astra.query.useOpenSearchParsing", "true");
```
bryanlb, line 83: This is dangerous, as it sets this for the rest of the tests and not just this one. An easy way to handle this is to read the value from the sysprop at the test init, and then in the cleanup reset it back to what it was or delete it if not originally provided. Since this variable is only expected to have a short lifespan that might be ideal.

Also consider switching the default value to inverted - so all tests run with it as true, and then override it to false at deploy time.

bryanlb (changes requested):
Couple questions, and a few suggestions but otherwise looks good. 👍

astra/src/test/java/com/slack/astra/elasticsearchApi/OpenSearchRequestTest.java
bryanlb: Ah my bad. I get `@BeforeAll`, `@BeforeEach`, `@Before`, and `@BeforeClass` all confused (the latter two of which are junit 4)

astra/src/main/java/com/slack/astra/logstore/opensearch/OpenSearchAdapter.java
```diff
@@ -192,11 +208,14 @@ public Query buildQuery(
 
         if (queryShardContext.getMapperService().fieldType(LogMessage.SystemField.ALL.fieldName)
             != null) {
-          queryStringQueryBuilder.defaultField(LogMessage.SystemField.ALL.fieldName);
           // setting lenient=false will not throw error when the query fails to parse against
           // numeric fields
           queryStringQueryBuilder.lenient(false);
         } else {
+          // The _all field is the default field for all queries. If we explicitly don't want
```
bryanlb, line 215: Ahh, good find

bryanlb (approved):

PR #1018, opened 2024-08-05
Rework HPA scaledown lock to be more fair

bryanlb:
###  Summary

Reworks the HPA scaledown lock to be fairer in which replicaset is selected for the lock. Using the HPA is only a temporary approach, as this really needs to be an operator.


kx-chen (approved):
yippee!!

PR #1022, opened 2024-08-07
Prioritize most recent assignments per cache node bin

kx-chen:
###  Summary

After creating assignments, persist the top N most recent ones

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
+++ b/astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
@@ -28,6 +28,7 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


astra/src/main/java/com/slack/astra/clusterManager/CacheNodeAssignmentService.java
```diff
@@ -194,6 +195,7 @@ protected void assignReplicasToCacheNodes() {
       // snapshots
       List<SnapshotMetadata> unassignedSnapshots =
           getUnassignedSnapshots(snapshotsWithReplicas, assignedSnapshots);
+      unassignedSnapshots.sort(Comparator.comparing(a -> a.snapshotId));
```
bryanlb, line 198: Why do we sort by `a.snapshotId?` If we want to randomize this list we could instead use Collections.shuffle?

bryanlb: As discussed via Slack, we're looking to ensure consistent assignments across runs at this point.

bryanlb (approved):

PR #1023, opened 2024-08-09
Added support for includes/excludes source

kyle-sammons:
###  Summary
This PR aims to add support for the `_source` field in the `_msearch` API (docs can be found [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#source-filtering)). This feature would enable folks to explicitly choose to include or exclude certain fields from their searches.

### Requirements

### DIFF
--- a/astra/pom.xml
+++ b/astra/pom.xml
@@ -50,7 +50,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>2.0.13</version>
+            <version>2.0.16</version>
         </dependency>
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>


astra/src/main/java/com/slack/astra/chunk/ReadOnlyChunkImpl.java
```diff
@@ -583,7 +583,9 @@ public SearchResult<T> query(SearchQuery query) {
           searchEndTime,
           query.howMany,
           query.aggBuilder,
-          query.queryBuilder);
+          query.queryBuilder,
+          query.includeFields,
```
bryanlb, line 587: I think making this a single `source` field probably makes more sense than individual parameters, so that it matches the original API closer. Especially as we'll be adding quite a few extra args here.

astra/src/main/java/com/slack/astra/logstore/search/LogIndexSearcher.java
```diff
@@ -12,5 +13,7 @@ SearchResult<T> search(
       Long maxTime,
       int howMany,
       AggBuilder aggBuilder,
-      QueryBuilder queryBuilder);
+      QueryBuilder queryBuilder,
+      AstraSearch.SearchRequest.FieldInclusion includeFields,
```
bryanlb, line 17: You'll notice that we don't actually pass the raw proto objects around in the code, and only use these for the wire communication. These get turned into internal objects with a combination of from/to methods (ie, see `SearchResultUtils.fromSearchRequest` for an example).

This is generally preferred, because it allows you to iterate and change the communication mechanism without having to change the internal data structures. It also opens up the opportunity to better test those internal method calls, and to re-use them later in other ways that aren't tied to the original request object.

astra/src/main/java/com/slack/astra/logstore/search/LogIndexSearcherImpl.java
```diff
@@ -79,6 +84,9 @@ public void afterRefresh(boolean didRefresh) {
 
     // initialize the adapter with whatever the default schema is
     openSearchAdapter.reloadSchema();
+    allowIncludeAndExcludeSource =
```
bryanlb, line 87: Since this is a net-new optional field, is there any risk to just shipping support without wrapping it in a boolean check? As the only folks theoretically invoking this branch are those explicitly requesting something with an include/excludes source?

astra/src/main/proto/astra_search.proto
```diff
@@ -23,6 +23,23 @@ message SearchRequest {
   // The fully query object, to be used in the leaf nodes and parsed by OpenSearch
   string query = 8;
 
+  // An explicit list of fields to return per document (effectively an allowlist)
+  FieldInclusion include_fields = 9;
```
bryanlb, line 27: As mentioned above, I think matching these to a single source option that contains both is likely a better option and more closely matches the original API.

bryanlb (changes requested):

bryanlb (approved):

PR #1054, opened 2024-08-27
add sampling rate in tracing config and flow through to armeria builder

autata:
###  Summary
Allow for configurable sampling at the zipkin export level via configuring armeria. This allows us to more efficiently trace astra itself.

This defaults the sample rate to 100% which is a no-op as that is the default behavior without a sampler.

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/server/ArmeriaService.java
+++ b/astra/src/main/java/com/slack/astra/server/ArmeriaService.java
@@ -5,6 +5,7 @@ import brave.context.log4j2.ThreadContextScopeDecorator;
 import brave.handler.MutableSpan;
 import brave.handler.SpanHandler;
 import brave.propagation.TraceContext;
+import brave.sampler.Sampler;
 import com.google.common.util.concurrent.AbstractIdleService;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.brave.RequestContextCurrentTraceContext;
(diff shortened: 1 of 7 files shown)


autata:
@bryanlb can you take a look?

astra/src/main/java/com/slack/astra/server/ArmeriaService.java
```diff
@@ -56,9 +57,16 @@ public static class Builder {
...
     private final List<SpanHandler> spanHandlers = new ArrayList<>();
 
-    public Builder(int port, String serviceName, PrometheusMeterRegistry prometheusMeterRegistry) {
+    private final float traceSamplingRate;
+
+    public Builder(
+        int port,
+        String serviceName,
+        PrometheusMeterRegistry prometheusMeterRegistry,
+        float traceSamplingRate) {
```
baroquebobcat, line 66: Since this is part of the TracingConfig, I think you could drop this from the constructor and instead assign it in `withTracing`

autata: good point. done.

bryanlb:
@autata this looks good! Can you also include documentation for this config, in this PR? https://github.com/slackhq/astra/blob/master/docs/topics/Config-options.md#tracingconfig

autata:
@bryanlb added the documentation update. ptal.

docs/topics/Config-options.md
```diff
@@ -306,7 +307,8 @@ a JSON array of span data.
 Optional common tags to annotate on all submitted Zipkin traces. Can be overwritten by spans at runtime, if keys 
 collide. 
 
-<tip>Recommended common tags: <code>clusterName</code>, <code>env</code></tip>
```
bryanlb, line 309: Did you intend to remove this?

autata: good catch. that was a mistake. restored

autata:
@bryanlb please take another look.

bryanlb (approved):

PR #1056, opened 2024-08-28
use otel collector to receive astra traces and send to astra preprocessor (self ingestion of traces)

autata:
###  Summary
Does self ingestion of astra traces rather than using the zipkin plugin.  I see in other tracing projects, they show functionality with self ingestion. I thought a good way to do this was to utilize otel collector to receive and send spans to the preprocessor.

Uses otel-collector for self-ingestion into the preprocessor.  This shows a small change to the bulk ingest api to support otlp collected traces exported using the opensearch exporter.

This is blocked by: https://github.com/slackhq/astra/pull/1054 (and setting a reasonable default, maybe 1% sampling) so we dont go into a spiral of self-ingestion.

Potentially, a better implementation would be to implement the OTLP API as an ingestion input.

### Requirements

### Screenshot
image

Note: it looks like a processor is needed to fix duration on this. (todo if we want to merge this example)

### DIFF
--- a/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
+++ b/astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
@@ -111,6 +111,10 @@ public class BulkApiRequestParser {
           ByteString.copyFromUtf8(
               String.valueOf(sourceAndMetadata.get(LogMessage.ReservedField.TRACE_ID.fieldName))));
       sourceAndMetadata.remove(LogMessage.ReservedField.TRACE_ID.fieldName);
+    } else if (sourceAndMetadata.get("traceId") != null) {
+      spanBuilder.setTraceId(
+          ByteString.copyFromUtf8(String.valueOf(sourceAndMetadata.get("traceId"))));
+      sourceAndMetadata.remove("traceId");
     }
     if (sourceAndMetadata.get(LogMessage.ReservedField.NAME.fieldName) != null) {
       spanBuilder.setName(
--- a/otel-collector-config.yml
+++ b/otel-collector-config.yml
@@ -1,0 +1,47 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+  jaeger:
+    protocols:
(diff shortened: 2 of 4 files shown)


astra/src/main/java/com/slack/astra/bulkIngestApi/opensearch/BulkApiRequestParser.java
```diff
@@ -111,6 +111,10 @@ public static Trace.Span fromIngestDocument(
           ByteString.copyFromUtf8(
               String.valueOf(sourceAndMetadata.get(LogMessage.ReservedField.TRACE_ID.fieldName))));
       sourceAndMetadata.remove(LogMessage.ReservedField.TRACE_ID.fieldName);
+    } else if (sourceAndMetadata.get("traceId") != null) {
```
autata, line 114: astra expects trace id to come in as `trace_id` to put into its trace model properly. the otel spec expects and exports it as `traceId`
https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext

I dont necessarily think this is the right fix to be compatible with OTLP and otel collector, but this is an example that works. and i'd like to open a discussion and get thoughts/ feedback.

this ignores understanding `spanId` because from my digging through the code, span_id is not reserved and displaying self ingested spans (in the zipkin grafana plugin) works with only this change. even though otlp exported spans have `spanId` rather than `span_id`

otel-collector-config.yml
```diff
@@ -0,0 +1,41 @@
+receivers:
+  otlp:
```
autata, line 2: it is nice to include the otlp receiver. so on local, you can push other traces in using `telemetrygen` with a command like this:
`$GOBIN/telemetrygen traces --otlp-insecure --service my-service --traces 3`

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/telemetrygen

otel-collector-config.yml
```diff
@@ -0,0 +1,41 @@
...
+    protocols:
+      grpc:
+      http:
+  zipkin:
```
autata, line 6: astra exports traces as zipkin format so this is needed for self ingestion

PR #1074, opened 2024-09-13
Move fully onto new query parsing

kyle-sammons:
###  Summary
This PR moves us fully onto the new query parsing path, removing all of the old cruft. This enables us to be far more flexible with the types of queries that we're accepting (for instance allowing queries with time ranges that don't use `gte`/`lte`), which while technically supported before were blocked by some of our old query path code

### Requirements


astra/src/main/proto/astra_search.proto
```diff
@@ -7,12 +7,14 @@ option java_package = "com.slack.astra.proto.service";
 import "schema.proto";
 
 message SearchRequest {
+  // Previously deleted field numbers
```
bryanlb, line 10: Nice

bryanlb (approved):

PR #1149, opened 2024-11-04
Revert "Remove hardcoded timeout in favor of relying on global ZK timeout"

bryanlb:
Reverts slackhq/astra#1130


bryanlb:
image

It appears that occasionally on new cache initialization, larger ZK clusters the `initialized` event is either lost or not correctly triggered in Curator.

PR #1215, opened 2025-01-24
Enable rate limits to automatically update based on the amount of available preprocessors

kyle-sammons:
###  Summary
This PR enables rate limits to automatically update based on the amount of available preprocessors. It accomplishes this by creating a new `MetadataStore` for preprocessors, and having each preprocessor add an ephemeral node into that store on boot. Each preprocessor then adds a listener to that store, so that when a new preprocessor comes online (i.e. adds a new ephemeral node to the store), or goes offline (i.e. an ephemeral node is removed), the rate limit gets updated simply by dividing the cluster-wide rate limit by total number of ephemeral nodes present. Just to prevent preprocessors for somehow running on outdated information (for example if they didn't properly get the ZK event when a new node comes online), the preprocessors will get the preprocessor counts and update their rate limits at configurable intervals (default is 60 seconds).

This should effectively unlock the ability to enable auto-scaling on preprocessors as a whole.

This PR also fixes a couple of hidden bugs surrounding synchronization in ZK that I ran into while building it out.

### Requirements

### DIFF
--- a/astra/pom.xml
+++ b/astra/pom.xml
@@ -17,7 +17,7 @@
         <javac.target>21</javac.target>
         <uberjar.name>astra</uberjar.name>
         <http.core.version>4.4.16</http.core.version>
-        <protobuf.version>3.25.5</protobuf.version>
+        <protobuf.version>3.25.6</protobuf.version>
         <grpc.version>1.70.0</grpc.version>
         <micrometer.version>1.14.3</micrometer.version>
         <armeria.version>1.27.3</armeria.version>


astra/src/main/java/com/slack/astra/bulkIngestApi/DatasetRateLimitingService.java
```diff
@@ -48,24 +62,61 @@ public DatasetRateLimitingService(
...
+    }
+  }
+
   @Override
   protected void startUp() throws Exception {
-    updateRateLimiter();
     datasetMetadataStore.addListener(datasetListener);
+    this.preprocessorMetadataStore.addListener(preprocessorListener);
+
+    // We need to await for te cache to be initialized _before_ we try
```
ermontross, line 102: nit: the

ermontross (approved):

PR #1314, opened 2025-05-12
Gate queries from non-active/hydrating nodes

kyle-sammons:
###  Summary
While a cache node is hydrating (when it's downloading data from S3), it uses a considerable amount of memory. Under the current implementation, after each chunk is downloaded, it's immediately available to be searched, and the cache node starts receiving queries. The combination of heavy memory use with the addition of additional query load, is occasionally more than enough to cause OOMs to be thrown. This causes undefined behavior and has been seen (in our production instances) as doing everything from getting a node stuck to crashing it outright. 

This PR attempts to solve that by adding in the concept of "searchability". In this implementation, a cache node boots and marks itself as unsearchable (formally in ZK it's noted as `searchable` being  `false`). Eventually when the cache node finishes its hydration (determined by there being no cache node assignments assigned to it that are in the `LOADING` state, and there being _at least_ one that is in the `LIVE` state), it marks itself as being searchable. Further, when each `SearchMetadata` is created, it _inherits_ the current `searchable` value from its cache node. That gives us the easy ability to filter out `SearchMetadata` that isn't currently marked as being hosted on searchable hosts in the query path. Once the manager node detects a cache node is searchable again, it marks all of its `SearchMetadata` as searchable as well, allowing queries to be served.

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/clusterManager/CacheNodeSearchabilityService.java
+++ b/astra/src/main/java/com/slack/astra/clusterManager/CacheNodeSearchabilityService.java
@@ -94,7 +94,7 @@ public class CacheNodeSearchabilityService extends AbstractScheduledService {
       boolean searchable = loadingCacheAssignments.size() <= 1 && !liveCacheAssignments.isEmpty();
 
       if (searchable) {
-        LOG.info("Marking cache node {} as searchable", cacheNodeMetadata.id);
+        LOG.info("Marking cache node {} as searchable", cacheNodeMetadata.hostname);
         cacheNodeMetadata.searchable = true;
         cacheNodeMetadataStore.updateSync(cacheNodeMetadata);
 


astra/src/main/java/com/slack/astra/clusterManager/CacheNodeSearchabilityService.java
```diff
@@ -0,0 +1,129 @@
...
+                  }
+                }
+              });
+
+      // This node is only searchable if it doesn't have more than 1 assignment that is
+      // loading AND it has at least one assignment that is live
+      boolean searchable = loadingCacheAssignments.size() <= 1 && !liveCacheAssignments.isEmpty();
+
+      if (searchable) {
+        LOG.info("Marking cache node {} as searchable", cacheNodeMetadata.id);
```
ermontross, line 97: Is the `id` useful in a log message here? Can we identify it without looking at the code?

astra/src/main/java/com/slack/astra/logstore/search/AstraDistributedQueryService.java
```diff
@@ -240,6 +245,13 @@ protected static Map<String, List<SearchMetadata>> getMatchingSearchMetadata(
         continue;
       }
 
+      if (!searchMetadata.isSearchable()) {
+        LOG.info(
+            "Skipping searching search metadata={} because it's not searchable!",
+            searchMetadata.name);
```
ermontross, line 251: What is the `searchMetadata.name`?

ermontross (approved):
had a few log message questions but otherwise this lgtm, pretty simple solution nice!

ermontross (approved):

PR #1341, opened 2025-06-10
Bburkholder/ng metadata

bryanlb:
###  Summary

Introduces an abstraction to explore migrating from Zookeeper to Etcd. 

Structure now follows:
* AstraMetadataStore
  * ZookeeperMetdataStore
  * EtcdMetdataStore
* AstraPartitionedMetadataStore
  * ZookeeperPartitionedMetdataStore
  * EtcdPartitionedMetdataStore

To support this structure we now pass around the previously existing MetdataStoreConfig instead of the ZkStoreConfig.
The Astra metadata stores now have a config value that indicates which store / combination of stores to resolve to at runtime.

If only one store is provided to AstraMetdataStore, it will use that store regardless of the config selected. This is to facilitate testing and validation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,104 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Building and Running
+
+### Building the Project
+```bash
+# Build the entire project
+mvn clean install
+
+# Build without running tests
+mvn clean install -DskipTests
+```
+
+### Running Tests
+```bash
+# Run all tests
+mvn test
+
+# Run a specific test
+mvn test -Dtest=AstraTest
+
+# Run a specific test method
+mvn test -Dtest=AstraTest#testDistributedQueryOneIndexerOneQueryNode
+```
+
+### Code Formatting
+```bash
+# Format code using Google's style guide
+mvn fmt:format
+```
+
+### Running Benchmarks
+```bash
+# Run all benchmarks
+cd benchmarks
+./jmh.sh
+
+# Run specific benchmark
+./jmh.sh IndexingBenchmark
+
+# Run specific benchmark method
+./jmh.sh IndexingBenchmark.measureIndexingAsKafkaSerializedDocument
+```
+
+## Development Environment
+
+### Docker Compose
+Astra is designed to be run with multiple components working together. The easiest way to set up a development environment is using the provided Docker Compose file:
+
+```bash
+# Start all services defined in docker-compose.yml
+docker build -t slackhq/astra .
+docker-compose up
+```
+
+This will start all required dependencies (Zookeeper, Kafka, S3Mock, OpenZipkin) and the Astra services (Preprocessor, Index, Manager, Query, Cache, Recovery).
+
+### Configuration
+Configuration is managed through YAML files and environment variables. The main configuration file is in `config/config.yaml`. All settings can be overridden by environment variables, as shown in the Docker Compose file.
+
+To run Astra with a specific config:
+```bash
+java -jar astra/target/astra.jar /path/to/config.yaml
+```
+
+## Architecture Overview
+
+Astra is a cloud-native search and analytics engine for log, trace, and audit data, built on Apache Lucene. The system consists of multiple services that work together:
+
+1. **Preprocessor** (port 8086): Handles data ingestion, schema validation, and writes to Kafka
+2. **Indexer** (port 8080): Consumes data from Kafka and builds Lucene indexes
+3. **Manager** (port 8083): Coordinates between components and handles metadata
+4. **Query** (port 8081): Provides API endpoints for searching data
+5. **Cache** (port 8082): Manages cached replicas for faster queries
+6. **Recovery** (port 8085): Handles data recovery operations
+
+Each component can be run separately using the `NODE_ROLES` configuration option, which makes Astra horizontally scalable.
+
+### Data Flow:
+1. Data is ingested through Preprocessor (or directly to Kafka)
+2. Indexer consumes data from Kafka and builds Lucene indexes
+3. Data is stored in chunks which can be persisted to S3
+4. Manager coordinates replica creation and assignment
+5. Query service handles search requests, distributing them across nodes
+6. Cache service provides faster access to frequently accessed data
+
+### Key Components:
+- **ChunkManager**: Manages data chunks (IndexingChunkManager, CachingChunkManager, RecoveryChunkManager)
+- **MetadataStore**: ZooKeeper-based metadata storage for datasets, replicas, schemas, etc.
+- **ArmeriaService**: HTTP/gRPC server for all components
+- **BlobStore**: Interface for S3 storage
+- **FieldRedaction**: Manages field-level redaction for sensitive data
+
+## API Compatibility
+- Astra provides OpenSearch/Elasticsearch API compatibility for easy integration with tools like Grafana
+- Zipkin API for tracing support
+
+## Testing
+Astra has comprehensive unit tests and integration tests. The main classes involved in testing are:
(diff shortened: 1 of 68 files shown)


astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
```diff
@@ -1,331 +1,708 @@
...
+    this.etcdStore = etcdStore;
+
+    // Override mode if one of the stores is null
+    if (zkStore == null && etcdStore != null) {
+      this.mode = MetadataStoreMode.ETCD_EXCLUSIVE;
+      LOG.info("ZK store is null, overriding mode to ETCD_EXCLUSIVE regardless of configured mode");
+    } else if (etcdStore == null && zkStore != null) {
+      this.mode = MetadataStoreMode.ZOOKEEPER_EXCLUSIVE;
+      LOG.info(
+          "Etcd store is null, overriding mode to ZOOKEEPER_EXCLUSIVE regardless of configured mode");
```
ermontross, line 71: How often would we hit these cases? Are these about to spew logs everytime a metadatastore is accessed?

astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
```diff
@@ -1,331 +1,708 @@
...
+      case ZOOKEEPER_EXCLUSIVE:
+        zkStore.createSync(metadataNode);
+        break;
+      case ETCD_EXCLUSIVE:
+        etcdStore.createSync(metadataNode);
+        break;
+      case BOTH_READ_ZOOKEEPER_WRITE:
+        // In migration to ZK mode, writes only go to ZK
+        zkStore.createSync(metadataNode);
+        break;
```
ermontross, line 120: Do we need a mode where it writes to both? Otherwise we go directly from only writing to ZK (reading from both except we won't have data in ETCD) to only writing to ETCD

astra/src/main/java/com/slack/astra/metadata/core/EtcdPartitioningMetadataStore.java
```diff
@@ -0,0 +1,236 @@
...
+  private final String ASTRA_ETCD_PARTITIONING_DELETE_CALL = "astra_etcd_partitioning_delete_call";
+  private final String ASTRA_ETCD_PARTITIONING_LIST_CALL = "astra_etcd_partitioning_list_call";
+  private final String ASTRA_ETCD_PARTITIONING_GET_CALL = "astra_etcd_partitioning_get_call";
+  private final String ASTRA_ETCD_PARTITIONING_UPDATE_CALL = "astra_etcd_partitioning_update_call";
+  private final String ASTRA_ETCD_PARTITIONING_ADDED_LISTENER =
+      "astra_etcd_partitioning_added_listener";
+  private final String ASTRA_ETCD_PARTITIONING_REMOVED_LISTENER =
+      "astra_etcd_partitioning_removed_listener";
+  private final String ASTRA_ETCD_PARTITIONING_CACHE_INIT_HANDLER_FIRED =
+      "astra_etcd_partitioning_cache_init_handler_fired";
```
ermontross, line 39: For ZK, I believe we have only one set of these and just label it with the store. I don't think we need separate metrics

astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
bryanlb: Writing to both gets to be extremely complicated to manage and reason about state - especially differences in how ephemeral nodes work. My current plan is a point-in-time where everything _new_ is cut over, and then we'll have a migration step to force move some of the stores that rarely change (ie, dataset).

CLAUDE.md
```diff
@@ -0,0 +1,104 @@
...
+
+## API Compatibility
+- Astra provides OpenSearch/Elasticsearch API compatibility for easy integration with tools like Grafana
+- Zipkin API for tracing support
+
+## Testing
+Astra has comprehensive unit tests and integration tests. The main classes involved in testing are:
+- TestKafkaServer and TestingZKServer for local test dependencies
+- AstraTestExecutionListener for JUnit test setup
+- S3MockExtension for S3 mock testing
```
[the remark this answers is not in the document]
bryanlb: This was the auto-generated one from Claude. We can explore hand tuning this as needed, but even the default helps reduce the amount of stuff it needs to "relearn" every time.

astra/src/main/java/com/slack/astra/metadata/core/EtcdPartitioningMetadataStore.java
ermontross: I meant that ZK has one set of metrics for both stores and partitioned stores, so we probably only need one set for etcd, instead of separate partitioned ones

astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
```diff
@@ -1,331 +1,708 @@
...
+                primaryStat -> {
+                  if (primaryStat != null) {
+                    return CompletableFuture.completedFuture(primaryStat);
+                  }
+                  // Not found in primary, check secondary
+                  CompletionStage<Stat> secondaryHas =
+                      mode == MetadataStoreMode.BOTH_READ_ZOOKEEPER_WRITE
+                          ? etcdStore.hasAsync(path)
+                          : zkStore.hasAsync(path);
+                  return secondaryHas.exceptionally(ex -> null);
```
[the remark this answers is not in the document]
bryanlb: I went ahead and removed this exception handling, as I think we want to bubble the underlying errors up instead of transforming them - with the ultimate goal of removing the AstraMetdataStore and it just being an EtcdMetadataStore again in the future.

astra/src/main/java/com/slack/astra/metadata/core/AstraPartitioningMetadataStore.java
```diff
@@ -1,368 +1,629 @@
...
+        // Try to delete from Etcd first (don't wait)
+        etcdStore.deleteAsync(metadataNode);
+        // Then update in ZK (this is the operation we wait for)
+        return zkStore.updateAsync(metadataNode);
+      case BOTH_READ_ETCD_WRITE:
+        // Delete from ZK and create in Etcd
+        // Try to delete from ZK first (don't wait)
+        zkStore.deleteAsync(metadataNode);
+        // Then update in Etcd (this is the operation we wait for)
+        return etcdStore.updateAsync(metadataNode);
```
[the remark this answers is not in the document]
bryanlb: Indeed, I missed the partitioned one. Let me also tweak the implementation slightly so it only does the checkAndUpdate on the mixed modes to retain the current failure modes.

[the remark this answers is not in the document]
bryanlb: Fixed - I also pulled forward the Stat refactor from the other branch since I'm already going to have a disaster of a rebase to sort though 😁 

I'll be able to add tests for the mixed modes once I have the etcd implementation finished, which should be in the following PR.

bryanlb:
@autata we're seeking to replace zookeeper in favor of etcd, which will be initially introduced in #1344. The plan is that you will be able to operate this for a period of time in mixed mode, with all new writes going to etcd. This should result in almost all of the data slowly moving to etcd, with the exception of a few data types that aren't time based (ie, dataset). We'll be providing a migration option to cut over that data, but that isn't yet implemented. 

The motivation for this change is to enable scaling Astra clusters past what is currently feasible with Zookeeper. We've successfully had the code in #1344 deployed to several of our clusters.

We expect that Zookeeper will continue to be supported as a secondary metadata store for now, but if the performance of etcd matches our expectations we'll likely seek to drop Zookeeper support from the main branch later this year.

PR #1353, opened 2025-06-18
Partitioned SearchMetadataStore

ermontross:
###  Summary
Move the `SearchMetadataStore` from a `AstraMetadataStore` to a `AstraPartitionedMetadataStore` to allow for more replica sets, as we are less likely to hit ZK limits by partitioning this store.

This has the new one and a legacy store for the transition period of writing to new, but reading from both. 

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/metadata/search/SearchMetadataStore.java
+++ b/astra/src/main/java/com/slack/astra/metadata/search/SearchMetadataStore.java
@@ -1,8 +1,9 @@
 package com.slack.astra.metadata.search;
 
-import com.slack.astra.metadata.core.AstraMetadataStore;
+import com.slack.astra.metadata.core.AstraPartitioningMetadataStore;
+import com.slack.astra.metadata.core.EtcdPartitioningMetadataStore;
 import com.slack.astra.metadata.core.InternalMetadataStoreException;
-import com.slack.astra.metadata.core.ZookeeperMetadataStore;
+import com.slack.astra.metadata.core.ZookeeperPartitioningMetadataStore;
 import com.slack.astra.proto.config.AstraConfigs;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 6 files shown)


astra/src/main/java/com/slack/astra/metadata/search/SearchMetadataStore.java
```diff
@@ -62,4 +71,66 @@ public CompletionStage<String> updateAsync(SearchMetadata metadataNode) {
...
+    legacyStore.removeListener(listener);
+    super.removeListener(listener);
+  }
+
+  @Override
+  public void close() {
+    try {
+      legacyStore.close();
+    } catch (Exception e) {
+      //      LOG.error("Error closing legacy search metadata store", e);
```
bryanlb, line 127: ?

astra/src/main/java/com/slack/astra/metadata/search/SearchMetadataStore.java
```diff
@@ -62,4 +71,66 @@ public CompletionStage<String> updateAsync(SearchMetadata metadataNode) {
...
+      LOG.error("Error getting search metadata", e);
+      return legacyStore.getSync(path);
+    }
+  }
+
+  @Override
+  public void deleteSync(SearchMetadata metadataNode) {
+    try {
+      super.deleteSync(metadataNode);
+    } catch (Exception e) {
```
bryanlb, line 90: In some cases you log an error, some you do not. You probably should log the fact that all of these are falling back to the legacy store?

bryanlb (changes requested):
I'd like to see some improved tests that validate the merging logic is working as expected. 

* Create a bunch of data into the legacy store
* Instantiate a new search metadata store
* Try to do all the operations and make sure it's what you expect to see

astra/src/main/java/com/slack/astra/metadata/search/SearchMetadataStore.java
ermontross: Is there a case where it would only need to delete from the legacy store?

bryanlb (approved):

PR #1372, opened 2025-07-11
Implement next gen distributed query planner

kyle-sammons:
###  Summary
This PR attempts to implement the first pass of the next generation distributed query planner. Specifically, it tries to solve the following issues in the current query planner:
1. Lack of retries for failing snapshots
2. Lack of failover support, even if multiple replicas are running
3. Lack of continuous partial aggregations, forcing us to aggregate all of the data at once, leading to a far higher chance of OOMs on the query nodes

It accomplishes that by reworking the current query path to divide the snapshots it intends to query into multiple batches. After each request to each batch completes, it keeps track of the requests that succeed vs the ones that fail. For all of the requests that fail, it attempts to resend the request to another node in a different replicaset if one exists (if one doesn't exist, it just marks it as a failure). For all of those that succeed, it performs a partial aggregation on the results. It continues to do this until there are no more snapshots left to query, performs a final aggregation, and then returns the results.

This change *will* impact performance, though from our testing it should be fairly minor on all but the largest clusters. Still, there are a few added config values added that should allow you to tune the behavior:
* `maxSnapshotsQueriedPerBatch` - Controls how many snapshots we query in a single batch. Increase this to increase performance at the expense of resource usage. Decrease it to do the opposite
* `perBatchQueryTimeoutMs` - Controls how long each batch is allowed to run before timing out. Increase it to have slower queries with a higher chance of success. Decrease it to have faster queries with a lower chance of success

**NOTE:** By default this behavior is disabled and the old query planner is still used. To enable the new query planner, on each query sent, include a `!astra.enableNextGenDistributedQueries` in the query string. For example, `"foo: bar !astra.enableNextGenDistributedQueries"`

### Requirements

### DIFF
--- a/astra/pom.xml
+++ b/astra/pom.xml
@@ -19,7 +19,7 @@
         <http.core.version>4.4.16</http.core.version>
         <protobuf.version>3.25.8</protobuf.version>
         <grpc.version>1.73.0</grpc.version>
-        <micrometer.version>1.15.1</micrometer.version>
+        <micrometer.version>1.15.2</micrometer.version>
         <armeria.version>1.32.5</armeria.version>
         <armeria-brave.version>1.27.3</armeria-brave.version>
         <kafka.version>3.5.0</kafka.version>
@@ -28,9 +28,9 @@
         <lucene.version>9.7.0</lucene.version>
         <opensearch.version>2.11.1</opensearch.version>
         <curator.version>5.8.0</curator.version>
-        <log4j.version>2.25.0</log4j.version>
-        <aws.sdk.version>2.31.77</aws.sdk.version>
-        <error.prone.version>2.39.0</error.prone.version>
+        <log4j.version>2.25.1</log4j.version>
+        <aws.sdk.version>2.31.78</aws.sdk.version>
+        <error.prone.version>2.40.0</error.prone.version>
         <junit.jupiter.version>5.13.3</junit.jupiter.version>
         <jetcd.version>0.8.5</jetcd.version>
     </properties>
@@ -289,7 +289,7 @@
         <dependency>
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
-            <version>3.2.1</version>
+            <version>3.2.2</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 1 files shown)


astra/src/main/java/com/slack/astra/logstore/search/AstraDistributedQueryService.java
```diff
@@ -429,15 +440,394 @@ public void recordQueryDuration(long requestedDataHours, long requestDuration) {
...
+                      snapshotToResult.entrySet()) {
+                    String nodeURL = entry.getKey();
+                    SearchResult<LogMessage> result = entry.getValue();
+
+                    if (result.equals(SearchResult.soft_error())) {
+                      LOG.debug("There was a soft error for the entire request!");
+                    } else if (result.equals(SearchResult.error())) {
+                      LOG.debug(
+                          "There was a hard error for the entire request! We need to retry all snapshots for {}",
+                          nodeURL);
```
ermontross, line 578: Would this be a timeout? Would it be useful to add instrumentation here?

astra/src/main/java/com/slack/astra/logstore/search/AstraDistributedQueryService.java
```diff
@@ -429,15 +440,394 @@ public void recordQueryDuration(long requestedDataHours, long requestDuration) {
...
+                    } else {
+                      List<String> allSnapshotsQueriedForNode =
+                          searchMetadataURLToSnapshotNames.get(nodeURL);
+
+                      // These are user errors and in MOST cases don't need to be retried. However,
```
ermontross, line 583: What is an example of a user error and when would it need to be retried?

astra/src/main/java/com/slack/astra/logstore/search/AstraLocalQueryService.java
```diff
@@ -25,6 +25,7 @@ public AstraLocalQueryService(ChunkManager<T> chunkManager, Duration defaultQuer
   @Override
   public AstraSearch.SearchResult doSearch(AstraSearch.SearchRequest request) {
     LOG.debug("Received search request: {}", request);
+
```
ermontross, line 28: ?

ermontross (approved):

PR #1393, opened 2025-07-24
ZK Dataset Metadata Migration to Etcd Feature

ermontross:
###  Summary

Add the ability to migrate ZK Dataset Metadata from ZK to Etcd, since it is the one store that never gets re-written (besides redaction metadata).

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/server/ManagerApiGrpc.java
+++ b/astra/src/main/java/com/slack/astra/server/ManagerApiGrpc.java
@@ -486,4 +486,9 @@ public class ManagerApiGrpc extends ManagerApiServiceGrpc.ManagerApiServiceImplB
       responseObserver.onError(Status.UNKNOWN.withDescription(e.getMessage()).asException());
     }
   }
+
+  @Override
+  public void migrateZKDatasetMetadataStoreToEtcd(ManagerApi.MigrateZKDatasetMetadataStoreToEtcdRequest request) {
+
+  }
 }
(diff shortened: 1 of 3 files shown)


astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
```diff
@@ -327,6 +347,24 @@ public boolean hasSync(String path) {
...
+   * Checks if an ETCD node exists synchronously.
+   *
+   * @param path the path to check
+   * @return true if the node exists, false otherwise.
+   * @throws RuntimeException from EtcdMetadataStore if a connection error or other unrecoverable
+   *     error occurs. Note that the node not existing is NOT treated as an error; in that case, the
+   *     method returns false.
+   */
+  public boolean hasEtcdOnlySync(String path) {
+    boolean existsInEtcd = etcdStore.hasSync(path);
```
bryanlb, line 360: Kinda wonky not just returning the boolean, but this is short-lived code. 🤷

bryanlb (dismissed):

astra/src/main/java/com/slack/astra/metadata/core/AstraMetadataStore.java
ermontross: Wow, I think I was copying from the other method and paring it down and missed this completely. I'll just return the boolean..

PR #1620, opened 2026-04-13
etcd: replace fixed-delay watch retry with exponential backoff + jitter

srri:
## Summary

- Replace fixed-delay watch retry with stateful ExponentialBackOff: 1.5x multiplier, ±50% proportional jitter, time-budgeted retry windows
- Two-tier retry: transient errors (GOAWAY, compaction) share one budget, escalate to error tier on exhaustion. Error budget exhaustion is fatal
- No more 10 retries x 1 second fixed interval.
- Compaction recovery via `REVISION_RESYNC`: re-lists keys from etcd, detects deletes, repopulates cache, notifies listeners. Handles the GOAWAY → OUT_OF_RANGE chain during etcd rolling updates (rejoined members have no history before their snapshot revision)
- Partition discovery watcher (`EtcdPartitioningMetadataStore`) now has full retry/backoff parity with per-partition content watchers
- KeepAlive errors use exponential backoff instead of immediate retry spin-loop
- Fix `ScheduledThreadPoolExecutor(0)` silently dropping scheduled tasks (corePoolSize must be >= 1)
- Remove dead `cacheInitThread`/`leaseInitThread` fields
- Extract helpers: `isDirectChild`, `extractPartition`, `handleFatalAsync` (package-visible), `scheduleRetryWithBackoff`
- New proto fields: `retry_total_duration_ms`, `max_retry_delay_ms`, `initial_retry_interval_ms`

## Motivation

During etcd rolling updates, all watchers on the departing node receive GOAWAY or disconnect simultaneously. The previous retry logic used fixed count + fixed delay (10 retries at 1s), then fatal. The new backoff spreads reconnections with jitter and independent time-budgeted windows per error category.

When a rejoined etcd member has compacted past the watcher's held revision (common during rolling updates since the member rebuilds from a leader snapshot), the watcher now resyncs from etcd instead of retrying with a stale revision or watching from HEAD with a gap.

## Test plan

* [x]  Unit tests for ExponentialBackOff: progression, cap, budget exhaustion (STOP), reset
* [x]  Unit tests for `isGracefulStop` and `isCompactionError` classification
* [x]  Integration tests for EtcdPartitioningMetadataStore with new retry config
* [x]  Deploy to dev environments, and trigger manual tests of: killing etcd pods, kill etcd leader member, replace astra pods, rolling etcd upgrades (as of 4/29 - all of these cases have succeeded)

## Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
+++ b/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
@@ -1066,6 +1066,13 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
           storeFolder,
           currentRevision + 1);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      if (isClosing || Thread.currentThread().isInterrupted()) {
+        LOG.warn(
+            "Ignoring revision fetch error during shutdown for store {}: {}",
+            storeFolder,
+            e.getMessage());
+        return;
+      }
       LOG.error("Failed to get current revision for watch setup on store {}", storeFolder, e);
       watchRetryError.increment();
       if (!scheduleRetryWithBackoff(
--- a/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
+++ b/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
@@ -1067,9 +1067,6 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
           currentRevision + 1);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
       LOG.error("Failed to get current revision for watch setup on store {}", storeFolder, e);
-      if (e instanceof InterruptedException) {
-        Thread.currentThread().interrupt();
-      }
       watchRetryError.increment();
       if (!scheduleRetryWithBackoff(
           retryState.errorBackoff,
--- a/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
+++ b/astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
@@ -2,6 +2,8 @@ package com.slack.astra.metadata.core;
 
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.slack.astra.proto.config.AstraConfigs.EtcdConfig;
+import com.slack.astra.util.ExponentialBackOff;
+import com.slack.astra.util.FatalErrorHandler;
 import com.slack.astra.util.RuntimeHalterImpl;
 import io.etcd.jetcd.ByteSequence;
 import io.etcd.jetcd.Client;
@@ -13,13 +15,18 @@ import io.etcd.jetcd.options.GetOption;
 import io.etcd.jetcd.options.PutOption;
 import io.etcd.jetcd.options.WatchOption;
 import io.etcd.jetcd.watch.WatchEvent;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
 import java.io.Closeable;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
@@ -30,6 +37,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -80,16 +88,41 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
   protected final boolean shouldCache;
   protected final MetadataSerializer<T> serializer;
   private final long etcdOperationTimeoutMs;
-  private final int etcdOperationsMaxRetries;
-  private final long retryDelayMs;
+  private final long retryTotalDurationMs;
+  private final long maxRetryDelayMs;
+  private final long initialRetryIntervalMs;
+  private final ExponentialBackOff keepAliveBackoff;
+
+  /** Fetch the current etcd revision (keysOnly) and watch from there. Does not touch the cache. */
+  private static final long REVISION_LATEST = 0;
+
+  /**
+   * Re-list all keys from etcd, diff against the in-memory cache, and watch from the list revision.
+   * Used on compaction recovery where the watcher's old revision has been compacted away.
+   */
+  private static final long REVISION_RESYNC = -1;
+
+  static final long DEFAULT_RETRY_TOTAL_DURATION_MS = 60_000;
+  static final long DEFAULT_MAX_RETRY_DELAY_MS = 10_000;
+  static final long DEFAULT_INITIAL_RETRY_INTERVAL_MS = 2_000;
+
+  private static volatile FatalErrorHandler fatalErrorHandler = new RuntimeHalterImpl();
+
+  static void setFatalErrorHandler(FatalErrorHandler handler) {
+    fatalErrorHandler = handler;
+  }
+
+  static void resetFatalErrorHandler() {
+    fatalErrorHandler = new RuntimeHalterImpl();
+  }
+
+  static long positiveOrDefault(long value, long defaultValue) {
+    return value > 0 ? value : defaultValue;
+  }
 
   private final MeterRegistry meterRegistry;
 
   private final CountDownLatch cacheInitialized = new CountDownLatch(1);
-
-  // These fields are no longer used but kept for compatibility
-  private volatile Thread cacheInitThread = null;
-  private volatile Thread leaseInitThread = null;
 
   private final String ASTRA_ETCD_CREATE_CALL = "astra_etcd_create_call";
   private final String ASTRA_ETCD_HAS_CALL = "astra_etcd_has_call";
@@ -103,6 +136,10 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
   private final String ASTRA_ETCD_LEASE_REFRESH_HANDLER_FIRED =
       "astra_etcd_lease_refresh_handler_fired";
 
+  private static final String ASTRA_ETCD_WATCH_RETRY = "astra_etcd_watch_retry";
+  private static final String ASTRA_ETCD_WATCH_RETRY_DELAY = "astra_etcd_watch_retry_delay";
+  private static final String ASTRA_ETCD_RESYNC_SKIP = "astra_etcd_resync_skip";
+
   private final Counter createCall;
   private final Counter hasCall;
   private final Counter deleteCall;
@@ -113,6 +150,11 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
   private final Counter removedListener;
   private final Counter cacheInitHandlerFired;
   private final Counter leaseRefreshHandlerFired;
+  private final Counter watchRetryError;
+  private final Counter watchRetryCompaction;
+  private final Counter watchRetryGracefulStop;
+  private final Timer watchRetryDelay;
+  private final Counter resyncSkip;
 
   /** Constructor that accepts an external etcd client instance with default persistent mode. */
   public EtcdMetadataStore(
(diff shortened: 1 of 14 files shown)


astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -152,8 +255,16 @@ public EtcdMetadataStore(
...
+    this.retryTotalDurationMs =
+        config.getRetryTotalDurationMs() > 0
+            ? config.getRetryTotalDurationMs()
+            : DEFAULT_RETRY_TOTAL_DURATION_MS;
+    this.maxRetryDelayMs =
+        config.getMaxRetryDelayMs() > 0 ? config.getMaxRetryDelayMs() : DEFAULT_MAX_RETRY_DELAY_MS;
+    this.initialRetryIntervalMs =
+        config.getInitialRetryIntervalMs() > 0
+            ? config.getInitialRetryIntervalMs()
+            : DEFAULT_INITIAL_RETRY_INTERVAL_MS;
```
ermontross: Do we need to be getting defaults here? Shouldn't this logic happen in the config since we already have `getRetryTotalDurationMs()`?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -103,6 +111,97 @@ public class EtcdMetadataStore<T extends AstraMetadata> implements Closeable {
...
+   * Stateful exponential backoff with proportional jitter. Each call to {@link
+   * #nextBackOffMillis()} returns a randomized delay and advances the internal interval by the
+   * multiplier. Returns {@link #STOP} when the elapsed time since the first call exceeds the
+   * configured maximum.
+   *
+   * <p>The randomization applies ±{@code RANDOMIZATION_FACTOR} around the current interval (e.g.
+   * ±50%), so jitter scales proportionally with the backoff. The multiplier advances the base
+   * interval each call, and overflow is avoided by capping before multiplication.
+   */
+  static class ExponentialBackOff {
```
kyle-sammons: This feels like a strange thing to have in this class. Perhaps we could move it to the `util` package?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -226,8 +345,54 @@ public EtcdMetadataStore(
...
+   * NO_ERROR, debug data: graceful_stop"
+   */
+  static boolean isGracefulStop(Throwable error) {
+    String msg = error.getMessage();
+    return msg != null && msg.contains("NO_ERROR") && msg.contains("graceful_stop");
+  }
+
   /** Creates KeepAlive GRPC connection and handles error and completed cases */
   private void startKeepAlive() {
+    keepAliveBackoff.compareAndSet(null, newBackoff());
```
kyle-sammons: NIT: Is there much point to this `compareAndSet`? It's called again before we call `keepAliveBackoff.get()` down below

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: I don't think we do, I'll re-work this.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -250,11 +416,28 @@ public void onError(Throwable t) {
...
+                long delayMs = backoff.nextBackOffMillis();
+                if (delayMs == ExponentialBackOff.STOP) {
+                  LOG.error(
+                      "KeepAlive retry budget exhausted for lease {} after {} ms — failing fatally",
+                      sharedLeaseId,
+                      backoff.getElapsedTimeMs());
+                  Thread fatalThread = new Thread(() -> new RuntimeHalterImpl().handleFatal(t));
+                  fatalThread.setName("etcd-fatal-keepalive-" + sharedLeaseId);
+                  fatalThread.setDaemon(true);
+                  fatalThread.start();
```
kyle-sammons: Hm, I'm a bit confused. Why are we fatal-ing in a new thread? Why not call `RuntimeHalterImpl().handleFatal(t)` here?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -250,11 +416,28 @@ public void onError(Throwable t) {
...
+                  return;
+                }
+                String errorType = isGracefulStop(t) ? "graceful GOAWAY" : t.getMessage();
```
kyle-sammons: NIT: Is there much reason for this variable? Wouldn't the `t.getMessage()` for a `GOAWAY` contain that?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -816,25 +999,50 @@ public List<T> listSyncUncached() {
...
+    final ExponentialBackOff errorBackoff;
+
+    WatchRetryState(long initialIntervalMs, long maxIntervalMs, long maxElapsedMs) {
+      this.goawayBackoff = new ExponentialBackOff(initialIntervalMs, maxIntervalMs, maxElapsedMs);
+      this.compactionBackoff =
+          new ExponentialBackOff(initialIntervalMs, maxIntervalMs, maxElapsedMs);
+      this.errorBackoff = new ExponentialBackOff(initialIntervalMs, maxIntervalMs, maxElapsedMs);
+    }
+  }
+
```
kyle-sammons, line 935: Is there a strong reason to have different backoffs per error? It feels like that might be complicating things

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -957,42 +1160,81 @@ private void addListener(
...
+                  LOG.error(
+                      "Watch retry budget exhausted for store {} after {} ms — failing fatally",
+                      storeFolder,
+                      retryState.errorBackoff.getElapsedTimeMs());
+                  // System.exit() runs shutdown hooks synchronously, which would block
+                  // the vert.x event loop and prevent other watchers from retrying.
+                  Thread fatalThread = new Thread(() -> new RuntimeHalterImpl().handleFatal(error));
+                  fatalThread.setName("etcd-fatal-" + storeFolder);
+                  fatalThread.setDaemon(true);
+                  fatalThread.start();
```
kyle-sammons: Also a bit confused by this logic here. If we're willing to fatal or call `System.exit()`, then would we really care about the results of other watchers retrying?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
kyle-sammons: Agreed, we should supply defaults in the `config.yaml` and let folks override those as needed

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -957,42 +1160,81 @@ private void addListener(
...
+                        retryState.goawayBackoff.getElapsedTimeMs());
+                    retryState.goawayBackoff.reset();
+                  }
+
+                  // Compaction: revision is behind the compacted point. Re-list from HEAD.
+                  if (isCompactionError(error)) {
+                    watchRetryCompaction.increment();
+                    if (scheduleRetryWithBackoff(
+                        retryState.compactionBackoff,
+                        () -> addListener(listener, 0, retryState),
```
ermontross: Starting from revision `0` is a lot of processing to get back up to the latest revision. I don't think we'd want to do this. Also compaction is configurable and should never be at a point where a watcher would be trying to connect to an already-compacted revision.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: The first implementation of this didn't break out the shutdown as its own thread, and that resulted in thousands of logs like (for each other watcher):
```
Thread vert.x-eventloop-thread-0 has been blocked for 86387 ms, time limit is 2000 ms
```
It might just be noise ultimately in the logs, so could revert this, but also don't think its bad to leave it.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: originally I didnt, but then after adding counters for the kinds of "errors" we see, nearly half are these "OUT_OF_RANGE" compaction responses. The risk here is that we eat through the retry budget erroneously (and backoff further), so that's why these are separated into their own buckets. The actual values (initial,max, etc.) should be the same. We could combined "GO_AWAY" and "COMPACTION" to be its own error budget, and leave errorBackoff as its own. Can see it broken out [here](https://grafana-dev.tinyspeck.com/explore?schemaVersion=1&panes=%7B%22zkh%22:%7B%22datasource%22:%22000000019%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%28astra_etcd_watch_retry_total%7Bastra_cluster_name%3D%5C%22presto%5C%22%7D%29%20by%20%28reason%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22000000019%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221776205305728%22,%22to%22:%221776216589310%22%7D%7D%7D&orgId=1).

srri: More clearly, this is the goal:

  1. Transient "Error" responses (GOAWAY or compaction): non-error disruptions with a shared budget.
  2. Error: real failures. When exhausted, fatal.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: Yeah can remove this.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: See my same comment as the other one re: blocked thread logs.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: Yeah not best for it to be here, will move it.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
srri: I think there is some nuance here. 0 is not fetching from the beginning, 0 represents HEAD (ie: current rev / now) which would be the latest [rev](https://github.com/slackhq/astra/pull/1620/changes/BASE..26434f5a59a34fff9a4da5a5e5b724c9c93e29b1#diff-11499b144495540a42683378c2d8293973796ac6dea56bdb4282239cb4cd14b8R1066-R1086). 
However, with the current logic, it's possible that we skipped/missed events so we need to modify this to keep the cache current.

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -151,9 +193,14 @@ public EtcdMetadataStore(
...
 
-    // Store retry configuration for watch operations
-    this.etcdOperationsMaxRetries = Math.max(0, config.getOperationsMaxRetries());
-    this.retryDelayMs = Math.max(0, config.getRetryDelayMs());
+    this.retryTotalDurationMs =
+        positiveOrDefault(config.getRetryTotalDurationMs(), DEFAULT_RETRY_TOTAL_DURATION_MS);
+    this.maxRetryDelayMs =
+        positiveOrDefault(config.getMaxRetryDelayMs(), DEFAULT_MAX_RETRY_DELAY_MS);
+    this.initialRetryIntervalMs =
+        positiveOrDefault(config.getInitialRetryIntervalMs(), DEFAULT_INITIAL_RETRY_INTERVAL_MS);
```
ermontross, line 201: The `config.get` functions should already be returning either the set value or the default, correct?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -182,7 +238,7 @@ public EtcdMetadataStore(
     // Initialize watch retry executor - scheduled cached thread pool that scales to 0
     this.watchRetryExecutor =
         Executors.newScheduledThreadPool(
-            0, // Use 0 core threads so it can scale down completely
+            1,
```
ermontross, line 241: Based on that old comment, will changing this prevent scaling down?

astra/src/main/java/com/slack/astra/metadata/core/EtcdMetadataStore.java
```diff
@@ -226,6 +282,70 @@ public EtcdMetadataStore(
...
   }
 
+  /** Closes a watcher, suppressing any exception. */
+  private static void closeQuietly(Watcher watcher) {
+    if (watcher != null) {
+      try {
+        watcher.close();
+      } catch (Exception e) {
+        LOG.debug("Error closing watcher", e);
+      }
```
ermontross, line 292: If a watcher fails to close, will the Astra node be prevented from shutting down?

astra/src/main/java/com/slack/astra/metadata/core/EtcdPartitioningMetadataStore.java
```diff
@@ -527,42 +707,72 @@ public List<T> listSyncUncached() {
...
+    // Remove stores for partitions that no longer exist in etcd
+    for (String existing : metadataStoreMap.keySet()) {
+      if (!discoveredPartitions.contains(existing)) {
+        EtcdMetadataStore<T> staleStore = metadataStoreMap.remove(existing);
+        if (staleStore != null) {
+          LOG.info("Removing stale partition store during resync: {}", existing);
+          staleStore.close();
+        }
+      }
+    }
```
ermontross, line 765: If the partition no longer exists, why would the store?

ermontross (approved):
comments addressed in person

PR #1641, opened 2026-05-13
fix: upload only IndexCommit files during chunk snapshot to S3

srri:
### Summary

Fixes an indexer crash during chunk snapshot to S3.

The previous implementation used `S3TransferManager.uploadDirectory()`, which scans the entire Lucene index directory at upload time. After `forceMerge(1)` during `preSnapshot()`, orphaned segment files briefly exist on disk before Lucene's `KeepOnlyLastCommitDeletionPolicy` reaps them. 

The directory scan discovers these files, but they get deleted before the SDK reads them, causing `NoSuchFileException` and crashing the indexer.

Now uploads only the specific files listed in the `IndexCommit`, the orphaned files should not be in the list, so the race disappears.

### Changes

- `BlobStore.upload` takes an explicit file list. uploads are capped at `MAX_CONCURRENT_UPLOADS` (100) in-flight at a time. This max concurrent is taken from the AWS SDK implementation.
- The 2-arg overload for upload is retained for test usage and delegates to the same implementation.
- Aggregated upload failures attach the first underlying exception as the cause for easier debugging.

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/blobfs/BlobStore.java
+++ b/astra/src/main/java/com/slack/astra/blobfs/BlobStore.java
@@ -3,9 +3,11 @@ package com.slack.astra.blobfs;
 import static software.amazon.awssdk.services.s3.model.ListObjectsV2Request.builder;
 
 import com.slack.astra.chunk.ReadWriteChunk;
+import java.io.File;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


astra/src/main/java/com/slack/astra/blobfs/BlobStore.java
```diff
@@ -45,46 +49,88 @@ public BlobStore(S3AsyncClient s3AsyncClient, String bucketName) {
...
+    List<Exception> failures = new ArrayList<>();
+    List<String> keysBatch = new ArrayList<>(MAX_CONCURRENT_UPLOADS);
+    List<FileUpload> uploadsBatch = new ArrayList<>(MAX_CONCURRENT_UPLOADS);
 
-        throw new IllegalStateException(
-            String.format(
-                "Some files failed to upload - attempted to upload %s files, failed %s.",
-                Objects.requireNonNull(directoryToUpload.toFile().listFiles()).length,
-                upload.failedTransfers().size()));
+    for (String fileName : fileNames) {
```
srri, line 92: Just as a note, this implementation is really just hoisting a modified version of the "upload directory" functionality of the AWS SDK, but making it support providing a list instead of a dir.

kyle-sammons (approved):

PR #1722, opened 2026-08-24
ETCD Pagination

ermontross:
###  Summary
In https://github.com/slackhq/astra/pull/1714, we added `keysOnly` and threaded through the `shouldCache` value but query nodes still need to fetch the entire ETCD `SearchMetadataStore` because they need a full cluster view to fan-out searches. This is failing on larger clusters with too many chunks of data due to the 4mb GRPC limit. 

This PR paginates on ETCD calls to prevent ever hitting that limit.

### Requirements

### DIFF
--- a/astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
+++ b/astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
@@ -12,8 +12,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 /**
- * Pages are read in ascending key order, each after the first pinned to the first
- * page's revision for a consistent snapshot.
+ * Pages are read in ascending key order, each after the first pinned to the first page's revision
+ * for a consistent snapshot.
  */
 final class EtcdRangePaginator {
   /** etcd treats revision 0 as latest. */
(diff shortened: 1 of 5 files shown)


astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Reads every key under a folder prefix, paging the etcd range request so no single gRPC response
+ * exceeds the client's inbound message size limit (4 MiB default). Pages are read in ascending key
+ * order, each after the first pinned to the first page's revision for a consistent snapshot.
+ */
+final class EtcdRangePaginator {
+  /** etcd treats revision 0 as latest. */
+  private static final long REVISION_LATEST = 0;
+
+  static final long DEFAULT_PAGE_SIZE = 500;
```
srri, line 22: just curious, how did we land on 500? is there a more dynamic way to choose this (querying size of the key first?)

srri (dismissed):

astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
ermontross: looking at ETCD, the largest entry is the cacheNodes which are ~260 bytes, so 500 entries would be 130 KiB max. We could definitely raise the 500 if we wanted

kyle-sammons (approved):

PR #1738, opened 2026-09-15
Fix/indexer shutdown alreadyclosed race

ermontross:
###  Summary
This PR addresses two issues:

1. We are seeing indexers not shut down cleanly and error out and restart. We do not ack messages to kafka until they are indexed so no data loss, but unnecessary fatal errors.

2. We are also seeing retries on deleting snapshots, even though they are already deleted, which is causing error storms. This is due to us actually caching etcd responses after https://github.com/slackhq/astra/pull/1714 and then basing those retries off an out-of-date cached list. 

### Requirements


astra/src/main/java/com/slack/astra/chunk/ReadWriteChunk.java
```diff
@@ -69,7 +69,7 @@ public abstract class ReadWriteChunk<T> implements Chunk<T> {
   public static final String INDEX_FILES_UPLOAD = "index_files_upload";
   public static final String INDEX_FILES_UPLOAD_FAILED = "index_files_upload_failed";
   public static final String SNAPSHOT_TIMER = "snapshot.timer";
-  public static final String LIVE_SNAPSHOT_PREFIX = "LIVE_";
+  public static final String LIVE_SNAPSHOT_PREFIX = SnapshotMetadata.LIVE_SNAPSHOT_PREFIX;
```
srri: why is this changing here, and further down?

srri (dismissed):

astra/src/main/java/com/slack/astra/chunk/ReadWriteChunk.java
ermontross: to keep it consistent, but this ended up being a red herring, updated

kyle-sammons (approved):

PR #1743, opened 2026-09-22
etcd: fix watch retry storms and halt-on-shutdown

srri:
###  Summary

Changes:
   ```
Watch failures re-listed the whole store on every retry, which creates a
    ton of load on large clusters. This has been fixed by scoping when we do
    a full re-list based on the errors we get on retry.

    Adds watchFatalAfterMs (default 5 mins) to fatal if we still are failing
    despite many retries and scopes retryTotalDurationMs to keep-alive and initial
    cache load.

    Extracts two classes from the duplicated retry code in EtcdMetadataStore
    and EtcdPartitioningMetadataStore. EtcdErrorClassifier is now the single
    place that decides reconnect vs. re-list vs. give up, replacing the
    scattered isCompactionError/isGracefulStop message checks
```

### Requirements

### CODE REVIEW

File: astra/src/main/java/com/slack/astra/metadata/core/EtcdPartitioningMetadataStore.java [hunk 1]
```diff
@@ -169,23 +171,14 @@ public EtcdPartitioningMetadataStore(
...
-    // Falls back to the connection timeout (used as the operation timeout for list reads here) when
-    // unset, preserving prior behavior.
-    this.listPageTimeoutMs =
+    // The connection timeout sizes establishing a socket; a paginated range read needs longer.
+    this.operationsTimeoutMs =
         EtcdMetadataStore.positiveOrDefault(
-            etcdConfig.getListPageTimeoutMs(), etcdConfig.getConnectionTimeoutMs());
+            etcdConfig.getOperationsTimeoutMs(), etcdConfig.getConnectionTimeoutMs());
+    this.listPageTimeoutMs =
+        EtcdMetadataStore.positiveOrDefault(etcdConfig.getListPageTimeoutMs(), operationsTimeoutMs);
```
ermontross: can we just get this from the config instead of double-defaulting here?

astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
```diff
@@ -35,6 +35,19 @@ private EtcdRangePaginator() {}
...
+   */
+  static long currentRevision(KV kvClient, ByteSequence prefix, long timeoutMs)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    return kvClient
+        .get(prefix, GetOption.builder().withPrefix(prefix).withLimit(1).withKeysOnly(true).build())
+        .get(timeoutMs, TimeUnit.MILLISECONDS)
+        .getHeader()
+        .getRevision();
+  }
+
```
ermontross: Why is this in the Paginator if it's not used here?

astra/src/main/java/com/slack/astra/metadata/core/EtcdPartitioningMetadataStore.java
srri: I think I can adjust this

astra/src/main/java/com/slack/astra/metadata/core/EtcdRangePaginator.java
srri: Think this is perhaps more of a naming issue for the file since this now encompasses ranged reading and not just paginating (which is how this gets used). Could rename this file to EtcdRangeReader perhaps? or just leave as-is?

ermontross: as-is is fine

ermontross (dismissed):
Talked this through in person, has been tested locally, LGTM

ermontross (approved):

